PR #1620, opened 2026-04-24
[codex] Document LeetCode 0146 LRU recency design

yaseralnajjar:
## Summary
- document the WS4 prerequisite design for LeetCode 0146 LRU recency storage
- choose integer node handles plus maps for O(1) get/put/update/eviction without object cycles or interior mutability
- record the design and validation in the phase execution ledger

## Validation
- `cargo fmt --check`
- `git diff --check`


yaseralnajjar (commented):
Review complete. The 0146 design chooses an O(1) map-plus-node-handle recency representation, avoids cyclic object references and interior mutability, and explicitly handles real cached `-1` values through key presence rather than value sentinels. No blocking issues found.

PR #1630, opened 2026-04-24
[codex] Track LeetCode 0148 owned merge sort blocker

yaseralnajjar:
## Summary
- Add a dedicated issue note for the remaining `0148_sort_list` canonical merge-sort blocker.
- Record that single-chain owned rewrites are working, but two-list owned merge still hits moved-value false positives.
- Mark the blocker as a capability gap rather than replacing it with another drain/sort/rebuild workaround.

## Validation
- `cargo fmt --check`
- `git diff --check`


yaseralnajjar (commented):
Reviewed the 0148 blocker tracking note. No blocking findings: it records the current owned two-list merge limitation, keeps the no-drain/sort/rebuild rule explicit, and links the remaining work to a concrete closure condition.

PR #1634, opened 2026-04-24
[codex] Track LeetCode WS6 fallback debt

yaseralnajjar:
## Summary

- adds a focused follow-up issue for the WS6 silent-fallback helper pattern flagged in review
- records why the 12 fixture helpers are temporary debt and should be lifted by proof rules or explicit safe APIs
- cross-links the follow-up from the phase execution report and closure review residual risks

## Recommendation Applied

Accepted:
- WS6 `unwrap-to-0` / `unwrap-to-empty-string` helpers should not become fixture style
- `0838` direction encoding and `0023` heap-values merge deserve explicit follow-up

Not treated as blockers here:
- trie terminal optional checks are required by current list-subscript semantics
- DSU step limit is an invariant guard, not the same as substituting a user-visible sentinel result

## Validation

- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/issues/ad-hoc-leetcode-divergence-closure-2026-04-24-execution.md
+++ b/issues/ad-hoc-leetcode-divergence-closure-2026-04-24-execution.md
@@ -209,6 +209,14 @@ Required gates:
 - `scripts/run_all_tests.sh --profile quick` PASS
 - `scripts/run_all_tests.sh` PASS
 
+### Post-Closure Review Follow-Up
+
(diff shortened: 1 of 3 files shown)


yaseralnajjar (commented):
Review pass: no blocking findings.

This correctly separates the reviewer concerns into accepted follow-up debt vs. non-blocking observations. It avoids changing the 12 fixtures without the required compiler proof work, and it gives concrete exit criteria so the current sentinel helpers do not become accepted corpus style. Docs-only validation is sufficient for this PR.

PR #3108, opened 2026-08-08
Fix dict missing-key augmented assignment semantics

yaseralnajjar:
## Summary

- model missing-key plain-dict augmented subscript assignment as a checked KeyError instead of a silent no-op
- preserve annotated defaultdict semantic aliases and factory insertion behavior
- honor existing exact-key presence flow facts so guarded augmented assignment remains accepted
- borrow dictionary lookup keys consistently in generated Rust

## Validation

- focused lowering: 5 passed
- full sifr_lowering: 962 passed, 1 ignored
- focused codegen augmented subscript assignment: 13 passed
- full sifr_codegen: 969 passed
- focused subscript assignment: 19 passed
- native pass fixture and exact fail diagnostic fixture passed
- pinned LeetCode 0438 regression fixture checks successfully
- workspace Clippy, rustfmt, file-size, HIR maintainability, and diff hygiene passed
- create-pr profile: all selected functional areas passed, including Python interop 19/19; aggregate Python step exceeded its host-sensitive 600s budget (latest 639.010s) while other worktrees contended for global Cargo locks. This exact variance is owned by plans/issues/active/adhoc_performance_budget_host_variance.md and is not a product failure or Phase 40 prerequisite.

## Review

- Opus round 1 found the guarded-key regression in LeetCode 0438.
- Remediation commit b341b47f9 integrates exact-key flow proofs and adds focused regressions.
- Opus round 2 verdict: SATISFIED, no blocking findings.

Phase: plans/issues/active/ad-hoc-dict-missing-key-augassign-semantics.md

### DIFF
--- a/crates/sifr/tests/e2e/fail/dict_augassign_unhandled_key_error.sifr
+++ b/crates/sifr/tests/e2e/fail/dict_augassign_unhandled_key_error.sifr
@@ -1,0 +1,6 @@
+# expect-error: SIFR-RESULT-0001
+
+
+def main():
+    values: dict[int, int] = {1: 2}
+    values[1] += 3
--- a/crates/sifr/tests/e2e/pass/dict_missing_key_augassign_checked_error.sifr
+++ b/crates/sifr/tests/e2e/pass/dict_missing_key_augassign_checked_error.sifr
@@ -1,0 +1,28 @@
+from sifr.collections import defaultdict
+
+
+def main():
+    present: dict[str, int] = {"ready": 4}
+    try:
+        present["ready"] += 3
+    except KeyError:
+        assert False
+    assert present["ready"] == 7
+
+    missing: dict[str, int] = {}
+    missing_caught = False
+    try:
+        missing["absent"] += 1
+    except KeyError as error:
+        assert error.message == "key not found"
+        missing_caught = True
+    assert missing_caught
+    assert len(missing) == 0
+
+    annotated: dict[str, int] = defaultdict(int)
+    annotated["new"] += 5
+    assert annotated["new"] == 5
+
+    inferred = defaultdict(int)
+    inferred["new"] += 2
+    assert inferred["new"] == 2
--- a/crates/sifr_codegen/src/error_refs.rs
+++ b/crates/sifr_codegen/src/error_refs.rs
@@ -256,10 +256,21 @@ fn collect_stmt_error_refs(
                 collect_stmt_error_refs(finalbody, referenced, builtin_error_classes);
             }
             HirStmt::SubscriptAssign { index, value, .. }
-            | HirStmt::SubscriptAugAssign { index, value, .. }
             | HirStmt::AttributeSubscriptAssign { index, value, .. } => {
                 collect_expr_error_refs(index, referenced, builtin_error_classes);
                 collect_expr_error_refs(value, referenced, builtin_error_classes);
+            }
+            HirStmt::SubscriptAugAssign {
+                index,
+                value,
+                missing_key_error,
+                ..
+            } => {
+                collect_expr_error_refs(index, referenced, builtin_error_classes);
+                collect_expr_error_refs(value, referenced, builtin_error_classes);
+                if let Some(error_ty) = missing_key_error {
+                    collect_type_error_refs(error_ty, referenced, builtin_error_classes);
+                }
             }
             HirStmt::NestedSubscriptAssign {
                 outer_index,
--- a/crates/sifr_codegen/src/lib_codegen_tests/structured_lowering_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/structured_lowering_codegen_tests.rs
@@ -774,6 +774,7 @@ fn test_structured_stmt_path_handles_nested_subscript_augassign_inside_loop_if()
                 op: "*=".to_string(),
                 value: HirExpr::IntLiteral(2),
                 object_ty: Type::List(Box::new(Type::Int)),
+                missing_key_error: None,
             }],
             elif_clauses: vec![],
             else_body: None,
--- a/crates/sifr_codegen/src/lower_stmt/simple_dispatch_and_bindings.rs
+++ b/crates/sifr_codegen/src/lower_stmt/simple_dispatch_and_bindings.rs
@@ -287,7 +287,15 @@ pub(super) fn try_lower_simple_stmt_with_ctx_and_bindings(
             op,
             value,
             object_ty,
-        } => try_lower_simple_subscript_augassign_stmt(object, index, op, value, object_ty),
+            missing_key_error,
+        } => try_lower_simple_subscript_augassign_stmt(
+            object,
+            index,
+            op,
+            value,
+            object_ty,
+            missing_key_error.as_ref(),
+        ),
         HirStmt::Delete { object, index } => try_lower_simple_delete_stmt(object, index),
         HirStmt::Yield { value } => try_lower_simple_yield_stmt(value, ctx),
         HirStmt::With { items, body } => {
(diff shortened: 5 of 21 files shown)


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `verification/areas/algorithmic_compatibility/corpora/leetcode/src/0438_find_all_anagrams_in_a_string.sifr:32` (introduced by `crates/sifr_lowering/src/lower/aug_assign_lowering.rs:313-323`), criteria 1 and 2. `plain_dict_missing_key_error` fires for every plain-dict subscript augassign with no key-presence refinement, so `sMap[old_ch] -= 1` — already guarded by `if old_ch in sMap:` on the preceding line — is now rejected with `error[SIFR-RESULT-0001]: plain dict augmented subscript assignment may fail with 'KeyError'`. Verified by running `./target/debug/sifr check` (built at the candidate SHA) on that fixture. This fixture is not among the 20 recorded failing slugs in `plans/issues/active/ad-hoc-algorithmic-full-corpus-preexisting-failures.md`, i.e. it passed at the base SHA and now fails check in a nightly-blocking pinned corpus; the candidate validation set did not include the algorithmic-compatibility lane, so the failure was not surfaced. Correction: refine the requirement so a statically proven-present key does not force try/except — mirroring the existing `refine_defaultdict_int_augassign_key` refinement invoked at `crates/sifr_lowering/src/lower/aug_assign_lowering.rs:300` — or, if the unconditional checked surface is the intended final semantics, re-run the algorithmic corpus, update the affected fixture, and record the newly failing fixture in its owning issue before merge.

Follow-up findings:
- pre-existing issue: `crates/sifr_codegen/src/stmt_support_emitter/subscript_augassign_delete.rs:269-280` — `del d[k]` on a plain dict lowers to a discarded `remove(&k)`, so a missing key silently no-ops instead of surfacing the Python `KeyError`. Same conditional-update class as the fixed augassign form, but a delete rather than an assignment form.
- pre-existing issue: `crates/sifr_codegen/src/intrinsic_method_emitters/recursive_method_calls.rs:57`, `crates/sifr_codegen/src/intrinsic_method_emitters/collection_methods.rs:172`, and `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_method_and_question_mark.rs:155` retain the `if let Some(__elem) = d.get_mut(&k) { … }` / `else_body: None` shape for `d[k].append(x)` on a plain dict, so missing-key mutation silently no-ops. Method-call mutation form, outside criterion 6's "assignment forms".
- pre-existing issue: `crates/sifr_codegen/src/stmt_support_emitter/subscript_augassign_delete.rs:93-96` and `crates/sifr_codegen/src/lower_stmt/subscript_assignment.rs:604-607` match only `__sifr_defaultdict_int`, so `defaultdict(list)` / `defaultdict(set)` augassign falls through to the plain-dict `get_mut` form with `else_body: None` — and is excluded from the new checked error by the `__sifr_defaultdict_` prefix test, leaving a silent no-op. Unchanged by this diff; criterion 4 holds for the `int` factory, which is the only factory with an `entry`/`or_insert` path.
- suggestion: `crates/sifr_codegen/src/stmt_support_emitter/subscript_augassign_delete.rs:166-172` wraps `RustLiteral::Str("key not found")` in a `.to_string()` method call, but `RustLiteral::Str` already renders as `"…".to_string()` (`crates/sifr_codegen/src/render/render_expr_and_blocks.rs:398`), so the emitted argument is `"key not found".to_string().to_string()`.
- suggestion: the attribute form `self.counts[k] += 1` takes the separate `AttributeSubscriptAssign` path (`crates/sifr_lowering/src/lower/aug_assign_lowering.rs:217-275`) and is not routed through `plain_dict_missing_key_error`; no test was added for it under criterion 6's audit.
- infrastructure issue: the create-PR profile remains in progress due to host contention and fixed 120-second Python interop subprocess timeouts; the affected readonly-check-doctor and binding-authoring suites passed standalone. Not treated as an implementation defect.

Note: the `build_dict_get_mut_key_arg` / `build_dict_lookup_key_arg_for_ir` change (removing the string-literal special case) is a correct fix, not a finding — `RustLiteral::Str` renders as an owned `String`, so the old unreferenced argument could not satisfy `HashMap::get_mut`'s `&Q`.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- pre-existing issue: Presence flow facts are not invalidated by `del d[k]` (`crates/sifr_lowering/src/lower/statements/statement_dispatch.rs:230` lowers `Stmt::Delete` without calling `clear_sequence_guards_for_target`). So `if k in d: del d[k]; d[k] += 1` keeps `key_is_proven_present == true` and re-enters the `else_body: None` conditional-update path. This is the same stale-guard hole the pre-existing guarded-read path in `crates/sifr_lowering/src/lower/guarded_index.rs:16` already has, so the new suppression inherits rather than introduces it; fixing it belongs to the guard machinery, not this item.
- pre-existing issue: `del d[missing_key]` lowers to `let _ = d.remove(&k);` (`crates/sifr_codegen/src/stmt_support_emitter/subscript_augassign_delete.rs:269`), silently succeeding where Python raises `KeyError`. Same silent-no-op class as the augassign bug, in a related plain-dict subscript form, but not an assignment form named by criterion 6 and unchanged by this diff.
- pre-existing issue: `lower_subscript_augassign_stmt_for_ir` only emits factory insertion for the `__sifr_defaultdict_int` alias (`crates/sifr_codegen/src/stmt_support_emitter/subscript_augassign_delete.rs:95`), while `plain_dict_missing_key_error` suppresses the checked error for every `__sifr_defaultdict_*` alias (`crates/sifr_lowering/src/lower/aug_assign_lowering.rs:41`). A `defaultdict(list)`/`defaultdict(set)` augassign therefore falls through to the plain `get_mut`/`else_body: None` shape and no-ops on a missing key. Behavior is identical before and after this diff (both the inferred and the newly aliased annotated form reach the same emitter), and no fixture exercises it, so criterion 4 is not regressed — but the gap is worth recording.
- suggestion: The generated error message is built as `RustLiteral::Str("key not found")` with an extra `.to_string()` method call, rendering as `"key not found".to_string().to_string()`, and the message omits the offending key that Python's `KeyError` reports. Dropping the redundant call and interpolating the key would improve both output and diagnosability.
- suggestion: The `Err` construction hardcodes the path `KeyError::new` instead of deriving the constructor from the `missing_key_error` type that lowering resolved via `ctx.class_types.get("KeyError")`. If a program defines its own `KeyError` with a different constructor shape, lowering would select that type while codegen still emits `KeyError::new(String)`.
- infrastructure issue: The create-PR profile is still validation-in-progress due to concurrent heavyweight gates in other worktrees and the fixed 120-second Python interop timeouts; the affected standalone Python suites passed, so this is host contention rather than an implementation defect.

yaseralnajjar:
Base advance audit: origin/main advanced from 1cb731fcb67c520d35fcf2376a88b2d2a4b255b1 to 604a6c1e9e4465ab7451476b1a58655030084a50 through const-specialization/frontend/diagnostic work. The intervening 80 changed paths have zero intersection with this phase candidate changed paths, so phase-closure evidence remains valid under the unrelated-base-change rule.

yaseralnajjar:
Final validation evidence for exact reviewed head `b341b47f9f8e81baa0d7403979eb6551886e5568`:

- The authoritative merge profile passed coverage-matrix, core/file-size/dependency/stdlib guardrails, audit fixtures, package management, and diagnostics. It then stopped at CPython differential because `bounded_int_arithmetic` hit the fixed 240-second Sifr subprocess timeout while the host was under concurrent worktree compilation; the next three cases completed at 224.9s, 66.8s, and 38.6s. The report remained within the lane wall-time budget and recorded no other failure.
- An immediate isolated replay of the exact selected CPython differential area passed 2/2 variants and all four runtime cases: `bounded_int_arithmetic` 36.3s, `boolean_string_logic` 25.0s, `list_iteration_indexing` 17.1s, and `dict_lookup_order_independent` 18.2s, with matching CPython exit codes and JSON output.
- This is the existing shared-host timing condition owned by `plans/issues/active/adhoc_performance_budget_host_variance.md`; no timeout, threshold, baseline, waiver, or source was changed.
- `origin/main` remains `604a6c1e9e4465ab7451476b1a58655030084a50`, so there is no post-review base drift. The PR is clean and mergeable.

The semantic implementation remains exactly the Opus-satisfied reviewed candidate.

PR #3237, opened 2026-08-18
verification: lock pre-v1 compatibility inventory

yaseralnajjar:
## Scope

Complete Item 0 (`pre_v1_compat_0_contract`) from the active pre-v1 compatibility-removal phase.

- lock canonical and retained contracts in a checked-in inventory
- add an ownership/receiver/mutation-tested inventory checker
- resolve Phase 7 naming conflicts in architecture documentation

## Validation

- `python3 verification/compatibility/check_pre_v1_compatibility_inventory.py`
- `python3 verification/compatibility/check_pre_v1_compatibility_inventory.py --self-test`
- `python3 verification/areas/documentation/check_structure.py`
- `git diff --check`

No compiler files changed; Sifr create-PR and merge gates are intentionally omitted per the phase rule.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Notes on what I verified independently (not re-running the supplied evidence): all 12 `baseline_counts` commands reproduce their recorded counts exactly on the candidate SHA (`Type::BigInt` 92, set helpers 58, copy helpers 13, `legacy_facade` 33, hidden prefixes 2, expect-stdout 17, workspace codes 12, `DiskSourceProvider::new` 20, `.rust_type()` 150, roots 18, exports 15, flat layout 3); `snapshot_base_sha` matches the stated base; all 61 surfaces carry exactly one classification, every `remove` row names exactly one `pre_v1_compat_*` item and every `retain` row one `contract:*` owner; all 13 required retained IDs including `retained-phase40-legacy-index` and `retained-lint-deprecated-status` are present and their consumer paths exist (`RuleStatus::Deprecated` confirmed at `crates/sifr_lint/src/lib.rs:50`); receiver rows lock four spellings with no source-compatibility wording; `sifr.math.fabs`/`sifr.math.pow` are real public re-exports, so the rewritten naming table in `internal_docs/architecture.md:265` is not asserting a nonexistent name; the only `unknown` token in the JSON is the `<unknown>` diagnostic literal, not a placeholder.

Follow-up findings:
- suggestion: `verification/compatibility/check_pre_v1_compatibility_inventory.py:224` — the `private-public-drift` self-test mutates `surfaces[-1]` (a retained row), so it actually exercises the retained-visibility rule, not the public-export vs private-implementation-import distinction. The distinction is enforced only by enum membership; a row could silently mislabel a `_sifr` import as `public-export`. Consider renaming the case and adding a real drift rule.
- suggestion: `check_pre_v1_compatibility_inventory.py:170` — `owner.startswith("pre_v1_compat_")` does not enforce *exactly one* item; `"pre_v1_compat_a and pre_v1_compat_b"` would pass. Current data is clean, but a regex anchor would make the criterion machine-enforced.
- suggestion: `check_pre_v1_compatibility_inventory.py:144` — placeholder detection is whole-value equality for non-owner fields, so `"resolved later"` inside `canonical_contract` would pass. Word-level matching (as already used for `owner`) would close it.
- infrastructure issue: the new checker is not referenced by any profile, runner, or `scripts/run_all_tests.sh` (`rg check_pre_v1_compatibility_inventory` outside its own directory returns nothing), so it will not run in CI until a later item wires it in. Outside this item's stated scope, but worth an owning item.
- suggestion: coarse consumers on some rows (e.g. `language-bigint-transition` lists `crates`, `stdlib`, `docs`) make the existence check near-vacuous; narrower paths would give the checker real drift signal.
- suggestion: `internal_docs/architecture.md:494` says both owned forms select `Owned`, while the inventory's `receiver_contract` labels them `owned` and `owned-mutable`. Same meaning, two vocabularies locked in one item; aligning the terms would avoid a later conflict when `pre_v1_compat_8_receivers` lands.

Issue #27, opened 2026-02-14
Update Codegen for Union Types and Narrowing


yaseralnajjar:
#### **Current Situation**

- The codegen (`crates/sifr_codegen/src/lib.rs`) maps sifr types 1:1 to Rust types (Int -> i64, Str -> String, etc.).
- There is no concept of generating Rust enums for union types.
- There is no concept of generating `match` expressions for type narrowing.
- Function parameters and return types only handle simple types.

#### **Desired Situation**

- Union types generate Rust enums: `int | str` -> `enum IntOrStr { Int(i64), Str(String) }`
- Enum definitions are emitted once at the top of the generated Rust file
- Function parameters/returns with union types use the generated enum name
- Narrowing (isinstance checks) generates `match` expressions that destructure the enum
- Literal types map to their base Rust type with value checking
- Optional types (`str | None`) generate `Option<String>` in Rust (special case)
- Unknown type generates `Box<dyn std::any::Any>` with downcast in narrowing

#### **Suggested Solution**

**Modified files:**
- `crates/sifr_codegen/src/lib.rs`:
  - Add `UnionRegistry`: tracks all union types encountered, generates unique enum names, deduplicates
  - Add `emit_union_enum_definitions()`: emits all union enum `enum X { ... }` definitions at top of file
  - Add `emit_union_construction()`: wraps a value in its enum variant (e.g., `IntOrStr::Int(42)`)
  - Add `emit_narrowing_match()`: generates `match x { EnumName::Variant(val) => { ... } }` for isinstance narrowing
  - Update `emit_function()`: handle union parameter types and return types
  - Update `emit_expr()`: handle union value construction
  - Update `emit_if()`: detect narrowing and emit match instead of if/else when appropriate
  - Special case: `T | None` -> `Option<T>` (Rust idiomatic), narrowing -> `if let Some(val) = x { ... }`
  - Handle `reveal_type()`: emit a comment in generated Rust showing the type

**Key design decisions:**
- Enum naming: concatenate type names (e.g., `IntOrStr`, `GetOrPostOrPutOrDelete`)
- For literal unions of same base type (e.g., `"GET" | "POST"`), use the base type with value validation at construction
- `None` in unions uses `Option<T>` when there are exactly 2 variants (T and None)
- For 3+ variant unions including None, use a full enum with a `None` variant

**Unit tests:**
- Generate enum for `int | str` and verify valid Rust
- Generate `Option<String>` for `str | None`
[...]

PR #3239, opened 2026-08-18
Design static class adapters and Pydantic ergonomics

yaseralnajjar:
## Summary

- define a package-neutral static class-adapter architecture
- redesign Pydantic Sifr around typed model, field, validator, and serializer ergonomics
- add an active, sequential M0-M12 implementation phase and link it from the roadmap and phase index
- retain the previous ad-hoc phase in the archive

## Why

The current Pydantic Sifr surface requires raw metadata and free-function calls. This separates the generic compiler substrate from Pydantic package semantics while planning the full set of ergonomics most users need.

## Validation

- `git diff --check origin/main...HEAD`
- `python3 scripts/check_docs_error_code_links.py`
- No code tests run; this is a documentation-only change and tests were omitted per request.


yaseralnajjar:
Exact-SHA Opus initial M0 review for candidate `b3c757e8bdd22917501a1bf485522c96f9e41464` (base `5285d7cb6d`). This review triggered the one remediation pass in PR #3240.

Reviewed both SHAs directly (`git show`/`git diff`), covering all five changed paths.

**Verdict: NOT SATISFIED**

Blocking findings:

- **in-scope omission** — `internal_docs/native_pydantic_sifr_architecture.md:376` (`### Marker bases`), `:414` (`### Declaration descriptors`), `:587` (`### Attached package APIs`). Criterion: M0 scope "Fix package-author syntax for adapter markers, descriptor functions, adapters, handler references, and attached APIs" and the exit gate "sufficient to implement M1–M12 without a new language mechanism". Reason: the contract fixes semantics but never fixes the package-author declaration form. The doc says only "A package can mark a field-less class as a class-adapter marker" (:378), "A package can mark a const function as one descriptor kind" (:416), and "A package can declare a fixed set of generic functions as an attached API" (:589) — the *marking/declaring* spelling is unspecified for all four descriptor kinds, the marker base, the adapter-provider binding (how a marker names its adapter function and its descriptor type `D`), and the attached-API set. The repo's convention is that package-author surface is spelled concretely (`internal_docs/const_specialization.md:13,16` gives `@const_specialize("package.module", "function")` and `@metadata("package.key", const_value)`), and the phase record defers the spelling to implementation instead (M2 "Add package declarations for field, class, method, and type descriptors", M3 "Add field-less erased class-adapter marker declarations", M7 "Add package declarations for fixed attached API sets"). M2/M3/M7 implementers must therefore invent public package-author syntax that M0 was scoped to lock. Correction: add a syntax subsection to the Static Class-Adapter Contract giving the concrete declaration form for each of: adapter-marker class, adapter/provider + descriptor-type `D` binding, the four descriptor-function kinds, and the attached-API set — in the same `@decorator("package.module", "symbol")` idiom already used for `@const_specialize`/`@metadata`.

- **in-scope omission** — `internal_docs/native_pydantic_sifr_architecture.md:1913` (`## Non-Goals`) and `plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md:169` (`### Deliberate exclusions`). Criterion: "Every excluded feature has a reason and no milestone dependency." Reason: the phase record's 20-item exclusion list carries no reasons at all (only ":193 These exclusions are terminal"), and the Non-Goals additions state a reason for just three (wildcard targeting, `@property` stacking, frozen-model emulation). Several newly excluded features have no reason in either file or in the pre-existing compatibility-disposition list (`:1818`–`:1868`, which covers `from_attributes`, `exclude_unset`, `arbitrary_types_allowed`, etc. but none of these): `validate_call`, Pydantic dataclasses, private attributes, `model_construct`, and "Public wrap-validator and wrap-serializer continuation APIs" (`:1929`). The wrap exclusion is the sharpest gap — the Core Schema node table at `:1135` retains "before, after, wrap and plain typed validators", so the engine supports wrap nodes and only the public API is excluded, with no stated reason. Correction: give each excluded feature one stated reason (either inline in the Non-Goals bullet or as an entry in the compatibility-disposition list), and mirror the reasons in the phase record's exclusion list.

Follow-up findings:

- **suggestion**: no artifact marks each public operation as replace-vs-net-new (`plans/issues/.../ad-hoc-static-class-adapters-and-pydantic-ergonomics.md:75` "Current Baseline" plus the `:229` issue-ownership list plus milestone verbs make this derivable, but only indirectly — e.g. whether `field_serializer`/`computed_field`/`RootModel[T]` are replacements or net-new must be inferred from `pydantic-sifr#14`/`#27` ownership). A short operation table (operation → existing path replaced | net-new → owning milestone) would make the criterion self-evident.
- **suggestion**: `internal_docs/native_pydantic_sifr_architecture.md:562`–`:583` places field/model `before` and `after` handler-stage semantics (including `ValidationError` aggregation) inside the package-neutral `## Static Class-Adapter Contract`. The text does attribute the behavior to "generated package glue", so it is not compiler behavior, but relocating it to the Pydantic package section would keep the "no Pydantic-specific compiler behavior" criterion checkable by section boundary.
- **pre-existing issue**: the candidate changes `ConstPackageIssue` to carry "one primary source-origin ID" (`:849`) while `internal_docs/const_specialization.md:49` still describes "source span". The arch doc defers only the `CallableIdentity` const-contract update to the implementation merge unit; the package-issue wording drift should be reconciled when M1 lands.

PR #3242, opened 2026-08-18
Remove the public bigint transition

yaseralnajjar:
## Summary
- remove the public `bigint` annotation, builtin, runtime type object, and `Type::BigInt` compiler branch
- remove transition-only diagnostics, fixtures, quarantine evidence, and generated documentation
- preserve canonical arbitrary-precision `int`, including exact exponentiation, and retain checked fixed-width boundaries

## Validation
- `cargo test -p sifr_type_system`
- `cargo test -p sifr_lowering` (986 passed, 1 ignored)
- `cargo test -p sifr_codegen` (1021 passed)
- `cargo test -p sifr_runtime int::tests`
- exact-int, heapq, and fixed-width positive/negative fixtures
- diagnostics sync/coverage/schema/hygiene checks
- generated-code representative panic scan and intrinsic panic lint
- create-PR gate reached an inherited verification-taxonomy failure; replay on base `5285d7cb6d8df3338d2964f5af4308c94b9a3f48` produced the identical failure, and none of the reported files changed in this item

Candidate: `87b9e0ea7ee7ccb49a6a064c07491c1a425e94da`


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: canonical `int` still lowers to `i64` inside containers, while `SifrInt`-stored expressions do not. `key: int = 10 ** 20` then `labels: dict[int, str] = {key: "big"}` emits `HashMap<i64, String>` with a `SifrInt` key (uncompilable Rust), and a bare large literal (`key: int = 100000000000000000000`) emits `compile_error!("structured statement emission missing …")`. This is the known `SifrInt`-vs-`i64` migration gap owned by the integer-model work, not by this removal; the new `**` → `SifrInt` branch (`crates/sifr_codegen/src/function_emitter/sifr_int_analysis.rs:835`, `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_binop.rs:763`) widens the set of expressions that reach it. Against the base it is still an improvement: base emitted `(10_i64).pow(20u32)`, i.e. i64 overflow at runtime.
- pre-existing issue: the exponent is emitted as `lowered_right as u32` (`stmt_expr_binop.rs:769` and the pre-existing fallback at `:779`), so a constant exponent above `u32::MAX` truncates silently — `2 ** 5000000000` emits `.pow((5000000000_i64) as u32)` = exponent 705032704 with only a "may overflow" warning. The cast predates this item; negative and non-provable exponents are correctly rejected by SIFR-INT-0005.
- suggestion: `lower_isinstance_call` (`crates/sifr_lowering/src/lower/builtin_calls/bytes_len_range.rs:186`) now calls `resolve_annotation_expr` purely for its diagnostic side effect and discards the result. This is what makes `isinstance(x, bigint)` reject, but it also turns previously-silent forms into hard errors: `isinstance(x, list)` → `SIFR-NAME-0003 unknown type: 'list'` and `isinstance(x, (int, str))` → `SIFR-TYPE-0007`. Neither ever narrowed (`narrowing.rs:20` uses the same resolution set), so failing loudly is defensible, but "unknown type: 'list'" reads wrong for a builtin container name — a purpose-built message would be clearer.
- suggestion: `validate_typevar_spec` (`crates/sifr_lowering/src/lower/typevar_annotations.rs:10`) hardcodes the bound-name list (`Comparable`, `Addable`, …) separately from `type_satisfies_bound` in `type_bounds.rs:447`. Forward-declared classes and aliases resolve correctly (verified), but the two lists can drift; consider a single source for the known spec names.
- infrastructure issue: the create-PR gate stop on verification-taxonomy reproduces identically at the base SHA and touches no file in this diff, consistent with the supplied evidence.

PR #3252, opened 2026-08-18
feat: add erased static class adapters

yaseralnajjar:
## Summary
- add canonical field-less adapter markers separated from ordinary data inheritance
- evaluate bounded typed declaration plans before final class lowering and apply typed metadata plus one specialization request
- erase marker declarations/imports from runtime HIR and add non-Pydantic adapter conformance coverage

## Validation
- cargo test -p sifr_driver early_adapters:: -- --nocapture (10 passed)
- cargo test -p sifr_driver typed_descriptors:: -- --nocapture (10 passed)
- canonical callable/type identity unit test
- direct sifr check and native sifr run of static_class_adapter fixture
- workspace clippy, formatting, HIR maintainability, file-size, and diff checks
- broad non-E2E command: 109 CLI + 12 build-output + 37/38 E2E harness tests passed; only the separately owned pre_v1 fixture protocol_bound_unknown_forwarded_typevar.sifr failed

Exact candidate: 28cb499ba96cabd65e31c7577927911eef9d6648


yaseralnajjar:
Exact-SHA Opus review

- Base: `79e2f6a51ca7166f5aef813b9325817c56dd0038`
- Candidate: `28cb499ba96cabd65e31c7577927911eef9d6648`

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_frontend/src/class_declarations.rs:162` (`field_contracts`) is compared against a plan echo built from `ClassDeclaration` items, but the two lists are derived from different sources. `field_contracts` reads `HirClass.fields`, which includes fields inferred from unannotated `self.x = ...` constructor assignments (`class_type_collection.rs:498`), while `ClassDeclaration` items only contain class-body `AnnAssign` fields. An adapted class that declares any field only in `__init__` can therefore never produce a plan that passes `validate_fields` (`early_adapters.rs:396`), and the resulting message ("must preserve every declared field identity, order, and type") points the blame at the adapter. The same asymmetry appears when a child re-annotates a parent field, since `own_fields` drops it but the declaration item keeps it. Worth either exposing the exact stored-field contract in `DeclarationInput` or restricting the comparison to source-declared fields in M4.
- suggestion: `crates/sifr_frontend/src/descriptor_exports.rs:26` exports every marker into `class_exports` without the `!name.starts_with('_')` boundary that both the ordinary class loop (`query_diagnostics.rs:234`) and the type-alias branch below it apply, while `store` does filter private markers out of `class_adapter_markers`. A `_Private` marker is then importable as a plain class type with no marker binding attached, so `class Model(_Private)` in another module is treated as an ordinary data parent whose struct was erased. This is the one place the M2 "public type-alias export boundary" follow-up is not honored by the new export kind.
- suggestion: marker misuse is only rejected in call position (`crates/sifr_lowering/src/lower/expressions/regular_calls.rs:20`). Annotation-only uses (`def f(x: Contract) -> None`) still resolve through the exported marker class type, and since the class is erased from every HIR module no struct is emitted, so the failure surfaces at codegen/rustc rather than as a local `META_MALFORMED_DECLARATION`. No test covers annotation position.
- suggestion: `crates/sifr_frontend/src/early_adapters.rs:555` (`bounded_plan_rejects_unknown_output_fields`) is vacuous — it constructs a record locally and asserts `fields.len() != 4` without calling `validate_plan`. The "reject method-body output" criterion is exercised only indirectly through the arity checks; a real test should feed an extra plan field through `validate_plan`.
- suggestion: `crates/sifr_lowering/src/lower/descriptor_declarations.rs:383` (`special_base`) changes ordinary-class parent selection relative to the deleted `get_parent_class`. The old function returned `None` as soon as the first `Name` base was special; the new one skips special bases and takes the first remaining one, so `class C(Protocol, Base)` or `class E(Error, Base)` now inherits `Base`. No `.sifr` source in the repo uses that shape, so the change is unexercised in either direction — worth an explicit test or a note if the new reading is intended.
- suggestion: `crates/sifr_frontend/src/early_adapters.rs:207` (`const_functions`) decides local-vs-external by searching `class_adapter_providers` for a provider whose *function name* matches and then comparing its module, even though the caller already has `module_name`. It reduces to the same answer today but couples provider lookup to name uniqueness; comparing `selection.provider_module == module_name` directly would be equivalent and clearer.
- infrastructure issue: neither `verification/areas/core_language/fixtures/static_class_adapter/**` nor the M2 `typed_declaration_descriptors` fixture is referenced by any e2e manifest, runner, or test (the only repo reference is `plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md`). The exit-gate fixture is validated manually per the evidence, so it will not guard against regressions; the in-process `early_adapters.rs` driver tests are the only automated coverage.
- suggestion: `crates/sifr_frontend/src/structural_shape.rs:443` (`validated_adapter_value`) relies on `expect` plus `#[allow(clippy::expect_used)]`. The invariant holds (integers round-trip through canonical decimal text and nothing here is deserialized), but a `let ... else` skip would keep compiler code free of a panic path in a file whose inputs cross module boundaries.
- pre-existing issue: `crates/sifr_frontend/src/structural_shape.rs:560` still uses `{:?}` to encode enum variant values inside the canonical structural identity, so the Debug-derived-encoding cleanup is not fully complete outside callable/type identities.
- suggestion: the diagnostic text at `crates/sifr_frontend/src/early_adapters.rs:265` reads "adapter plan must contain exactly fields, metadata, …" — a word is missing after "exactly".

yaseralnajjar:
Create-PR gate evidence for exact candidate 28cb499ba96cabd65e31c7577927911eef9d6648

The one allowed create-PR gate was run with scripts/run_all_tests.sh --profile create-pr. It stopped in coverage_matrix_checks at the verification-taxonomy checker before M3-specific lanes. The checker rejected the separately owned pre_v1 compatibility inventory/checker and ordinary source text in M2/M3 files (including contract, fixture, M3, and architecture phase links). Coverage-matrix readiness, profile assignment, and taxonomy self-tests passed; only the taxonomy area-check failed. Per phase rules, this gate will not be rerun on the unchanged candidate.

yaseralnajjar:
Merge-gate evidence for exact candidate 28cb499ba96cabd65e31c7577927911eef9d6648

The one allowed merge gate was run with scripts/run_all_tests.sh. It stopped in coverage_matrix_checks at the same verification-taxonomy checker before M3-specific lanes. The checker rejected the separately owned pre_v1 compatibility inventory/checker and ordinary source text in M2/M3 files. Coverage-matrix readiness, profile assignment, and taxonomy self-tests passed; only the taxonomy area-check failed. Per phase rules, there will be no second merge gate.

PR #3254, opened 2026-08-18
Remove text and data stdlib compatibility names

yaseralnajjar:
## Summary
- remove legacy text and structured-data stdlib aliases while preserving canonical behavior
- migrate repository consumers, docs, parity fixtures, and generated demo baselines
- keep private implementation imports private and fix user-defined `filter` shadowing in codegen

## Validation
- focused Item 4 E2E manifest: 27/27 pass
- affected demos and idiomatic companions: pass
- stdlib/compiler package tests: pass
- installed sysroot boundary equivalence: pass
- `cargo test -p sifr -- --skip test_e2e_pass --skip test_e2e_fail`: pass
- formatting, file-size, HIR, compatibility inventory, module parity, and complexity guardrails: pass
- create-PR profile stopped at the inherited verification-taxonomy failure; Item 4-owned checks preceding it passed

Candidate: `b98864c82c61f42909b3395aae7714cc53215937`
Base: `eb8a58c812a32c199f58c08081caca4b3f8c349c`


yaseralnajjar:
**Verdict: SATISFIED**

Blocking findings:
- none

Verified against the acceptance criteria:
- All nine inventory surfaces owned by `pre_v1_compat_4_text_data_names` (`stdlib-regex-helper-prefixes`, `-json-loads`, `-json-dumps`, `-toml-loads`, `-base64-names`, `-fnmatch-filter`, `-html-intrinsics`, `-calendar-intrinsics`, `-url-name-pairs`) are removed and each retains only the canonical spelling from the inventory's `canonical_contract`.
- `dumps` (`stdlib/sifr/json.sifr:410`) keeps the full approved `json_dumps` input union (`JsonValue | bool | int | float | str | decimal | bigdecimal`); `decimal`/`bigdecimal` remain covered by `decimal_conversions.sifr` / `decimal_runtime_operations.sifr`.
- Regex keyword conflict resolves to one spelling: `Pattern.is_match` (`stdlib/sifr/re.sifr:162`, documented in the adjacent comment), exercised by `cpython_re_subset.sifr:97` and `stdlib_re_consolidated.sifr:64`.
- No implementation import re-exports an old name: every `_sifr.*` import in re/json/tomllib/base64/html/calendar/url is `as _*_impl` (or `_CompiledPattern`), and `bootstrap_tests.rs:417` plus the three `stateless_private_codegen_tests.rs` assertions now assert absence rather than presence.
- Repo-wide grep for the removed names over `stdlib/`, `crates/sifr/tests/e2e/`, `verification/`, `demos/*/main.sifr|idiomatic.rs`, `docs/`, `internal_docs/` is clean; remaining hits are `sifr_stdlib`/`_sifr.*` Rust-side intrinsic names (out of scope) and archived plan records.
- Compiler change is correctly narrow: dropping the `func == "filter"` name check in `leaves_and_plain_calls.rs:730` only affects plain `Call` nodes; builtin `filter` still lowers via `HirIteratorOp::Filter` (`leaves_and_plain_calls.rs:535` → `iterators_and_callables.rs:297`), which is the only shape `lower_filter_call` produces. New test at `classes_and_basics_codegen_tests.rs:4` pins the shadowing fix.

Follow-up findings:
- suggestion (introduced by this diff): `crates/sifr/tests/e2e/pass/cpython_fnmatch.sifr:2` and `crates/sifr/tests/e2e/pass/cpython_re.sifr:2-3` had their header comments mechanically renamed and are now self-contradictory ("filter is used instead of the CPython alias 'filter'…", "search has no alias (match is a Python keyword), so it stays as search"). Behavior is unaffected; the text should be rewritten to point at `Pattern.is_match`.
- suggestion: the same `has_match(pattern, text)` helper is hand-duplicated in four fixtures (`cpython_re.sifr`, `cpython_re_subset.sifr`, `error_subclass_handling.sifr`, `parse_safety_error_paths.sifr`, `panic_free_stdlib_errors.sifr`) instead of using the documented `compile(...).is_match(...)` spelling. Equivalent semantics (both unanchored), but the canonical spelling would get the consumer coverage.
- suggestion: `stdlib/sifr/json.sifr` keeps a private `_loads_impl` with public `loads` delegating to it, while `tomllib.sifr` has internal callers use the public `loads` directly. One of the two shapes is redundant indirection.
- pre-existing issue: `map` is still resolved by name in `try_lower_simple_call_expr` (`leaves_and_plain_calls.rs:730`), so a user-defined/imported `map` retains the shadowing bug that was just fixed for `filter`. Out of this item's stated compiler scope.
- pre-existing issue: cross-module unaliased imports still leak duplicate public names — `sifr.base64` re-exports `encode_utf8`/`decode_utf8`/`bytes_to_hex`/`bytes_from_hex` (`base64.sifr:16`), `sifr.tomllib` re-exports `read_text`/`FileHandle`, `sifr.json` re-exports `FileHandle`. Not old names, and bytes surfaces belong to `pre_v1_compat_5_binary_hashing`.
- infrastructure issue: as recorded in the evidence, the create-PR gate stopped on the inherited verification-taxonomy checker, and aggregate stdlib parity carries pre-existing failures (bare math namespace demo references, `audit_fixtures.py` import setup). The full E2E pass suite therefore was not observed green end-to-end for this candidate; the focused 27/27 plus `sifr_codegen` 1023 cover the changed surfaces.

PR #3255, opened 2026-08-18
M4: normalize adapter defaults, inheritance, and identity

yaseralnajjar:
## Summary
- normalize required, constant, and checked factory defaults across the provisional/final adapter boundary
- flatten nested Annotated descriptors, preserve inherited declaration identities, and substitute concrete generic parents
- separate adapter invocation/post-adapter cache identities and bind specialization identity to the finalized adapter output
- expand non-Pydantic native fixture coverage, including independent mutable factories

## Validation
- focused adapter/default, early-adapter, typed-descriptor, and special-base tests
- cargo test -p sifr_driver --lib (499 passed, 76 ignored)
- cargo test -p sifr -- --skip test_e2e_pass (only pre_v1-owned protocol_bound_unknown_forwarded_typevar failure)
- native static_class_adapter fixture run
- cargo fmt --check
- HIR maintainability and 900-line file-size guardrails

Exact candidate: fc06fb578f3cdc67c45143652d05d807cdb502e8

### DIFF
--- a/crates/sifr_driver/src/tests/adapter_defaults.rs
+++ b/crates/sifr_driver/src/tests/adapter_defaults.rs
@@ -120,8 +120,8 @@ class Model(Contract):
 
 #[test]
 fn adapter_identities_ignore_source_movement_and_distinguish_default_states() {
-    fn identities(source: &str) -> ([u8; 32], [u8; 32]) {
-        let compiled = compile(source);
+    fn identities(source: &str, contract: &str) -> ([u8; 32], [u8; 32]) {
+        let compiled = compile_with_contract(source, contract);
         let selection = compiled
             .external_defs
             .class_adapter_selections
@@ -134,13 +134,12 @@ fn adapter_identities_ignore_source_movement_and_distinguish_default_states() {
         )
     }
 
-    let constant = identities(
-        r#"
+    let constant_source = r#"
 from fixture.defaults import Contract, contract_field
 class Model(Contract):
     tags: list[str] = contract_field(default=[])
-"#,
-    );
+"#;
+    let constant = identities(constant_source, CONTRACT);
     let moved = identities(
         r#"
 
@@ -150,6 +149,7 @@ from fixture.defaults import Contract, contract_field
 class Model(Contract):
     tags: list[str] = contract_field(default=[])
 "#,
+        CONTRACT,
     );
     let factory = identities(
         r#"
@@ -157,8 +157,13 @@ from fixture.defaults import Contract, contract_field
 class Model(Contract):
     tags: list[str] = contract_field(default_factory=list)
 "#,
+        CONTRACT,
     );
     assert_eq!(constant, moved);
+    assert_eq!(
+        constant,
+        identities(constant_source, &format!("\n\n{CONTRACT}"))
+    );
     assert_ne!(constant, factory);
 }
 
@@ -287,6 +292,30 @@ class Child(Parent):
         error.code == sifr_diagnostics::DiagnosticCode::TYPE_MISMATCH.code()
             && error.message.contains("cannot be re-annotated")
     }));
+}
+
+#[test]
+fn compatible_inherited_reannotation_preserves_local_default_ordering() {
+    let compiled = compile(
+        r#"
+from fixture.defaults import Contract, contract_field
+
+class Parent(Contract):
+    value: int
+
+class Child(Parent):
+    value: int
+    enabled: bool = contract_field(default=True)
+"#,
+    );
+    let defaults = compiled
+        .external_defs
+        .function_defaults
+        .get("main")
+        .and_then(|functions| functions.get("Child"))
+        .expect("compatible override keeps the local default on enabled");
+    assert_eq!(defaults.len(), 1);
+    assert_eq!(defaults[0].0, 1);
 }
 
 #[test]
--- a/crates/sifr_driver/src/tests/adapter_defaults.rs
+++ b/crates/sifr_driver/src/tests/adapter_defaults.rs
@@ -1,0 +1,415 @@
+use super::support::parse_suite;
+use crate::project::ProjectLowering;
+use crate::{collect_project_hir_modules, compile_stdlib};
+use sifr_ir::{AdapterFieldDefault, StaticProgramValue};
+use sifr_type_system::Type;
+use std::collections::HashMap;
+
+const CONTRACT: &str = r#"
+from sifr.meta import CallableIdentity, DeclarationInput, DeclarationPlan, PlannedField, PlannedIssue, PlannedMetadata, StaticValue
+
+class DefaultDescriptor:
+    default_kind: str
+    default_value: StaticValue | None
+    default_factory: CallableIdentity | None
+
+@class_adapter_provider("fixture.defaults", "DefaultDescriptor")
+@const_eval
+def adapt_defaults(declaration: DeclarationInput[DefaultDescriptor]) -> DeclarationPlan[DefaultDescriptor]:
+    fields: list[PlannedField] = []
+    metadata: list[PlannedMetadata[DefaultDescriptor]] = []
+    issues: list[PlannedIssue[DefaultDescriptor]] = []
+    for item in declaration.declaration.items:
+        if item.kind == "field":
+            declared_type: str | None = item.declared_type
+            if declared_type is not None:
+                identity: str = item.identity
+                default_kind: str = item.default_kind
+                default_value: StaticValue | None = item.default_value
+                default_factory: CallableIdentity | None = None
+                for descriptor in declaration.descriptors:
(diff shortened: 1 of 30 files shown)


yaseralnajjar:
I inspected the frozen diff and surrounding implementation (`early_adapters.rs`, `early_adapters/inheritance.rs`, `class_type_collection.rs`, `adapter_field_plans.rs`, identity plumbing, stdlib contract, fixture, tests).

Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_lowering/src/lower/classes/class_type_collection.rs:746-750`, criterion "Build constructor parameters from normalized field states" / "An incompatible inherited field override is rejected" (i.e. a *compatible* override must be accepted). After the plan is applied, `own_field_default_indices` is rebuilt as `index.checked_sub(inherited_field_count)`, which assumes `own_fields[i]` corresponds to `fields[i + inherited_field_count]`. That assumption is broken by the new override path at lines 392-425: a local re-annotation of an inherited field pushes into `own_fields` but does **not** push into `fields` (`if inherited_override.is_none()`). Failure scenario: adapted `class Parent(Contract): value: int` and `class Child(Parent): value: int` plus `enabled: bool = contract_field(default=True)`. `fields = [value, enabled]`, `inherited_field_count = 1`, `own_fields = [value, enabled]`. The plan's const default sits at `fields` index 1, so the remap yields own index `0` → `own_fields[0]` (`value`) is marked as defaulted and `enabled` is then treated as a required field after a default, emitting `CLASS_REQUIRED_FIELD_AFTER_DEFAULT` at line 787 and rejecting valid code (the same misalignment also silences genuine ordering violations for such classes). Correction: record the `fields`-index → `own_fields`-index mapping while collecting annotations (the loop already knows both `field_idx` and `own_field_idx`) and remap planned defaults through that mapping instead of subtracting `inherited_field_count`. The same off-by-one from the override also mis-derives `inherited_count` in `early_adapters/inheritance.rs:569` (`class.fields.len().saturating_sub(local_field_count)`) on the no-`parent_type` fallback path.
- in-scope omission: `crates/sifr_frontend/src/early_adapters.rs:113` (feeding lines 114-123 and 138-145), criteria "Exclude source locations … from shape identity" and "Unrelated source movement does not invalidate the schema program". The adapter-invocation key hashes `format!("{functions:#?}")` — a `Debug` dump of the provider's const-eval HIR, which retains source byte offsets: `HirExpr::MethodCall.source: MethodCallSource { call_range, receiver_range, arg_ranges }` (`crates/sifr_ir/src/hir_nodes.rs:729`, populated from `call.range()` in `crates/sifr_lowering/src/lower/expressions/method_calls.rs:318`), plus `IntrinsicCall`/`PythonRecordExpansion` spans. Failure scenario: prepend a blank line or comment to the provider module (e.g. the `adapt_contract`/`adapt_defaults` module, whose body contains `fields.append(...)`) — every span shifts, so `provider_hir` changes, which changes `adapter_invocation_identity`, `post_adapter_identity`, and therefore the `("adapter", …)` input to `static_program_identity` in `specialization_runner.rs:179`, invalidating the schema program for a semantically identical adapter. The two new identity tests only move *consumer* source (`adapter_defaults.rs:144-153`, `early_adapters.rs:500-518`), so this is untested. Correction: key the provider on a location-free canonical encoding (as `const_canonical`/`canonical_types` do for values and types) rather than a `Debug` dump — this is also the pattern the M3 follow-up "replace remaining Debug-derived canonical identity" targets, and add a provider-side source-movement case to the identity test.

Follow-up findings:
- pre-existing issue: an adapted parent's const/factory default at an inherited index now yields a constructor with a defaulted parameter ahead of a required local one, and the `checked_sub` filter drops it from `own_field_default_indices`, so the required-after-default check no longer sees it. Worth an explicit decision on dataclass-style inheritance ordering rather than leaving it unchecked.
- infrastructure issue: the validation note says the native `static_class_adapter` fixture covers "inheritance, and generic parent substitution", but `verification/areas/core_language/fixtures/static_class_adapter/app.sifr` only exercises defaults/factories (local, static-method, and imported callables); inheritance and concrete generic substitution are covered by `sifr_driver` unit tests only. Also `fixture/contract_types.sifr` is now stale (2-field `ContractDescriptor`) and unused by the fixture while still being loaded by `adapter_defaults.rs:296`.
- suggestion: `inheritance::declaration_value` is recomputed from scratch in `adapter_input`, `validate_fields`, and `parse_metadata` (once per metadata entry); computing it once per `run_one` would remove the repeated parent/plan resolution work.

yaseralnajjar:
**Verdict: SATISFIED**

Both prior blocking findings are corrected:

1. **Index remapping** — `class_type_collection.rs:387,425,749` now records an explicit `own_field_index_by_field_index` map at annotation time (`field_idx` = flattened index, including the `inherited_override` case) and derives `own_field_default_indices` from it, so a compatible local reannotation no longer shifts default ownership. `class_declarations.rs:230-262` resolves the field through `parent_type` first and otherwise offsets the local index by the parent field count — consistent with `class_body_lowering.rs:400-411`, where `HirClass::fields` is `all_fields` minus parent names, and with the flattened keying of `class_field_defaults`. `early_adapters/inheritance.rs:50-54,135-171` replaces the subtraction fallback with a resolved local or uniquely imported parent shape. Regression test `compatible_inherited_reannotation_preserves_local_default_ordering` asserts default index 1 for the local field.
2. **Provider identity** — `early_adapters.rs:113` now uses `adapter_program_identity::canonical_const_functions`, which clones the module and recursively strips the only source-relative data actually reachable in the provider HIR: `MethodCallSource` (`*source = None`), `IntrinsicCall.call_range`/`arg_ranges`, `PythonRecordExpansion.span`, plus interop declarations; statements, nested functions, patterns, comprehension generators, lambda/param defaults and async-with contexts are all covered. `const_functions` builds the module with empty `classes`/`constants`, so restricting the walk to `functions` is complete for this input. Coverage now prepends whitespace to the provider module (`adapter_defaults.rs:152-156`).

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `BindingId`s are allocated from one module-wide counter (`scope.rs:139,341`, single `Scope` per `LowerCtx` at `mod_context.rs:265`) and are retained unstripped in `HirExpr::Name.binding_id`. Canonical provider HIR therefore still changes when declarations are added or reordered ahead of the adapter in the provider module; the new regression only exercises whitespace-level provider movement, so the "provider movement does not invalidate" guarantee holds for span shifts but not declaration-order shifts.
- pre-existing issue: a compatible-but-unequal inherited reannotation (mutually assignable yet not `==`, e.g. an alias over the parent's base type — `type_rendering.rs:459-485`) is normalized through `parent_type` for the declaration/validation path only. `class_type_collection.rs` still stores the local type in the finalized class, so `adapter_field_plans::defaults_for_class` hits `plan.declared_type != *ty` and emits a `META_MALFORMED_DECLARATION` "normalized field contract changed before finalization" for valid code. Normalizing the finalized field type through `parent_type` would close it.
- suggestion: `fallback_parent_shape` (`early_adapters/inheritance.rs:135-171`) reads local parents from `result.module.classes`, whose `fields` are own-fields only, and rejects a parent name that resolves in more than one external module without consulting the importing module's bindings. Both are narrow given the single-adapted-parent scope, but the ambiguity case turns resolvable code into a hard `adapted data parent shape is unavailable` error.
- suggestion: `strip_function_locations` clears `rust_interop`/`python_interop` wholesale, so two provider functions differing only in interop declarations share an invocation identity. The comment justifies this for deterministic const evaluation; a span-only scrub would preserve the semantic distinction.

yaseralnajjar:
Create-PR gate evidence for exact candidate `ab9047d8cdf3a297d6af19a8ac143165e7d61205`.

Command: `scripts/run_all_tests.sh --profile create-pr`

Result: stopped in `coverage_matrix_checks` at the verification-taxonomy checker (26.55s wall time). The checker rejected `verification/compatibility/pre_v1_compatibility_inventory.json`, its checker, and broad matches including existing M4 fixture/test/documentation vocabulary. This taxonomy mechanism and the pre_v1 inventory are owned by the separate pre_v1 session and are explicitly out of scope for M4. Coverage-matrix readiness, profile assignment, and negative self-tests passed before the taxonomy case failed.

Per phase rules, this one-shot gate will not be rerun on the unchanged candidate.

yaseralnajjar:
Merge gate evidence for exact candidate `ab9047d8cdf3a297d6af19a8ac143165e7d61205`.

Command: `scripts/run_all_tests.sh`

Result: stopped in `coverage_matrix_checks` at the same externally owned verification-taxonomy checker (6.09s wall time). The failure is the separate pre_v1 inventory/checker plus broad false-positive matches in M4 and existing documentation; coverage-matrix readiness, profile assignment, and negative self-tests passed.

This was the one permitted merge-gate run. It will not be repeated on this candidate.

PR #3258, opened 2026-08-18
M5: add handler-bearing method descriptors

yaseralnajjar:
## Scope
- resolve descriptor-planned user methods into typed static method slots
- support static, class, shared, mutable, and owned receivers, including infallible and `Result` handlers
- preserve descriptor identity, source origin, and deterministic inherited/local order
- enforce `Self` and owned-receiver move semantics
- update runtime coverage and specialization documentation

## Validation
- `cargo test -p sifr_driver handler_descriptors`
- `cargo test -p sifr_lowering -p sifr_frontend -p sifr_codegen -p sifr_driver --lib -- --skip test_e2e_pass --skip test_unknown_parent_class_has_class_code`
- `cargo test -p sifr_driver test_method_slot_runtime -- --ignored --nocapture`
- `cargo test -p sifr_driver test_method_slot_lifetime_thread_and_shared_context_rejections -- --ignored --nocapture`
- `cargo test -p sifr_codegen receiver`
- `cargo fmt --check`
- `python3 scripts/check_docs_error_code_links.py`
- `git diff --check`

## Deferred existing phase work
The M4-owned unknown-parent diagnostic mismatch remains excluded from M5 validation and is recorded for the later phase closure item. The separate `pre_v1` session/worktree is wholly out of scope.

Candidate: `2cdc8380ccc710f7a1dc86e9b5bbeedfa07d9d66`
Base: `4f3a475428b6c9ca789acd76212d6619f2e32d6b`

### DIFF
--- a/crates/sifr_codegen/src/lib_codegen_tests/receiver_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/receiver_codegen_tests.rs
@@ -106,7 +106,7 @@ class Consumable:
 
     assert!(rust_code.contains("fn read(&self) -> i64"));
     assert!(rust_code.contains("fn bump(&mut self)"));
-    assert!(rust_code.contains("fn close(&self)"));
+    assert!(rust_code.contains("fn close(self)"));
 }
 
 #[test]
--- a/crates/sifr_codegen/src/static_program_codegen.rs
+++ b/crates/sifr_codegen/src/static_program_codegen.rs
@@ -339,6 +339,12 @@ mod tests {
             output_type: Type::Str,
             context_type: None,
             context_mutable: false,
+            descriptor_type: None,
+            descriptor_value: None,
+            descriptor_origin: None,
+            descriptor_range: None,
+            declaration_order: None,
+            is_fallible: true,
         });
         output.method_slot_context = Some(StaticMethodSlotContext::None);
 
--- a/crates/sifr_codegen/src/static_program_slots_codegen.rs
+++ b/crates/sifr_codegen/src/static_program_slots_codegen.rs
@@ -108,8 +108,17 @@ fn slot_arm(index: usize, slot: &StaticMethodSlot, context: &StaticMethodSlotCon
                 .is_some_and(|param| param.convention.is_mut_borrow()));
     let binding = if mutable_input { "mut " } else { "" };
     let call = slot_call_expression(slot, context);
+    let dispatch = if slot.is_fallible {
+        format!(
+            "match {call} {{\n                    Ok(output) => {STRUCTURAL}::StructuralProject::structural_project(&output, &mut {STRUCTURAL}::SlotSinkVisitor::new(sink)).map_err({STRUCTURAL}::SlotError::Contract),\n                    Err(error) => Err({STRUCTURAL}::SlotError::Slot(error.to_string())),\n                }}"
+        )
+    } else {
+        format!(
+            "let output = {call};\n                {STRUCTURAL}::StructuralProject::structural_project(&output, &mut {STRUCTURAL}::SlotSinkVisitor::new(sink)).map_err({STRUCTURAL}::SlotError::Contract)"
+        )
+    };
     format!(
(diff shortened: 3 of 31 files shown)


yaseralnajjar:
Verdict: SATISFIED

Reviewed at base `4f3a475` → candidate `2cdc838` (read-only; no builds, no file changes). Traced each scope line and acceptance criterion through the diff:

- Descriptor→checked-method resolution (`typed_descriptors.rs:341-370`, `callable_identities.rs:62-95`), sealed `CallableIdentity` carried into `AdapterHandlerPlan` and validated against the re-derived signature (`slot_table.rs:458-492`).
- Slots extended with descriptor type/value/origin/range/order and `is_fallible` (`specialization_metadata.rs:314-326`); infallible and `Result` handlers both dispatch through checked signatures (`static_program_slots_codegen.rs:108-124`).
- Receiver coverage: static, class (classmethod rejection removed), shared, mutable, owned; owned must return exactly `Self`/`Result[Self, E]` (`slot_table.rs:660-672`); constructors still rejected (`slot_table.rs:592`); async and non-borrowed context still rejected.
- `own self` now lowers to `ReceiverConvention::Owned` (`parameter_conventions.rs:17`) with move/use-after-move/borrowed-parameter-escape checks at call sites; `Self` annotation resolves to the current class specialization with `current_class` correctly saved/restored during class-type collection.
- Order determinism: inherited handlers precede local ones by `declaration_order`, and shape handlers are sorted by that order before the specializer selects them.
- Identity/caches: handler `CallableIdentity` enters `canonical_value` and the descriptor/target enter `canonical_method`, both feeding `static_program_identity` (`specialization_runner.rs:171-181`); the driver test asserts identity changes with the selected target.
- Non-Pydantic pipeline fixture plus typed-error path added; driver tests cover all listed rejection and ordering criteria.

No regression or in-scope omission found.

Blocking findings:
- none

Follow-up findings:
- suggestion: `consume_declared_owned_receiver` (`crates/sifr_lowering/src/lower/expressions/method_calls.rs:342`) only tracks `HirExpr::Name` receivers, so `self.child.finish()` on a declared `own self` method gets no move tracking and no diagnostic. The Rust-opaque cleanup path rejects field/temporary receivers explicitly (`method_calls.rs:243`); the declared-owned path should get the same guard so the ownership story is uniform.
- suggestion: the `Self` diagnostic says "valid only in an ordinary class method annotation" (`annotation_resolution.rs:21`) but `current_class` is set for classmethods and staticmethods too, so `Self` is silently accepted there. Either accept it deliberately or restrict it to match the message.
- suggestion: the stacked-form check only runs when `@classmethod` is present (`typed_descriptors.rs:185`); a method descriptor stacked with `@staticmethod` has no order requirement (the new tests and fixtures put the descriptor inside `@staticmethod`, the opposite nesting from the classmethod rule). Worth stating the intended rule for staticmethod explicitly.
- suggestion: adapter selections are matched by bare class name in `slot_table.rs:366` and `structural_shape/methods.rs:140` rather than by class identity; an imported class sharing a name with a locally adapted class could match the wrong selection.
- suggestion: `SlotIdentitySignature` still emits `handler: None` and omits fallibility, so a `str` vs `Result[str, E]` handler differs only via the enclosing program identity. That is currently sufficient, but recording fallibility in the slot identity would make the table contract self-describing.

yaseralnajjar:
Gate evidence for exact candidate `2cdc8380ccc710f7a1dc86e9b5bbeedfa07d9d66`:\n\n- create-PR gate: attempted once; stopped in `coverage_matrix/readiness/verification_taxonomy` on the separate out-of-scope `pre_v1` verification mechanism.\n- merge gate: attempted once; stopped at the identical out-of-scope `pre_v1` taxonomy failure.\n- no gate reruns were performed.\n- M5 scoped validation and exact-SHA Opus review remain green.\n\nThe external taxonomy mechanism scans broad existing vocabulary across its own inventory and unchanged compiler/docs/tests; it is owned by the parallel `pre_v1` worktree/session and is not modified here.

PR #3260, opened 2026-08-18
Remove binary and hashing compatibility helpers

yaseralnajjar:
Phase item: pre_v1_compat_5_binary_hashing. Makes first-class bytes and bytes-native hashing authoritative, removes the sifr.bytes transition module and stale hash aliases, migrates binary consumers, and updates generated demos and boundary evidence. Candidate SHA: 3d34ad6289096f7588a75dda8689eb19b4f1ad62.


yaseralnajjar:
## Review — `pre_v1_compat_5_binary_hashing`

Diff reviewed: `c570aec5..3d34ad62` (exact SHAs), read-only.

### What I validated locally

| Check | Result |
|---|---|
| `cargo test -p sifr_driver / sifr_codegen / sifr_lowering / sifr_stdlib_manifest / sifr_stdlib` | pass (incl. new `binary_and_hashing_exports_use_only_first_class_bytes_contracts`) |
| `cargo test -p sifr -- --skip test_e2e_pass` | 37 pass / 1 fail — the fail is the separately owned pre-existing `protocol_bound_unknown_forwarded_typevar.sifr` mismatch (recorded in `plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md:532`), not a regression here |
| 4 modified `fail` fixtures via `sifr check` | all still reject (the fail suite aborts before reaching `sha3_object_model_unsupported.sifr`, so I checked them by hand) |
| 20 modified `pass` fixtures + 16 migrated demos via `sifr run` | all pass |
| `emitted.rs` reproducibility (`hashlib`, `bytes_module`, `binary_storage`, `safe_edge_cases`, untouched `logging`) | byte-identical to current `sifr emit` |
| Boundary dependency plan for `stdlib_boundary_recertification.sifr` | `sifr_stdlib` with no features → matches the updated snapshot |
| `check_stdlib_module_parity.py --scope full`, `check_stdlib_manifest_schema.py`, `check_file_size_guardrails.py` (PASS, 3145 files), HIR guardrails, `cargo fmt --check`, `cargo clippy --workspace -D warnings` | all pass |
| Repo-wide grep for `sifr.bytes`, `encode_utf8`, `bytes_to_hex`, `new_bytes`, `update_bytes`, `digest_bytes`, `*_obj` | only archived `plans/` history and the negative assertions in `bootstrap_tests.rs` |

The core of the item is done well: `_sifr.bytes` and `sifr.bytes` are gone, `sifr_stdlib::bytes` and its cargo feature are gone, the string hash entrypoints (`hash::sha256(s)` … `blake2s(s)`) are deleted from both `_sifr/crypto.sifr` and `crates/sifr_stdlib/src/hash.rs`, `HashObject.update`/`new`/named constructors are bytes-only, `file_digest` now reads binary instead of text, `gzip` moved off `list[int]`, and search/prefix/hex ownership moved to first-class `bytes` methods with typed diagnostics (verified: `bytes.startswith() argument must be 'bytes', got 'str'`, `bytes.hex() takes no arguments`).

### Blocking findings

**1. `bytes.hex()` lowering evaluates its receiver twice (`crates/sifr_codegen/src/methods/bytes.rs:275-326`).**
`lower_hex` clones `object` at line 292 (`.len()`) and again at line 305 (`.iter()`), with no temp binding. Verified with a real program:

```python
def make(mut self) -> bytes:      # increments self.calls
    ...
h: str = c.make().hex()           # prints calls == 2
```

Effectful receivers are silently executed twice. This is also a stdlib regression: `demos/hashlib/emitted.rs:629-638` shows

```rust
fn _hash_hex(algorithm: &String, data: &Vec<u8>) -> String {
    let mut __hex = String::with_capacity(_hash_bytes(algorithm, data).len().saturating_mul(2));
    for __byte in _hash_bytes(algorithm, data).iter() { ... }
```

so every `HashObject.hexdigest()` (`stdlib/sifr/hashlib.sifr:104`) now computes the digest twice, where the removed `bytes_to_hex_strict(_hash_bytes(...))` computed it once. Fix: bind the receiver to one `RustStmt::Let` and reference it for both `len()` and `iter()`. (Same root cause pre-exists in `lower_find` — measured 3× receiver evaluation — see follow-ups.)

**2. A legacy `list[int]` binary storage/conversion path survives in `sifr.io.BytesIO` (`stdlib/sifr/io.sifr:476-521`).**
`_buffer: list[int]`, `self._buffer = initial.to_ints()` (481), `bytes.from_ints(values)` on every read (498), `data.to_ints()` on every write (521). This is the only remaining consumer of `bytes.to_ints()` in the stdlib, and it contradicts required scope "Remove legacy `list[int]` conversion paths outside explicit construction" and acceptance 1 ("Typed binary APIs use `bytes` for storage and transport"). Public inputs are `bytes`, so acceptance 4 holds; acceptance 1 does not. Nothing in the diff, the issue item, or `verification/compatibility/pre_v1_compatibility_inventory.json` dispositions it — per the work-boundary rule it must either be migrated in this item or recorded as an owned out-of-scope failure.

### Non-blocking follow-ups

- `stdlib/sifr/hashlib.sifr:123-131`: `file_digest` skips `_file_close_impl` when `_file_read_bytes_impl` raises, leaking a runtime handle-table entry. No panic; typed `HashlibError` is still returned.
- `crates/sifr_codegen/src/methods/bytes.rs:143-251`: `lower_find` (renamed from `lower_index` here) evaluates the receiver 3× — same defect class as finding 1, pre-existing but worth fixing together.
- `crates/sifr/tests/e2e/fail/bytes_buffer_protocol_unsupported.sifr` and `bytes_implicit_str_bytes_coercion_unsupported.sifr` now assert "no method `decode`" on a class/`str` rather than the buffer-protocol / implicit `str`→`bytes` coercion the `Reference:` ids name. Implicit-coercion rejection is still covered by `bytesio_text_write_unsupported.sifr`, but the fixture intent drifted from its parity anchor.
- No fail fixture covers the new `startswith`/`endswith`/`hex` diagnostics (I verified them manually); only codegen unit tests were added in `methods/mod.rs`.
- `demos/random_hashing/main.sifr:27` hardcodes `b"random_hashing_seed"` instead of encoding the existing `payload` — the demo no longer demonstrates the explicit-encoding contract it exercises.
- `scripts/check_stdlib_manifest_schema.py:46` adds `sifr.bytes::primitive_constructors` to `PLANNED_REARCHITECTURE_DELETIONS` for what is really a *rename* to `first-class-bytes::primitive-construction`; the allowlist entry is permanent slack in the transition guard.
- Item 5's acceptance checkboxes in `plans/issues/active/ad-hoc-pre-v1-compatibility-removal.md:432-437` are still unchecked and no record entry exists — consistent with this repo's separate "Record …" commit pattern, so noted only.
- Very large `emitted.rs` churn in demos whose `main.sifr` is untouched (`logging`, `shutil`, `tempfile`, …) is stale-artifact refresh, not a defect: I confirmed the regenerated files match current compiler output exactly.
- Housekeeping: this worktree's private `target/` is 23 GiB (over the 20 GiB guardrail). I left it alone since the review is read-only.

Acceptance status: 2 ✅, 3 ✅, 4 ✅, 5 ✅, 6 ✅ (no `unwrap`/`expect`/`panic!` in the affected generated paths; error paths return typed `ParseError`/`ValueError`/`IOError`), 7 ✅ (guardrail PASS). 1 ✗ (finding 2). Correctness ✗ (finding 1).

VERDICT: NOT SATISFIED

yaseralnajjar:
## Remediation review — `pre_v1_compat_5_binary_hashing`

Scope inspected: `git diff c570aec5..0d6b6519` (full item), with the remediation commit `0d6b6519` reviewed in isolation. Read-only; no files modified. I did not re-run the gates — I relied on the stated validation and verified the mechanisms in source and in the committed emitted output.

### Blocker 1 — receiver evaluated twice: FIXED

- `bind_receiver_once` (`crates/sifr_codegen/src/methods/bytes.rs:80`) binds `let __bytes_receiver = &<object>;` as the first statement of the lowered block; `count`, `contains`, `find`, and `hex` now reference `bound_receiver()` at every use site. `&<temp>` in a `let` gets temporary-lifetime extension, so an rvalue receiver stays alive for the block.
- The regression is confirmed gone at the site that mattered: `demos/binary_hashing/emitted.rs` previously emitted `_hash_bytes(algorithm, data)` twice inside `_hash_hex` (double hashing per `hexdigest()`); it now emits `let __bytes_receiver = &_hash_bytes(algorithm, data);` once (`demos/binary_hashing/emitted.rs:652`).
- Evaluation order improved as a side effect: receiver now precedes the needle/bounds args, matching Python order.
- Nesting is safe: an inner bytes-method block shadows `__bytes_receiver` only within its own block, and `find`'s `__len`/`__bound` resolve innermost-first.
- Coverage: unit assertions pin single-occurrence for `contains`/`find`/`hex` (`crates/sifr_codegen/src/methods/mod.rs:796,809,829`), and `crates/sifr/tests/e2e/pass/bytes_helpers.sifr:26-30` asserts `factory.calls == 1` / `== 2` after `.hex()` and `.find(114)` on an effectful receiver. Constants check out (`b"bytes-rule0"` → `62797465732d72756c6530`, `'r'` at index 6).

### Blocker 2 — `BytesIO` list[int] storage: FIXED

- `stdlib/sifr/io.sifr:476` declares `_buffer: bytes`; `_slice_to_bytes` and both `to_ints()`/`from_ints` conversions are gone. `write_bytes` is now `left + data + right` slicing/concatenation, `getvalue` returns the buffer directly.
- Behavior preserved: `seek` clamps `next_pos` to `len(buffer)` (`io.sifr:539`), so `_cursor <= len` always holds and the `left`/`right` splits cannot produce a gap; overwrite-in-place plus extend semantics match the previous element loop. Emitted slicing is bounds-clamped and panic-free (`demos/in_memory_streams/emitted.rs:1610-1690`) — no indexing, `unwrap`, or `expect` on the new paths.
- The old `raise IOError("bytes write invariant violation")` dead path is removed rather than papered over.

### Acceptance criteria

| Criterion | Result |
|---|---|
| Typed binary APIs use `bytes` | Met — `BytesIO`, `sifr.gzip.compress/decompress`, hashlib all bytes; no `list[int]` binary transport remains in `stdlib/` (remaining `list[int]` signatures are tuple/handle payloads: fs, time, process, net, python). |
| Text reaches hashing only via explicit encoding | Met — string one-shots, `HashObject.update(str)`, `new(str)`, and `*_obj(str)` removed; `_sifr.crypto` string hash declarations and `sifr_stdlib::hash::{sha256,md5,…}` string fns deleted. Guarded by `bootstrap_tests.rs:595`. |
| No wrapper duplicates a first-class bytes op | Met — `stdlib/sifr/bytes.sifr`, `stdlib/_sifr/bytes.sifr`, `sifr_stdlib::bytes`, and the `bytes` cargo feature deleted; `base64.b16encode/decode` now use `str.encode`/`bytes.hex`/`bytes.from_hex`/`bytes.decode`. Only self-contained local helpers named `bytes_to_hex` remain in hand-written `idiomatic.rs` reference files. |
| No typed binary API takes `list[int]` canonically | Met — only the explicit named constructor `bytes.from_ints` accepts it (retained language glue, live in `sifr.random`). |
| Private binary intrinsics have a live consumer | Met — all eight `_sifr.crypto` `*_bytes` declarations are asserted live from `sifr.hashlib` (`stateless_crypto_codegen_tests.rs`), dead `_sifr.bytes` policy/codegen tests removed with the module, retained-intrinsics TOML entry re-owned. |
| No user-triggered panic in generated code | Met on the changed paths (clamped slices, `saturating_mul`, no `unwrap`/`expect`/indexing). |
| Files under 900 lines | Met — largest touched hand-maintained file is `sources.rs` at 842; `methods/mod.rs` 840, `io.sifr` 590, `bytes.rs` 396. |

### Non-blocking / later-item follow-ups (do not gate this item)

1. **`hashlib.file_digest` leaks the native handle on read failure** — new in this item (`stdlib/sifr/hashlib.sifr:123-127`): if `_file_read_bytes_impl` raises, control jumps to `except` and `_file_close_impl(handle)` never runs. The previous `read_text` path closed internally. Resource leak only; no panic.
2. **`BytesIO.write_bytes` is now O(n) per write with several full buffer clones** (slice lowering emits `&self._buffer.clone()` per slice, plus two concat clones), making sequential writes quadratic where the old element loop amortized appends. Behavior-preserving, throughput-regressing.
3. **Borrow-conflict shape newly reachable in generated code**: `__bytes_receiver` holds `&receiver` across evaluation of argument expressions, so an argument that mutably borrows the same place (e.g. `self._data.count(self.bump())` with `mut self`) would now be a rustc `E0502` rather than compiling. The `mutable_receiver_*_overlap` fail-fixture family suggests the frontend rejects most of this class first; worth an explicit fixture rather than relying on rustc.
4. **Minor coverage gaps**: the single-occurrence assertion exists for `contains`/`find`/`hex` but not `count` (implementation is correct — see `demos/binary_storage/emitted.rs:1822`), and the effectful-receiver e2e covers only `hex`/`find`, not `count`/`contains`.
5. **`BytesIO.getvalue` still returns `Result[bytes, IOError]` though now infallible**, and (pre-existing) does not raise on a closed stream unlike CPython.

Both original blockers are fully remediated with regression coverage, and no new blocking defect surfaced.

VERDICT: SATISFIED

yaseralnajjar:
Gate evidence for exact candidate 0d6b6519806afc86fd7c57d9b1472d96ff65db88: the one create-PR profile and one merge gate were each run once and each stopped in coverage-matrix readiness. Both reported the inherited verification-taxonomy failure assigned to Item 16 and the newly surfaced stale sifr_stdlib/bytes feature-classification row in verification/areas/coverage_matrix/data/cargo_metadata_classification.json, assigned to Item 7. Per phase rules, neither gate was rerun.

PR #3264, opened 2026-08-18
pre-v1 compatibility: remove collection helper APIs

yaseralnajjar:
## Summary
- replace list-backed set helpers with first-class generic set operations
- remove copy-returning heapq and bisect compatibility helpers
- delete the private collections Rust adapter and Cargo feature
- migrate fixtures, demos, docs, manifests, and generated companions

## Validation
Focused Item 6 validation is recorded in the phase update after merge. The exact candidate receives one Opus review and the prescribed one create-PR and one merge gate.


yaseralnajjar:
Exact-SHA Opus review for candidate 9d2cecd719dd619f172aa3bddfce5cfc2ad1367c

**Verdict: SATISFIED**

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr/tests/e2e/pass/cpython_collections_subset.sifr:123` and `crates/sifr/tests/e2e/pass/stdlib_collections_consolidated.sifr:205` build an empty set via `set([0])` followed by `.remove(0)`, while `cpython_collections.sifr:49` and `demos/python_regressions/main.sifr:257` use the direct `set()` form. The indirect form is unnecessary and reads as if empty-set construction were unsupported; normalize on `set()`.
- infrastructure issue: `demos/stdlib_ownership/emitted.rs` regeneration (646 insertions / 578 deletions) carries far more than the helper removals — `std::` → `::std::` paths, `Counter` → `__SifrStdlib_sifr_x2ecollections_x2eCounter` mangling, added `__iter__`/`__getitem__`, and duplicated `PartialOrd` bounds. That drift means the companion was already stale against live emission before this item, so the companion-freshness gate is evidently only exercised for demos touched by a change. Worth a standing all-companions check.
- pre-existing issue: the duplicated trait bound emitted as `T: Clone + ::std::fmt::Display + PartialOrd + 'static + PartialOrd` in the regenerated companions comes from codegen, not from this diff. Harmless to rustc, but it is redundant output.
- pre-existing issue: `verification/compatibility/pre_v1_compatibility_inventory.json:401,412,423` now list `verification/areas/stdlib_parity` as the sole consumer for the three removed rows, yet `stdlib_parity_matrix.md` no longer names any of those helpers. The checker (`check_pre_v1_compatibility_inventory.py:145-155`) requires a non-empty list of existing paths, so a fully-removed surface cannot record "no consumers." Same pattern as `pre_v1_compat_5`; the schema, not this item, is the constraint.
- pre-existing issue: `demos/generic_stdlib/main.sifr:9` imports `heapify`, `heappop`, and `heappush` without using any of them (only `nsmallest`/`nlargest` are called).
- pre-existing issue: the `protocol_bound_unknown_forwarded_typevar` diagnostic/CFG fail-fixture defect reported in the evidence is outside the changed paths and unrelated to this removal.

PR #3270, opened 2026-08-18
Remove remaining public stdlib alias residue

yaseralnajjar:
## Summary

- compile and publish an exact public stdlib API inventory
- remove remaining same-operation aliases and private implementation leaks
- harden private re-export, map-shadowing, signature collision, bytes I/O, and hashing ownership paths
- reconcile all generated demo companions and add a standing freshness checker

## Validation

- codegen: 1,027 passed
- frontend: 91 passed
- driver stdlib: 76 passed, 2 ignored
- installed sysroot and private re-export tests passed
- 411/411 algorithmic namespace fixtures passed
- module parity full/inventory/self-test passed
- 283/284 demo namespace fixtures passed; the package-scoped raw-Python demo is incorrectly invoked without its manifest by the Item 13-owned runner
- all generated demo companions are byte-exact and fresh
- Clippy, formatting, compatibility inventory, coverage matrix, sysroot certification, native-adapter reachability, file-size, and HIR guardrails passed
- one create-PR gate stopped at the inherited verification-taxonomy conflict; it was not rerun

## Review

Exact-SHA Opus review pending.


yaseralnajjar:
**Verdict: NOT SATISFIED**

Blocking findings:
- in-scope omission: `scripts/check_demo_emitted_freshness.py:1` (whole file), scope bullet "Complete the deferred Item 2-6 findings assigned to Item 7" (issue row `plans/issues/active/ad-hoc-pre-v1-compatibility-removal.md:1045`, which requires "add standing freshness evidence"). The new checker is not registered in `verification/policy/guardrails.json`, is not invoked by `verification/runner/sifr_verify/profile_runner.py`, and is referenced nowhere else in the repo (`grep -rI check_demo_emitted_freshness` returns only the file itself), so nothing in any profile, gate, or record ever runs it. The reconciliation of the 246 companions is therefore one-time evidence only, and the exact drift the finding was raised for (Item 3 row 1030, Item 6 row 1045) can silently return. I verified the tool itself is correct — running it on the candidate reports "all generated demo companions are fresh" — so the omission is purely the missing standing registration. Correction: add a guardrail entry (name/entrypoint/args/timeout/report) for the script in `verification/policy/guardrails.json` following the `file-size` and `stdlib-complexity-resource` rows, or wire it into the parity/demo area steps, so the check runs in the same gates that already run the file-size and HIR guards.
- in-scope omission: `plans/phases/06_stdlib_architecture.md:26` (and 37 further lines in the same file; also `plans/phases/13_type_system_completion.md:64,575`), same scope bullet, deferred row `plans/issues/active/ad-hoc-pre-v1-compatibility-removal.md:1029` ("Current stdlib **architecture** and parity records still name the obsolete `lib/sifr/*.sifr` source path"). The candidate corrected the parity side only — `plans/phases/07_stdlib_parity.md:275` plus the `verification/areas/stdlib_parity/reports/*` traceability files — while the stdlib architecture record still names `lib/sifr/math.sifr`, `lib/sifr/bisect.sifr`, `lib/sifr/statistics.sifr`, etc., in 38 lines. Item 4 already migrated some lines in that same file (`stdlib/sifr/json.sifr`, `stdlib/sifr/re.sifr` at lines 73 and 76), so the record is maintained as current, not frozen history. Correction: rewrite the obsolete source-path references in `plans/phases/06_stdlib_architecture.md` (and the two `lib/sifr/` audit references in `plans/phases/13_type_system_completion.md`) to `stdlib/sifr/`, leaving the legitimate installed-sysroot paths (`<sysroot>/lib/sifr/stdlib/**` in `internal_docs/sifr_sysroot_and_stdlib_architecture.md`, `crates/sifr_sysroot/src/tests.rs`, `scripts/distribution/verify_release_archive.py`) untouched.

The rest of the item checks out on inspection: the alias removals in `stdlib/sifr/{base64,bisect,fnmatch,statistics}.sifr` and `stdlib/_sifr/math.sifr` are same-operation only and every consumer (stdlib, demos, e2e fixtures, docs) was migrated with no dangling references; the imported-alias guard (`crates/sifr_driver/src/private_re_exports.rs` + the `_sifr.`-filter in `stdlib/re_exports.rs:20`) plus the new `bootstrap.rs:250-256` retains genuinely strip class/error/type-param/constant re-exports, and `should_export_stdlib_declaration` now has one owner in `sifr_type_system`; `url` `parse`/`build` and the keyword-conflict names (`move_file`, `mod_val`, `search`/`is_match`) are already canonical; `docs/stdlib/public-api.mdx` is byte-pinned by `compiled_stdlib_exports_match_public_reference` and contains no implementation-name or alias residue; the deferred hashlib close-on-error, `BytesIO` append fast path and `getvalue -> bytes` shape, `map` iterator-HIR routing with shadowing coverage, collision-safe private signature registration, generic-bound dedup, `time()` evidence token, and the bare-`math` demo residue are all present. I independently confirmed the demo mechanism: all 246 companions byte-match the candidate compiler's `emit` output, and the diffs are pre-existing emission drift (`10 as i64` → `10_i64`, `let mut` → `let`, whole-`_sifr.math` module emission), not new codegen behavior.

Follow-up findings:
- pre-existing issue: the canonical `pow` wrapper is still emitted unused (`demos/core_libraries/emitted.rs:1761`), so the first half of deferred row 1025 is unresolved and no decision is recorded. It is indistinguishable from the general whole-module stdlib emission (unused `sin`/`cos`/`pow_val` appear the same way), so it is not specific to `pow`; it belongs with whichever item owns emission pruning.
- pre-existing issue: `demos/python_regressions/idiomatic.rs:1811` keeps a hand-authored `BytesIO` whose backing store is a `Vec` of elements written one-by-one and whose `getvalue` returns `Result<Vec<u8>, IOError>` — residue of the `list[int]` storage removed in Item 5 and of the pre-change `Result` shape — while `demos/python_regressions/main.sifr` does not use `BytesIO` at all. Hand-authored companions are outside the new freshness checker's scope (`*/emitted.rs` only).
- suggestion: the mechanical alias renames leaked into prose that now reads worse than before — `demos/bisect/main.sifr:1,32` ("`# Reference: bisect_right`", "bisect_right bisect_right parity demo"), `crates/sifr/tests/e2e/pass/cpython_bisect.sifr:2` ("Sifr's insort_right functions return new lists", which is also factually wrong for the in-place `mut` API), and `crates/sifr/tests/e2e/pass/cpython_bisect_subset.sifr:1` / `demos/python_regressions/main.sifr:76` where a section covering both `bisect_left` and `bisect_right` is now labelled `bisect_right`. The namespace validator skips comments, so these edits were not required.
- suggestion: `crates/sifr_codegen/src/stdlib_import_signatures.rs:54` makes the source-name fallback first-import-wins for two `_sifr.*` modules exporting the same name; the aliased local entries are exact, so this only affects an ambiguous fallback, but rejecting the colliding entry outright would be less order-dependent.
- suggestion: only the function path deduplicates extra type-param bounds (`function_emitter/generic_bounds.rs:65`); `class_emitter.rs:190` still extends base bounds without a dedup filter. No duplicate survives in the current 246-companion corpus, so this is latent only.
- infrastructure issue: Item 13 still owns package-context execution for `demos/m16_raw_api` and the `audit_fixtures.py` PYTHONPATH gap, and Item 16 the taxonomy conflict that stopped the single create-PR gate; neither affects this item.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `verification/areas/stdlib_parity/reports/stdlib_parity_governance_inventory.md:55` (and `:191`), criterion "deferred Item 3 row: current stdlib architecture and parity records must name `stdlib/sifr/` instead of the obsolete `lib/sifr` source path" (prior blocking finding 2). Reason: this is an active parity record whose canonical per-module section is titled "Per-Module Readiness Inventory (`lib/sifr`)" and whose exit-gate summary says "all shipped `lib/sifr` modules", while the same commit range deliberately rewrote the identical obsolete source path in eleven sibling reports under `verification/areas/stdlib_parity/reports/` (e.g. `stdlib_parity_matrix.md`, `concurrency_runtime_legacy_surface_traceability.md:` `lib/sifr/asyncio.sifr` → `stdlib/sifr/asyncio.sifr`). The claimed "exact obsolete-path scans pass" does not hold for this file; the plain scan `grep -rn 'lib/sifr' --exclude-dir=archive` minus `stdlib/sifr` and installed-sysroot `lib/sifr/stdlib` hits surfaces it directly. Correction: rewrite both occurrences to `stdlib/sifr` and rerun the same scan over `verification/areas/**` and `plans/phases/**` rather than the two files named in the prior review.
- in-scope omission: `plans/phases/32_async_ecosystem.md:1027`, same criterion. Reason: a current phase record still names the obsolete source path `lib/sifr/asyncio.sifr` (plus `lib/sifr/concurrent.sifr`-era siblings by reference), which is exactly the string the candidate corrected to `stdlib/sifr/asyncio.sifr` in the concurrency parity report in this same range; correcting only `06_stdlib_architecture.md` and `13_type_system_completion.md` leaves the row partially open. Correction: apply the same rewrite in `32_async_ecosystem.md` (the module no longer exists under either path, so the record should read `stdlib/sifr/asyncio.sifr` for consistency with the parity report wording).

Prior blocking finding 1 is corrected: `demo-emitted-freshness` is registered in `verification/policy/guardrails.json` (valid JSON, entrypoint `scripts/check_demo_emitted_freshness.py`, empty args, 180 s, stdout) and invoked unconditionally from `ProfileRunner.run_core_guardrails` (`verification/runner/sifr_verify/profile_runner.py:333`) immediately after file-size; `core_guardrails` is a non-optional member of `LEGACY_FACADE_STEPS_BEFORE_GENERATED`, so both create-pr and merge execute it. I re-ran the checker on the candidate tree: all 246 companions fresh.

Follow-up findings:
- new mechanism-level defect: `scripts/check_demo_emitted_freshness.py:18` hardcodes `REPO_ROOT/target/debug/sifr` and hard-errors (`parser.error`, exit 2) when it is absent, instead of using the repository's established `resolve_sifr_binary` helper (`verification/areas/common/sifr_binary.py`), which honors `CARGO_TARGET_DIR`, rebuilds a missing binary, and rebuilds a stale one against `crates/`, `stdlib/`, `Cargo.lock`. Consequence now that it is a gate step: when `CARGO_TARGET_DIR` is set (a configuration this repo supports and uses, e.g. `verification/areas/common/sifr_binary.py:44`, `verification/areas/stdlib_parity/tools/run_stdlib_namespace_corpus_validation.py:54`), `ProfileRunner.__init__` builds `$CARGO_TARGET_DIR/debug/sifr` and the guardrail then aborts the whole `core_guardrails` step; when `SIFR_GCQ_BIN` is preset, no build happens at all and the checker may compare committed companions against a stale binary. It fails loudly rather than silently passing, so evidence is not fabricated, but the gate is not configuration-independent.
- new mechanism-level defect: step-budget headroom was not adjusted for the added work. The checker takes ~127 s wall on this host with an already-warm debug binary (measured, 8 jobs, 246 emissions), while `core_guardrails` carries a blocking budget of 360 000 ms in both `verification/profiles/create-pr.json:12` and the merge profile, on a step that already runs a dozen guardrails plus audit-fixture, package-management, and stdlib-parity inventory suites. Because the item's single create-PR gate was consumed before this commit and is not rerun, the combined step time was never exercised; a budget-enforcement failure of `core_guardrails` is plausible on slower or cold hosts.
- suggestion: `demo-emitted-freshness` is absent from `guardrail_steps` in `verification/profiles/create-pr.json:223` and `merge.json:337`. That list has no runtime consumer (only `verification/schemas/profile.schema.json` requires the key) and already omits several guardrails that `run_core_guardrails` executes, so this is declarative drift rather than a coverage hole, but adding the row keeps the profile metadata truthful.
- pre-existing issue: `demos/intrinsics/main.sifr:6` and `:14` tell readers that `sifr.test` is resolved from `lib/sifr/test.sifr`. These are user-facing demo comments rather than architecture or parity records, so they sit outside the deferred row's wording, but they carry the same obsolete source path.
- pre-existing issue: remaining `lib/sifr` references in `internal_docs/sifr_sysroot_and_stdlib_architecture.md:310,1115`, `internal_docs/distribution_pipeline.md:159`, `scripts/distribution/*`, and `crates/sifr_sysroot/src/tests.rs` are installed-sysroot layout paths and correctly left untouched; `plans/reviews/active/**` and `plans/issues/archive/**` occurrences are historical review snapshots.

yaseralnajjar:
Final exact candidate: `2a1eee336ac531f8ab566499a940aede461e5ac7`. The one permitted merge-gate run stopped in `coverage_matrix_checks` at the repository-wide verification-taxonomy conflict (including the active compatibility inventory and concurrently introduced adapter vocabulary). No item-specific compilation failure was reached; per phase rules, the gate will not be rerun. The second and final Opus review is already published above. Its concrete stale-path findings were corrected in this SHA, while its newly identified freshness-resolver and guardrail-budget mechanisms are recorded for Item 13; no third review was run.

PR #3279, opened 2026-08-18
Enforce explicit receiver semantics

yaseralnajjar:
## Summary
- make receiver syntax authoritative, including distinct owned-mutable receivers
- reject implicit mutable receiver behavior and fixed-trait convention violations
- migrate repository and algorithmic Sifr sources to explicit canonical receivers

## Validation
Targeted validation and the required exact-SHA gates are recorded in the phase item.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `e2e_class_unknown_parent` baseline mismatch, `hashlib.file_digest` generated-Rust scoping defect (Item 7), inherited `meta_specialization_warning` source-hash drift, and the repository-wide taxonomy checker rejection are all outside this candidate's diff and unaffected by it.
- suggestion: `interop_owned_receiver` (`crates/sifr_lowering/src/lower/classes/parameter_conventions.rs:28`) still silently rewrites a declared `self`/`mut self` receiver to `Owned` when a Rust/Python interop declaration marks the method consuming, instead of requiring `own self` at the source. This is the one remaining place where a declared convention is replaced rather than validated; a future item could make it a diagnostic in the same shape as the fixed-trait check (`fixed_trait_receiver_convention` + SIFR-PROTO-0006).
- suggestion: `static_program_slots_codegen.rs:205,220` collapses `Owned` and `OwnedMutable` into one `SlotReceiver`/`SlotReceiverIdentity::Owned`, while `structural_shape/methods.rs:411` distinguishes `own` from `own_mut`. This is ABI-correct today (both pass `self` by value), but the two identity vocabularies now differ; worth a note if structural identity ever gains receiver-mutability meaning.

Verification basis: receiver convention now maps 1:1 from syntax (`declared_receiver_convention`, with `declared_method_receiver_convention` deleted so fixed-trait names no longer override source); body analysis only validates and annotates (`validate_and_annotate_class_receivers` never assigns `method.receiver`, patching only the scope binding for recovery via `patch_receiver_mutability_for_recovery`); shared/owned-immutable mutation emits exactly one OWN-0005 at the declaration (fixed-trait methods route to PROTO-0006 instead, so no double report); protocol conformance compares explicit conventions with `OwnedMutable` conformance tests added; `own mut self` renders as `mut self` through the new `RustParam::MutableSelfValue` with native pass coverage (`explicit_owned_mutable_receiver.sifr`); repo-wide grep finds no `.sifr` operator/display dunder with a non-canonical receiver, and the corpus submodule pointer equals the stated companion merge SHA `ebd0f21`; `report_legacy_name_conflicts` is renamed to `report_same_binding_conflicts` with no stale identifiers left in code.

PR #3281, opened 2026-08-18
feat: construct adapted structural defaults

### LINKED ISSUE #3278: M7e: preserve adapted factory defaults through structural construction
M8 discovery: adapted factory defaults are finalized for constructors, but structural ShapeField currently derives requiredness only from const-evaluable HIR defaults. A factory call therefore becomes ShapeField.default=None, is emitted as required, and cannot be omitted by the existing static validator/construction path. Implement the package-neutral root cause: retain explicit required/const/factory default state in structural-shape delivery, and make generated StructuralConstruct honor missing defaulted fields by evaluating the already checked class default expression. Add a non-Pydantic adapted fixture proving required/const/factory distinction and omitted-field construction. This must not add package vocabulary, fallback construction, or runtime reflection.

yaseralnajjar:
Closes #3278\n\nAdds package-neutral structural field default states, canonical adapter factory identities, and partial named-record construction that evaluates checked defaults. Extends the static-class-adapter fixture with a native bridge covering omitted defaults and malformed record rejection.\n\nExact candidate: fde6eee6dfbc5e11d6bd704f7ced37c2c6fafe52


yaseralnajjar:
## Review: M7e Structural Construction Defaults

Reviewed `3ecb0efc..fde6eee6` (35 files). Traced index-space alignment (`adapter_field_plans` ↔ `fields` ↔ `class_field_defaults`), the new `field_default_identities` side channel through HIR → bound checking → shape identity, the rewritten `structural_construct_impl` codegen, the `ShapeFieldDefault` state split in the frontend shape, and the fixture bridge.

Key checks that hold:
- `adapter_field_plans::defaults_for_class` validates `plans.len() == fields.len()` and per-index name/type equality before producing defaults, so the `enumerate()` index in `class_shape_metadata::field_default_identities` (`crates/sifr_lowering/src/lower/classes/class_shape_metadata.rs:12`) is a true field index.
- `HirClass.fields` is `own_fields` (parent-filtered) while `field_defaults` indices are all-fields indices; this would misalign, but `structural_record_supported_in` requires `class.parent_class.is_none()` (`crates/sifr_codegen/src/structural_impl_codegen.rs:181`), so the construct/identity paths only see classes where the two spaces coincide.
- `compile_record` sets `required: default.is_none()` and `field_default_identity` now returns the sealed callable canonical, so factory fields are non-required and identity changes with callable identity (`structural_identity_codegen.rs:445,560`).
- `validated_adapter_value`'s `expect` precondition still holds: `AdapterFieldDefault::Const` is only built from `static_program_value()` in `early_adapters.rs:534`.
- Missing/unknown/duplicate edges all return typed `StructuralContractError` values from generated code; no generated panic. `lower_class_expr_strict` panics are compiler-side and match the existing constructor-default path.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_codegen/src/structural_impl_codegen.rs:113` — the parameter renames to `__sifr_source`/`__sifr_node`/`__sifr_construct_token` prevent collisions with rendered default expressions, but `description`, `child_nodes`, and `field_index` are still plain locals in the same scope where `render_expr(&value)` output is spliced. A Sifr `default_factory=description` (or a class/function of that name) would emit `description()` against the shadowing `let description = ...` binding and fail to compile. Applying the same `__sifr_` prefix to those locals closes the gap.
- suggestion: `crates/sifr_lowering/src/lower/classes/class_body_lowering.rs:1` — several explanatory comments and the `/// Second pass: lower class method bodies into HirClass.` doc comment were deleted in this diff, leaving the file at 898 lines (limit 900). The guardrail's prescribed remedy is splitting by responsibility, not trimming documentation; the next addition to this file will hit the limit again.
- suggestion: `verification/areas/core_language/fixtures/static_class_adapter/app.sifr:77` — the freshness check asserts `another.tags == []` after `defaulted.tags.append("one")` but never asserts `defaulted.tags == ["one"]`. The assertion would still pass if the append silently targeted a throwaway value; adding the positive assertion makes the non-sharing claim two-sided.
- pre-existing issue: generated construction evaluates field defaults as sequential `let` bindings in declaration order, so a factory default on an earlier field runs before a later required-field omission is detected and returns `ArityMismatch`. Side effects from that factory are not unwound. Harmless for the pure factories the adapter contract validates today, but it is an observable ordering commitment.

yaseralnajjar:
Gate evidence for exact candidate fde6eee6dfbc5e11d6bd704f7ced37c2c6fafe52:

- create-PR gate: attempted exactly once; coverage readiness passed, then the shared verification-taxonomy preflight failed before compiler tests. The scanner flags the explicitly out-of-scope pre_v1 inventory/checker and broad ordinary contract/milestone strings across existing compiler, fixture, and documentation files.
- merge gate: attempted exactly once; same shared verification-taxonomy preflight failure before compiler tests.
- neither gate was repeated.

Candidate-specific validation remains green: focused frontend/codegen/lowering/driver tests, workspace clippy -D warnings, formatting, HIR maintainability, file-size guardrail, and the native static-class-adapter fixture build/run. The separate compiler unit command passed 109 CLI tests and 37/38 harness tests; its sole failure is the pre_v1-owned protocol_bound_unknown_forwarded_typevar.sifr expectation.

PR #3299, opened 2026-08-19
Preserve inherited handler contracts in structural shapes

yaseralnajjar:
M9 compiler prerequisite for pydantic-sifr validators.\n\nPreserves checked inherited handler signatures, exposes successful Result output separately from the declared result, and uses finalized adapter fields in structural shapes.\n\nTargeted evidence: 13 structural-shape tests; sifr_frontend clippy; formatting; HIR maintainability; 900-line guardrail; release compiler successfully built and used by the M9 package demo.

### DIFF
--- a/crates/sifr_frontend/src/structural_shape.rs
+++ b/crates/sifr_frontend/src/structural_shape.rs
@@ -362,7 +362,14 @@ fn describe_class(
                     .map(|selection| selection.field_plans.as_slice()),
             )
         };
-    let described_fields = fields
+    let planned_fields = field_plans.map(|plans| {
+        plans
+            .iter()
+            .map(|field| (field.name.clone(), field.declared_type.clone()))
+            .collect::<Vec<_>>()
+    });
+    let effective_fields = planned_fields.as_deref().unwrap_or(fields);
+    let described_fields = effective_fields
         .iter()
         .enumerate()
         .map(|(index, (name, ty))| ShapeField {
@@ -641,13 +648,15 @@ fn canonical_method(method: &ShapeMethod) -> String {
         .as_ref()
         .map_or_else(|| "none".to_string(), canonical_value);
     format!(
-        "{}:{}:{}:{}:{}:target[{target}]:descriptor[{descriptor}]:params[{params}]:result[{}]:meta[{}]",
+        "{}:{}:{}:{}:{}:target[{target}]:descriptor[{descriptor}]:params[{params}]:result[{}]:output[{}]:fallible[{}]:meta[{}]",
         method.name.len(),
         method.name,
         method.kind,
         method.receiver.as_deref().unwrap_or("none"),
         method.is_async,
         canonical_node(&method.result),
+        canonical_node(&method.output),
+        method.fallible,
         canonical_metadata(&method.metadata)
     )
 }
--- a/crates/sifr_frontend/src/structural_shape/methods.rs
+++ b/crates/sifr_frontend/src/structural_shape/methods.rs
@@ -30,6 +30,9 @@ pub struct ShapeMethod {
     pub is_async: bool,
     pub params: Vec<ShapeParameter>,
     pub result: Box<ShapeNode>,
+    /// Successful checked output, with a `Result` carrier removed.
+    pub output: Box<ShapeNode>,
+    pub fallible: bool,
     pub metadata: Vec<ShapeMetadata>,
 }
 
@@ -110,6 +113,8 @@ pub(super) fn const_value(method: &ShapeMethod) -> ConstValue {
             ),
         ),
         ("result".to_string(), node_const_value(&method.result)),
+        ("output".to_string(), node_const_value(&method.output)),
+        ("fallible".to_string(), ConstValue::Bool(method.fallible)),
         (
             "metadata".to_string(),
             metadata_const_value(&method.metadata),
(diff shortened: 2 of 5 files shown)


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_frontend/src/structural_shape/methods.rs:309` (imported-handler branch, `described_method(..., &HashMap::new(), ...)`), criterion "Substitute direct generic parent arguments in inherited handler input and output types". The new imported branch resolves the handler identity and preserves its checked signature, but passes empty bindings, so a handler inherited from a *generic imported* direct parent keeps its type parameters. For `class Child(pkg.Parent[int])` with an inherited handler `def normalize(cls, own value: T) -> T`, `params[0].declared_type`, `result`, and `output` describe `ShapeNode::TypeParameter("T")` (canonical identity `param:T`) instead of `int`, so a package checking the handler contract structurally sees an unbound parameter and rejects (or mis-accepts) a valid handler — the same case the local branch handles correctly and asserts in `tests.rs:436-451`. This is a gap in newly added code, not a regression from the base SHA (that path previously produced the `checked_handler` placeholder). Correction: build the bindings for the imported owner the same way `inherited_handler_bindings` does for local owners — take the child's resolved `parent_type` `type_args`, substitute them through the child `bindings`, and zip them against `external_defs.class_type_params.get(source_module).and_then(|classes| classes.get(source_name))` (the same source `described_exported_methods` already uses at `methods.rs:394-406`), guarding that the parent identity matches `handler.callable.owner`. Add a test with an imported generic parent handler alongside the existing local one.

Follow-up findings:
- pre-existing issue: transitively inherited handlers (grandparent → parent → child, propagated by `early_adapters/handler_plans.rs::inherited_handler_plans` cloning the parent's plans) reach `inherited_handler_bindings` (`methods.rs:341`) with a non-matching `parent_type` and get an empty binding map, so grandparent type parameters stay unsubstituted. The stated criterion covers only *direct* parents, so this is outside this item, but the propagation makes it reachable.
- suggestion: the imported branch passes `Vec::new()` metadata and `&[]` declaration metadata (`methods.rs:310-311`), while `described_exported_methods` passes the external module's `declaration_metadata`. Inherited handlers from an imported parent therefore expose no method or parameter metadata. Not required by the listed criteria, but it makes local and imported inherited handlers structurally asymmetric.
- suggestion: `imported_handler_method` (`methods.rs:374`) has no test coverage; all 13 passing `structural_shape::tests` exercise only the local-owner path. A fixture using `external_defs.structural_methods_for` would lock in both identity resolution and the substitution fix above.
- suggestion: `inherited_handler_bindings` short-circuits on `child.name == owner.name` (`methods.rs:345`) and falls back to `name != &owner.name` comparison at `methods.rs:361`; both compare bare class names while the surrounding code keys on `identity`. Comparing identities consistently would keep the resolution package-neutral by construction.
- suggestion: `structural_shape.rs:365-371` now indexes `shape_field_default` against the plan order, which matches `defaults.rs:9` (`field_plans.get(index)`) — this actually removes a latent index mismatch between HIR field order and plan order. Worth a short note in `internal_docs/const_specialization.md` next to the new `output`/`fallible` paragraph, since described fields are now plan-ordered whenever an adapter selection exists.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_frontend/src/structural_shape.rs:365-377` (`planned_fields` / `effective_fields` / field `describe_node`), criterion "Use finalized adapter field plans so inherited normalized fields are present" (and "Keep canonical identity sensitive to output/fallibility" by contamination of the identity string). Reason: the described fields now come from `AdapterFieldPlan.declared_type`, which is the *declaration-level* type, and no type-argument substitution is applied — unlike methods, which go through `substitute_type_vars` with the `type_params`/`type_args` bindings in `structural_shape/methods.rs`. Adapter selections on generic classes are allowed (`crates/sifr_frontend/src/early_adapters.rs:808-825` only skips the specialization request for an unbound generic owner; field plans are still applied), and for `class Parent[T](Contract): value: T` the finalized plan holds `Type::TypeVar("T")` (`crates/sifr_lowering/src/lower/adapter_field_plans.rs:32-45` enforces plan type == HIR field type). Before this commit, `fields` came from the instantiated `Type::Class`, whose field types are substituted at annotation resolution (`crates/sifr_lowering/src/lower/typing_and_functions/annotation_resolution.rs:477-486`). Failure scenario: describing a concrete instantiation of a generic adapted class — e.g. a field annotated `Parent[int]` (the documented `RootModel[T]` pattern, `internal_docs/native_pydantic_sifr_architecture.md:1172`) — previously yielded `ShapeField { name: "value", declared_type: Primitive("int") }` and now yields `TypeParameter("T")`, so the canonical identity of a fully concrete shape contains `param:T` while that same shape's method contracts correctly report `int`. The same loss applies on the imported side, since the external branch reads `external_defs.class_adapter_selections[...].field_plans` identically. Correction: build the class bindings once (local `class.type_params` / `external_defs.class_type_params` zipped with `type_args`, as `described_methods` and `described_exported_methods` already do) and apply `substitute_type_vars` to each `field.declared_type` before `describe_node`; add a regression test describing a concrete instantiation of a generic adapted class and asserting `int` fields plus no `param:T` in the canonical identity.

Follow-up findings:
- suggestion: `structural_shape/methods.rs:336-337` changes inherited-handler `origin` to `None` (const value `None`). This is the correct fail-closed choice given `internal_docs/const_specialization.md:40-41,97` (a package may not forward an origin from another declaration), but the doc update in this commit only covers `result`/`output`/`fallible` and does not state that an inherited handler exposes no origin, leaving packages without documented guidance to anchor inherited-handler issues at the child declaration origin.
- suggestion: `structural_shape/methods.rs:344-395` — `imported_handler_bindings` and `inherited_handler_bindings` return empty bindings on any identity mismatch, and only the *direct* parent is matched. A handler inherited from a generic grandparent therefore silently exposes raw `TypeVar`s rather than failing closed. Direct-parent-only is the stated scope, but the silent degradation is worth an explicit invalid-plan diagnostic or a recorded limitation.
- pre-existing issue: `structural_shape.rs:410-436` — the imported-class branch (`described_exported_methods`) never consults `external_defs.class_adapter_selections[...].handler_plans`, so a fully imported adapted class exposes its handlers as plain exported methods with no `target`/`descriptor`/`declaration_order`. Unchanged by this commit; the new import test exercises a locally declared child, not an imported child.
- infrastructure issue: the recorded evidence is `cargo test -p sifr_frontend structural_shape` (18 tests) only. `canonical_method` gained `output[..]:fallible[..]` segments and `const_value` gained `output`/`fallible` keys, both of which feed const-specialization package inputs; no full-crate, driver, or PR-gate run is recorded on this candidate.

yaseralnajjar:
Create-PR gate evidence for exact candidate `2f4eceb110a84e98f6ed24dfb303df4ac51dae99`: one invocation of `scripts/run_all_tests.sh --profile create-pr`. The SHA was unchanged before/after. Cache setup, HIR maintainability, and the 900-line guardrail passed. The gate stopped at `guardrail_demo_emitted_freshness` after reporting the shared stale demo-emission set owned by the parallel pre_v1 session; no M9-specific failure was reported. Per the phase rule, this gate will not be rerun and the external artifacts will not be changed.

yaseralnajjar:
Merge-gate evidence for exact candidate `2f4eceb110a84e98f6ed24dfb303df4ac51dae99`: one invocation of `scripts/run_all_tests.sh`. The SHA was unchanged before/after. Cache setup, HIR maintainability, and the 900-line guardrail passed. The gate stopped at the same `guardrail_demo_emitted_freshness` shared stale demo-emission set owned by the parallel pre_v1 session; no M9-specific failure was reported. Per the phase rule, no second merge gate will run.

PR #3301, opened 2026-08-19
Remove string-based Rust type rendering

yaseralnajjar:
## Summary
- make structured `RustType` the canonical representation for every supported Sifr type
- remove `Type::rust_type` and `Type::rust_type_for_struct_field`
- preserve generic call type arguments structurally in HIR and add a mutation-tested residue guard

## Validation
- `cargo test -p sifr_codegen` (1,045 passed)
- workspace Clippy and rustfmt
- structured-type guard and self-test, raw-code gate, file-size and HIR guardrails
- focused channel E2E batch (12/12 passed)
- full E2E: 661 passed; inherited failures isolated to nullable-union scope, hashlib scoping, and the fixed channel constructor regression
- generated-code quality smoke suites passed; representative corpus reached the inherited structured-concurrency project-splitting defect
- sole create-PR gate stopped at the inherited 98 stale demo companions and was not repeated

Exact candidate: `0f1b0e8d755dd7445585f736c7d929260e78e387`


yaseralnajjar:
Item 15 initial Opus review — exact candidate SHA `0f1b0e8d755dd7445585f736c7d929260e78e387`, base SHA `e936e94c72287122713994ffb467a8c54b82e0ef`. The following comment is the complete atomic review response.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_codegen/src/generic_bounds_helpers.rs:135` (`rust_ir_type_with_generics`), criterion "Rust imports and generic nesting come from structured nodes" / "Generated Rust snapshots preserve required output semantics". The deleted string renderer `rust_type_with_generics` applied emitter-owned generic knowledge *recursively* — every composite arm (`List`, `Dict`, `Set`, `Tuple`, `Result`, `Task`/`BlockingTask`/`JoinSet`/`TaskResult`, optional-union payload, `Failure`, `TimeoutResult`, `Select2`, `AsyncGenerator`, `Callable`/`AsyncCallable`, `Alias`) recursed into itself, and the `Class` arm mapped its own `type_args` through itself. The replacement only inspects the outermost node: if it is not a `Class`, or is a `Class` with non-empty `type_args`, the whole subtree is handed to `crate::sifr_type_to_rust_type`, which has no access to `generic_classes` / `generic_class_templates` / `generic_class_params`. So a generic class that needs inferred or parameter-name arguments and appears nested — `list[Counter]` inside a generic impl, or `Stack[Counter]` as a type argument — now renders `Vec<Counter>` / `Stack<Counter>` instead of `Vec<Counter<T>>` / `Stack<Counter<T>>`, i.e. emitted Rust that references a generic struct without its arguments. The same delegation drops the old `Type::Never => "std::convert::Infallible"` mapping: bare `Never` now renders `!` (only `task_error_type_to_rust_type` still yields `Infallible`), so a `Result[T, Never]` reaching any of these 43 call sites renders `Result<T, !>`. Correction: keep the conversion emitter-aware for the whole subtree — walk composite variants in `rust_ir_type_with_generics` (or pass an emitter-backed hook into `sifr_type_to_rust_type`) so `Class` nodes at any depth get inferred/parameter arguments, and preserve the `Never -> std::convert::Infallible` rule at the non-task positions that previously used it. Add a nested case to the `generic_bounds_helpers` tests (`Vec<Counter<T>>`), which currently only cover top-level classes.
- in-scope omission: `crates/sifr_codegen/src/preamble/types_and_errors.rs:7` (`try_sifr_type_to_rust_type`) with `crates/sifr_codegen/src/preamble/type_validation.rs:3`, criterion "Unsupported types produce one structured compiler error" and scope bullet "Replace fallback panics with structured codegen errors at approved boundaries". The fallible converter has no production caller — the only references in the tree are `crates/sifr_codegen/src/type_conversion_tests.rs:164` and `:191`. Production code calls the infallible `sifr_type_to_rust_type`, whose `callable_param_types` uses `conventions.get(index)` and silently drops missing conventions (`types_and_errors.rs:294`), so the one shape the validator classifies as unsupported (params/conventions arity mismatch) produces no compiler error in the compiler itself; the criterion is satisfied only inside a unit test. Correction: invoke `try_sifr_type_to_rust_type` (or `validate_codegen_type`) at the codegen boundaries that already return `CodegenError` — function signature, struct field, and let-type emission — and propagate the single structured error, keeping the infallible converter for internal recursion.

Follow-up findings:
- suggestion: `crates/sifr_type_system/src/types/union_identity.rs:16,18` — the collapse and non-union arms of `union_enum_name()` changed from returning the collapsed member's real Rust name to a synthetic `__SifrUnion_<key>` identifier. `register_union_type_with_usage` canonicalizes before registering, so no enum with that name is ever emitted; today every caller guards on `Type::Union` first, so this is latent rather than live, but it converts a previously self-consistent corner into a dangling type name. The accompanying test `raw_identity_duplicate_union_renders_as_its_canonical_member` now asserts the opposite of what its name states.
- suggestion: `crates/sifr_lowering/src/lower/python_interop/callbacks.rs:597` — `error_channel_codegen_payload_collision` previously compared emitted Rust names for all members; it now compares `class_rust_name` for classes and `display_name` for everything else, so a collision between a source class and a builtin that share an emitted name (e.g. a user class `Decimal` alongside `Type::Decimal`) is no longer reported.
- pre-existing issue: `test_unknown_parent_class_has_class_code` (`crates/sifr_lowering/src/lower/expressions_tests/control_flow_and_strings.rs:573`) asserts a `CLASS_INVALID_BASE` message/code/range for an undefined parent class. Nothing in this candidate touches class-base diagnostics, ranges, or diagnostic codes; the lowering changes are the `GenericCall` node plumbing, the payload-collision helper, and snapshot projections. Inherited.
- pre-existing issue: hashlib sequential-try scoping defect and nullable-union safe-operation coercion in the E2E sweep; `structured_concurrency_demo` emitting `ScopeFailure` outside the project nominal module without `From<ScopeFailure> for Error`. All base mechanisms, unrelated to type conversion.
- infrastructure issue: the create-PR gate stop on the 98 stale generated demo companions introduced by the static-class-adapter base change.

yaseralnajjar:
Item 15 one permitted remediation Opus review — exact candidate SHA `fdfbc82c49aca299985a12e41e131eca6dec8cfb`, base SHA `e936e94c72287122713994ffb467a8c54b82e0ef`. The following comment is the complete atomic review response. It confirms both original blockers resolved, then identifies new mechanisms; under the phase rule, those new mechanisms are deferred to final closure with no third review.

yaseralnajjar:
Verdict: NOT SATISFIED

Both prior blockers are resolved: `rust_ir_type_with_generics` is now genuinely recursive (verified `Vec<Counter<T>>`, `Result<i64, ::std::convert::Infallible>`, and structured containers/tuples/futures/callables/class args at any depth), and malformed callable arity now produces one production `compile_error!` through `reject_invalid_codegen_module_types` at `crates/sifr_codegen/src/lib_emitter_state.rs:363`. The guard, inventory count (0), and `.rust_type()` removal all check out. New regressions surfaced in the mechanical replacement itself.

Blocking findings:
- regression: `crates/sifr_codegen/src/stmt_support_emitter/recursive_constructor_args.rs:37`, criterion "generated Rust snapshots preserve required output semantics" (behavior-preserving replacement of `.rust_type()`). The removed gate `param_ty.rust_type().starts_with("Option<Box<")` was live, not dead: it selected constructor params whose Rust field type is `Option<Box<...>>` (optional protocol, `Iterator[T]`, `Any`, `Unknown`, `Intersection`), not only recursive fields. With the gate gone, the `Some(Box::new(..))` adaptation is skipped. Verified at the candidate SHA: a class with field `entity: Greetable | None` and `def __init__(self, own entity: Greetable | None)` emits `Holder::new(Some(p))` against `fn new(entity: Option<Box<dyn Greetable>>)`, and `sifr build` fails with rustc `E0308 ... help: store this in the heap by calling Box::new`. At base the string gate produced `Some(Box::new(p))`. Correction: restore the condition structurally, e.g. keep the early return only when neither `is_recursive_ctor_param` nor `matches!(crate::sifr_type_to_rust_type(context.param_ty), RustType::Option(inner) if matches!(*inner, RustType::Boxed(_)))` holds.
- regression: `crates/sifr_codegen/src/lower_stmt/with_yield_and_match.rs:447` (and the same construct at `:508`), `crates/sifr_codegen/src/class_method_emitter.rs:301`, `crates/sifr_codegen/src/lower_expr/leaves_and_plain_calls.rs:523`, criterion "generated Rust snapshots preserve required output semantics". `Type::rust_type()` returned only `class_rust_name(..)` for `Type::Class`, dropping `type_args`; `render_type(&sifr_type_to_rust_type(ty))` renders `Base<i64>`, which is illegal in these path/pattern positions. Verified at the candidate SHA: a `match` on `Box[int]` emits `Box<i64> { value: 0, .. }` and `sifr build` fails with `error: generic args in patterns require the turbofish syntax`; `class IntBox(Base[int])` with `super().__init__(value)` emits `let __sifr_parent = Base<i64>::new(value);` and `super().label()` emits `Base<i64>::label(self)`, failing to parse. Correction: use the base-name-only projection already applied in `canonical_constructor_class_name` (`RustType::Generic { base, .. } => base`) at these name positions, or emit turbofish (`Base::<i64>::new`) where a path is legal; struct patterns must not carry bare `<...>`. No existing fixture covers generic-parent inheritance or generic-class match patterns, which is why the focused suites stayed green.

Follow-up findings:
- suggestion: `try_sifr_type_to_rust_type` (`crates/sifr_codegen/src/preamble/types_and_errors.rs:7`) still has no production caller; production validation goes through `validate_codegen_type` directly. Either route a boundary through it or drop the wrapper.
- suggestion: `scripts/check_structured_rust_types.py` anchors its indirect-fallback checks to the text window between `pub fn sifr_type_to_rust_type` and `pub fn build_error_type_items` in one file; renaming or moving `build_error_type_items` silently disables those checks (the direct-pattern checks stay repo-wide).
- suggestion: `reject_invalid_codegen_module_types` emits a fixed-name `__sifr_codegen_type_error` item per module, so two invalid modules would produce duplicate item names alongside the intended `compile_error!`.
- pre-existing issue: comparing an `Option<Box<dyn Trait>>` field with `None` (`assert x.entity is None`) emits `==` on a non-`PartialEq` type (rustc `E0369`); observed while probing, independent of this diff.

yaseralnajjar:
Final exact-candidate validation for `fdfbc82c49aca299985a12e41e131eca6dec8cfb`: the sole merge gate (`scripts/run_all_tests.sh`) passed Cargo cache setup, HIR maintainability, and the 900-line file-size guard, then stopped at the inherited demo-emitted freshness check with the same 98 stale `demos/**/emitted.rs` companions introduced by the static-class-adapter base change. Per the phase rule, the gate was not rerun.

PR #3303, opened 2026-08-19
Support receiver and value method slots

yaseralnajjar:
M10 compiler prerequisite. Extends package-neutral checked method slots so an instance handler can receive one owned structural value before an optional borrowed context. The sealed input is a structural receiver/value tuple; generated dispatch reconstructs both values. Includes frontend signature coverage and runtime dispatch evidence.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Reviewed at the exact SHAs (`git diff 1babdde1..576fd680`, six paths). What I verified:

- `crates/sifr_frontend/src/slot_table.rs:632-659` — receiver-only (`[]` → `input_type = owner_type`, no context) and receiver-plus-borrowed-context (`[p] if p.convention.is_borrowed()`) are preserved byte-for-byte in effect; the new `[value]` and `[value, context]` arms seal `Type::Tuple([owner, value])`. Non-receiver slots are unaffected (their max was already 2). Context still must be a borrow (`slot_table.rs:669-680`), so ownership/context typing rules hold, and the tuple flows through `structural_slot_type_supported`, so a nonstructural value input still fails.
- `crates/sifr_codegen/src/static_program_slots_codegen.rs:101-182` — the arm destructures `(receiver, value)` from one sealed `structural_construct::<(Owner, Val), _>` call, and mutability of the receiver binding is still driven only by `ReceiverConvention::MutableBorrow`, matching the old `mutable_input` rule for the receiver-only and static branches. Argument order is `receiver.method(value[, context])`, matching declaration order; `argument_for_convention` keeps the value's declared convention. Fallible/infallible dispatch and `SlotError::Contract` mapping are untouched, so panic containment and error surfaces are unchanged.
- Slot identity: `slot_signature_expression`/`slot_identity_expression` still key on `input_type`'s shape identity, so the composite form gets a distinct sealed identity; `structural_construct` rejects a shape mismatch (`crates/sifr_runtime/src/interop/structural.rs:281`) and the tuple impl checks kind, arity, and index edges (`implementations.rs:165-196`) — the input stays sealed.
- Evidence: `slot_table_tests.rs:171` asserts the composite input type and `StaticMethodSlotContext::None`; the package-neutral fixture adds `serialize(self, own value: str, mut context: AppContext)` with a `(Record, String)` arena at slot index 2, asserts `"input-serialized"` and `context.calls == 3`, and correctly rebases the shared-context assertion to `"value-shared-3"`. Receiver-only, static, shared-context, and no-context dispatch remain exercised. Doc contract updated at `internal_docs/const_specialization.md:166-168`.

Follow-up findings:
- suggestion: `slot_table.rs:648` — the `[value, context]` arm does not require the first parameter to be owned, so `def m(self, value: str, mut ctx: C)` (params default to borrow) is silently reinterpreted as the value input, while the same borrowed parameter alone is a context. The generated `receiver.m(&value, context)` compiles and is ownership-safe, but the accepted surface is wider than the "one owned structural value" contract; add a non-borrowed check plus a `SIFR-RUST-SLOT-####` diagnostic and a rejection test.
- suggestion: `static_program_slots_codegen.rs:179` — `has_receiver_value_input` re-infers the composite form by pattern-matching a 2-element tuple on `input_type` instead of reading a recorded frontend flag. It is correct today only because `owner_type` is always `Type::Class`; an explicit field on `StaticMethodSlot` would remove the implicit coupling.
- suggestion: `static_program_slots_codegen.rs:109-117` — `value_binding` can only ever be `""` in the composite branch (a mut-borrow first parameter is routed to the context arm), so that branch is dead in practice; it is harmless but reads as supported behavior.
- suggestion: `slot_table.rs:629` — the arity diagnostic text ("exactly one value input and at most one context parameter") is still accurate but now covers a wider receiver form and has no test coverage anywhere in the repo.
- suggestion: no codegen-level unit test pins the emitted `let (receiver, value) = …` arm; the only coverage of the emitted shape is the runtime fixture.
- infrastructure issue: the broad all-target Clippy stop on inherited test-only lints outside this diff is unrelated to these paths and correctly left unabsorbed.

yaseralnajjar:
Create-PR gate evidence for exact candidate 576fd68023837d241701f170432ef6e5569d79b6: the single scripts/run_all_tests.sh --profile create-pr invocation passed Cargo cache setup, HIR maintainability, and the 900-line file-size guard. It then stopped at guardrail_demo_emitted_freshness on the shared stale generated demo companions. Wall time was 180.18 seconds, within budget. Per the phase rule, the gate was not rerun.

yaseralnajjar:
Merge gate evidence for exact candidate 576fd68023837d241701f170432ef6e5569d79b6: the single scripts/run_all_tests.sh invocation passed Cargo cache setup, HIR maintainability, and the 900-line file-size guard. It then stopped at the same shared guardrail_demo_emitted_freshness boundary. Wall time was 103.63 seconds, within budget. Per the phase rule, the gate was not rerun.

PR #3327, opened 2026-08-19
Preserve structural bounds on ordinary generics

yaseralnajjar:
## Summary
- classify ordinary `Structural` type parameters separately from `StringStructural`
- emit `StructuralConstruct + StructuralProject` for plain and attached generic Sifr functions
- retain the additional clone/static contract only for `StringStructural`

## Phase item
M11 compiler prerequisite 6 for `ad-hoc-static-class-adapters-and-pydantic-ergonomics`.

## Validation
- `cargo test -p sifr_codegen` (1053 passed)
- `cargo clippy -p sifr_codegen -- -D warnings`
- `cargo fmt --check`
- HIR maintainability guardrails, touched-file 900-line guardrail, and `git diff --check`
- release compiler regenerated the package M11 attached wrappers with the required construction/projection bounds; the build advanced to the package's separate registry-dependency policy failure

### DIFF
--- a/crates/sifr_codegen/src/function_emitter/generic_bounds.rs
+++ b/crates/sifr_codegen/src/function_emitter/generic_bounds.rs
@@ -20,10 +20,13 @@ impl RustEmitter {
             .iter()
             .map(|type_param| {
                 let string_structural = self
-                    .structural_type_params
+                    .string_structural_type_params
                     .get(&func.name)
                     .is_some_and(|params| params.contains(type_param));
-                let structural = string_structural
+                let structural = self
+                    .structural_type_params
+                    .get(&func.name)
+                    .is_some_and(|params| params.contains(type_param))
                     || func.rust_interop.iter().any(|declaration| {
                         declaration.kind == sifr_ir::RustInteropDecoratorKind::Structural
                     });
--- a/crates/sifr_codegen/src/lib_codegen_tests/structural_impl_demand_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/structural_impl_demand_codegen_tests.rs
@@ -515,7 +515,36 @@ fn plain_string_structural_generic_emits_projection_bounds() {
 }
 
 #[test]
-fn attached_structural_generic_preserves_attached_api_bounds() {
+fn plain_structural_generic_emits_projection_bounds() {
+    let mut retain = ordinary_function("retain", Type::TypeVar("T".to_string()));
+    retain.return_type = Type::TypeVar("T".to_string());
+    retain.type_params = vec!["T".to_string()];
+    retain.body = vec![sifr_ir::HirStmt::Return {
+        value: Some(HirExpr::Name {
+            name: "value".to_string(),
+            binding_id: None,
+            ty: Type::TypeVar("T".to_string()),
+        }),
+    }];
+    let mut module = module(vec![retain], Vec::new());
+    module.type_param_bounds.insert(
+        "retain".to_string(),
+        std::collections::HashMap::from([("T".to_string(), vec!["Structural".to_string()])]),
+    );
+
+    let rust_code = generate_rust(&module);
+
+    assert!(
+        rust_code.contains(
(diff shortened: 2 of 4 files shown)


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Review notes (how I checked, not findings): `self.structural_type_params` has exactly one consumer — `crates/sifr_codegen/src/function_emitter/generic_bounds.rs:26` — so widening it from "StringStructural only" to "Structural or StringStructural" cannot leak into other emitter paths (the same-named fields in `rust_interop_bridge_contract.rs` / `rust_interop_probe.rs` are a different struct). Branch order in `generic_bounds.rs:53-56` keeps `string_structural` ahead of `structural`, so StringStructural still yields construct/projection + `Clone + 'static`, and the `structural && static_program` branch at line 48 is unreachable for a single-bound param, so it is unchanged. The new prescan block (`module_prescan.rs:11-24`) reuses the existing single-bound `matches!` convention used by the StaticProgram/MethodSlots/Context maps, keeping classification compiler-owned. The inverted test at `structural_impl_demand_codegen_tests.rs:539` is the intended behavior change, and the added `plain_structural_generic_emits_projection_bounds` test covers the ordinary-function criterion. No compatibility shim or fallback branch was added.

Follow-up findings:
- suggestion: `module_prescan.rs:11-36` now runs two near-identical passes over `module.type_param_bounds` (superset + string subset); a single pass building both maps would avoid the duplicated traversal, but this matches the surrounding style and is not a defect.
- suggestion: the new negative assertions `!rust_code.contains("T: Clone + 'static")` (`structural_impl_demand_codegen_tests.rs:541,568`) are coarse — asserting the full rendered bound string, as `plain_string_structural_generic_emits_projection_bounds` does, would pin the exact bound set.
- pre-existing issue: automatic `Clone` inference from a body is only applied to class methods (`generic_bounds_helpers.rs:717-728`), not free functions. A `Structural`-bound generic free function whose emitted body clones the value now has no `Clone` bound to satisfy it. This follows directly from the stated criterion (Structural emits construct/projection only, Clone is StringStructural-specific), so it is not an omission against scope, but it is the failure mode to watch when more M11 wrappers land.
- infrastructure issue: the package regeneration evidence stopped at an arrayvec registry-dependency policy error, i.e. rustc never type-checked the three regenerated wrapper bodies. The bound text is confirmed; end-to-end compilation of those wrappers is not yet demonstrated by that run.

yaseralnajjar:
Create-PR gate evidence for exact candidate `d4615cf9bd072349e9ee75b1059bc2e5f8195c0c`:

- Command: `scripts/run_all_tests.sh --profile create-pr`
- Result: stopped at the externally owned Rust-interop fixture matrix.
- Passed before stop: Cargo cache setup; HIR maintainability; repository file-size guardrail; generated-demo freshness; dependency-direction, ownership, sysroot resource, intrinsic/reachability/manifest/bootstrap, driver maintainability, verification hardening, and runner-foundation guards.
- External failures: missing `negative/package_generated_type_import_rejected.sifr`; `examples/method_slot_runtime/src/schema_program.sifr` contains an empty placeholder body.
- Policy: this single create-PR gate will not be rerun, and the external fixture work is not absorbed into this item.

yaseralnajjar:
Merge gate evidence for exact candidate `d4615cf9bd072349e9ee75b1059bc2e5f8195c0c`:

- Command: `scripts/run_all_tests.sh`
- Result: stopped at the same externally owned Rust-interop fixture matrix as the create-PR gate.
- Passed before stop: Cargo cache setup; HIR maintainability; repository file-size guardrail; generated-demo freshness; dependency-direction, ownership, sysroot resource, intrinsic/reachability/manifest/bootstrap, driver maintainability, verification hardening, and runner-foundation guards.
- External failures: missing `negative/package_generated_type_import_rejected.sifr`; `examples/method_slot_runtime/src/schema_program.sifr` contains an empty placeholder body.
- Policy: this was the single merge gate for this exact SHA. It will not be rerun, and the external fixture work is not absorbed into this item.

PR #3329, opened 2026-08-19
Pin retained generated dependencies

yaseralnajjar:
## Summary
- pin every compiler-retained registry dependency to the version authorized by the sysroot lock
- add a manifest invariant test for exact retained versions
- update generated dependency expectations and parity snapshots

## Phase item
M11 compiler prerequisite 7 for `ad-hoc-static-class-adapters-and-pydantic-ergonomics`.

## Validation
- `cargo test -p sifr_stdlib_manifest`
- affected-crate Clippy with warnings denied; formatting, maintainability, file-size, diff, and demo-freshness guards
- locked pydantic-sifr M11 native build completed through rustc; generated resolution retained `rust_decimal 1.41.0` and authoritative `arrayvec 0.7.6`


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Evidence reviewed (read-only): the diff is confined to pinning the six retained registry specs in `crates/sifr_stdlib_manifest/src/features/dependency_plan.rs:266-282`, the new invariant test at `:339-360`, and the matching expectation/snapshot updates. Independent checks I ran:
- `retained_dependency_specs` is the only production source of registry dep lines (`git grep` over `crates/**/*.rs` for the pinned versions returns only that function), so no unpinned sibling path remains.
- All six pins match the authoritative lock exactly (`Cargo.lock`: bigdecimal 0.4.10, num-bigint 0.4.6, num-traits 0.2.19, rayon 1.12.0, rust_decimal 1.41.0, tokio 1.52.3; arrayvec 0.7.6 present as the transitive resolution the criterion names).
- Every remaining occurrence of the old unpinned strings is in `plans/reviews/archive/**` historical review records, not in expectations or snapshots; the codegen assertion, both parity snapshot files, and the traceability report are updated consistently.
- Other spec-consuming tests are prefix- or shape-based (`crates/sifr_stdlib_manifest/src/features_tests.rs:89`, `crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs`) or use synthetic hand-built input (`crates/sifr_driver/src/build/cargo_manifest.rs:472`), so none is silently stale.
- No vendor bypass, fallback, or compatibility branch is added; `CargoVendorMode` and the fingerprint path are untouched apart from the pinned strings flowing through them. The test is package-neutral — it asserts the spec shape, with no pydantic-sifr coupling.

Follow-up findings:
- suggestion: `crates/sifr_stdlib_manifest/src/features/dependency_plan.rs:341-348` enumerates the six features by hand while `retained_dependency_specs` ends in a `_ => &[]` catch-all, so a future feature that adds an unpinned retained spec would neither fail to compile nor fail this test. `StdlibFeature` has no `ALL`/iterator today, so making this exhaustive needs a separate variant inventory — out of scope here, but worth adding when one exists.
- suggestion: the assertion's `specification.contains("version = \"=")` arm is a substring check rather than a parse of the dependency table; it accepts the current inline-table shape but would also accept a malformed table containing that substring. Low risk given the specs are `const` and adjacent to the test.

yaseralnajjar:
Create-PR gate evidence for exact candidate `70fcee1338c8bba04ba8cc87f884abd458fe078f`:

- Command: `scripts/run_all_tests.sh --profile create-pr`
- Result: stopped at the externally owned Rust-interop fixture matrix.
- Passed before stop: Cargo cache setup; HIR maintainability; repository file-size; generated-demo freshness; dependency-direction, ownership, sysroot resource, intrinsic/reachability/manifest/bootstrap, driver maintainability, verification hardening, and runner-foundation guards.
- External failures: missing `negative/package_generated_type_import_rejected.sifr`; `examples/method_slot_runtime/src/schema_program.sifr` contains an empty placeholder body.
- Policy: this single create-PR gate will not be rerun, and the external fixture work is not absorbed into this item.

yaseralnajjar:
Merge gate evidence for exact candidate `70fcee1338c8bba04ba8cc87f884abd458fe078f`:

- Command: `scripts/run_all_tests.sh`
- Result: stopped at the same externally owned Rust-interop fixture matrix as the create-PR gate.
- Passed before stop: Cargo cache setup; HIR maintainability; repository file-size; generated-demo freshness; dependency-direction, ownership, sysroot resource, intrinsic/reachability/manifest/bootstrap, driver maintainability, verification hardening, and runner-foundation guards.
- External failures: missing `negative/package_generated_type_import_rejected.sifr`; `examples/method_slot_runtime/src/schema_program.sifr` contains an empty placeholder body.
- Policy: this was the single merge gate for this exact SHA. It will not be rerun, and the external fixture work is not absorbed into this item.

PR #3332, opened 2026-08-19
fix: specialize adapted generic field plans

yaseralnajjar:
## Summary
- substitute concrete owner arguments into finalized adapted field-plan types
- apply the same structural-shape rule to local and imported generic classes
- split generic field specialization from the near-limit structural-shape module
- document the static-program contract

## Validation
- `cargo test -p sifr_frontend` (100 passed)
- `cargo clippy -p sifr_frontend --lib -- -D warnings`
- `cargo fmt --check`
- `git diff --check`
- create-PR gate ran once on `1d15737a6bb89f759cd0967876495f9edd4cbadc`; all preceding guards passed, then the separately owned Rust-interop matrix stopped on the known missing generated-type-import negative source and empty method-slot schema placeholder. It was not rerun.

This is the first bounded M12 compiler hardening item.

### DIFF
--- a/crates/sifr_frontend/src/structural_shape.rs
+++ b/crates/sifr_frontend/src/structural_shape.rs
@@ -19,6 +19,7 @@ use crate::const_canonical::canonical_value;
 use canonical_helpers::canonical_sequence;
 mod defaults;
 use defaults::shape_field_default;
+mod generic_fields;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct StructuralShape {
@@ -362,13 +363,15 @@ fn describe_class(
                     .map(|selection| selection.field_plans.as_slice()),
             )
         };
-    let planned_fields = field_plans.map(|plans| {
-        plans
-            .iter()
-            .map(|field| (field.name.clone(), field.declared_type.clone()))
-            .collect::<Vec<_>>()
-    });
-    let effective_fields = planned_fields.as_deref().unwrap_or(fields);
+    let effective_fields = generic_fields::effective_fields(
+        local_class,
+        source_module,
+        source_name,
+        type_args,
+        fields,
+        field_plans,
+        external_defs,
+    );
     let described_fields = effective_fields
         .iter()
         .enumerate()
--- a/crates/sifr_frontend/src/structural_shape/generic_fields.rs
+++ b/crates/sifr_frontend/src/structural_shape/generic_fields.rs
@@ -1,0 +1,44 @@
+//! Concrete field types selected by finalized adapted generic declarations.
+
+use sifr_lowering::{substitute_type_vars, AdapterFieldPlan, ExternalDefs, HirClass};
+use sifr_type_system::Type;
+use std::collections::HashMap;
+
+#[allow(clippy::too_many_arguments)]
+pub(super) fn effective_fields(
+    local_class: Option<&HirClass>,
+    source_module: &str,
+    source_name: &str,
+    type_args: &[Type],
+    declared_fields: &[(String, Type)],
+    field_plans: Option<&[AdapterFieldPlan]>,
+    external_defs: &ExternalDefs,
+) -> Vec<(String, Type)> {
+    let Some(field_plans) = field_plans else {
+        return declared_fields.to_vec();
+    };
+    let type_params = local_class
+        .map(|class| class.type_params.as_slice())
+        .or_else(|| {
+            external_defs
+                .class_type_params
+                .get(source_module)
+                .and_then(|classes| classes.get(source_name))
+                .map(Vec::as_slice)
+        })
+        .unwrap_or_default();
+    let bindings = type_params
+        .iter()
+        .cloned()
+        .zip(type_args.iter().cloned())
+        .collect::<HashMap<_, _>>();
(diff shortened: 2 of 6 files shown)


yaseralnajjar:
Reviewed exact SHA `1d15737a6bb89f759cd0967876495f9edd4cbadc` (HEAD confirmed, tree clean) against base `b62c92df992afc2d92372ade128375feaf594e03`.

**Verdict: SATISFIED**

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr_frontend/src/structural_shape/generic_fields.rs:37` reuses `sifr_lowering::substitute_type_vars`, which rewrites every matching `Type::TypeVar` name anywhere in the field type, including inside a nested generic declaration that reuses the same parameter name (e.g. an owner `Box[T]` whose plan field type mentions an unrelated inner `T`). This capture behavior is inherited from the shared substitution helper and matches every other call site (`structural_shape/methods.rs:370` path, `early_adapters/inheritance.rs:231`), so it is not introduced here; a scope-name-aware substitution would need to be an owning item on the lowering helper.
- suggestion: `effective_fields` now returns an owned `Vec<(String, Type)>` even on the no-plan path (`generic_fields.rs:17`), replacing what used to be a borrow of the caller's `fields` slice. Correctness is unaffected and the clone is per-described-class, but a `Cow<'_, [(String, Type)]>` would keep the common non-adapted path allocation-free if shape derivation ever shows up hot.
- infrastructure issue: the create-PR gate stopped at two separately owned Rust-interop matrix inputs (missing generated-type-import negative source, empty method-slot schema placeholder). Both are outside this item's changed paths and unrelated to adapter field-plan specialization; per the work boundaries they belong in their owning issue rather than this item.

Basis for the verdict:
- Substitution is applied at the single funnel `describe_class` uses for effective fields (`structural_shape.rs:369-376`), and that same `describe_type_with_externals` entry point feeds static-program specialization (`specialization_runner.rs:64`) and slot-table support (`slot_table.rs:760`), so one substitution point covers both structural shape and static-program specialization as the criterion requires.
- Local and imported owners share one rule: type params come from `HirClass::type_params` for the local case and from `external_defs.class_type_params[module][class]` otherwise (`generic_fields.rs:20-29`), keyed identically to the `class_adapter_selections` lookup that produced `field_plans` (`structural_shape.rs:355-366`) and consistent with the existing method-specialization lookup at `structural_shape.rs:406-410`.
- Nested concrete use is covered by both new regressions: the local `Container.item: Box[int]` case asserts the inner plan field is `Primitive("int")` and that `canonical_identity` has no `param:T` (`structural_shape/tests.rs:347-401`), and the imported case does the same across a `models`/`consumer` boundary after `collect_module_exports` (`structural_shape_import_tests.rs:122-174`).
- Unbound owners are untouched: `zip` over empty `type_args` yields empty bindings, `substitute_type_vars` returns the declared type unchanged, so no runtime static program is newly requested. No transitive generic ancestor handling, imported handler metadata, or package demo files are in the diff.
- Package neutrality and determinism hold: no package identifiers introduced; the `HashMap` is lookup-only and never iterated, plan order (and hence `shape_field_default(index, …)` pairing at `structural_shape.rs:381`) is preserved.
- `substitute_type_vars` recurses through `Type::Class` fields, methods, aliases, and all container/async forms (`sifr_lowering/src/lower/mod_context.rs:701-833`), so deeply nested parameter uses are specialized, not just top-level ones.
- Size guardrail: `structural_shape.rs` 885, `generic_fields.rs` 44, `structural_shape/tests.rs` 653, `structural_shape_import_tests.rs` 347 — all under 900. The contract is documented in `internal_docs/architecture.md:894-898` and `internal_docs/const_specialization.md:144-146`.

yaseralnajjar:
Create-PR gate evidence for exact candidate `1d15737a6bb89f759cd0967876495f9edd4cbadc`:

The gate ran exactly once. Cargo cache setup, HIR maintainability, the 900-line guard, generated-demo freshness, dependency-direction and ownership guards, sysroot/resource checks, stdlib intrinsic/reachability/manifest/bootstrap checks, driver maintainability, and verification self-tests passed. The Rust-interop area then stopped at the separately owned matrix inputs `negative/package_generated_type_import_rejected.sifr` (missing evidence source) and `examples/method_slot_runtime/src/schema_program.sifr` (empty placeholder body). It was not rerun.

yaseralnajjar:
Merge-gate evidence for exact reviewed candidate `1d15737a6bb89f759cd0967876495f9edd4cbadc`:

The merge gate ran exactly once. Cache setup, HIR maintainability, the 900-line guard, generated-demo freshness, dependency-direction and ownership guards, sysroot/resource checks, stdlib intrinsic/reachability/manifest/bootstrap checks, driver maintainability, and verification self-tests passed. The Rust-interop area then stopped at the same separately owned matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` evidence and the empty `examples/method_slot_runtime/src/schema_program.sifr` placeholder. It was not rerun.

Exact-SHA Opus review: SATISFIED, no blocking findings: https://github.com/sifr-lang/sifr/pull/3332#issuecomment-5344383312

PR #3334, opened 2026-08-19
fix: preserve generic inherited handler plans

yaseralnajjar:
## Summary
- specialize checked handler signatures across transitive generic ancestors
- export adapted handler contracts relative to the selected owner
- reconstruct fully imported handler targets, descriptors, order, and metadata
- keep descriptor origins scoped to their declaring source registry
- cover local, fully imported, and imported-transitive cases

## Validation
- `cargo test -p sifr_frontend` (103 passed)
- `cargo test -p sifr_driver --lib adapter_defaults` (14 passed)
- `cargo clippy -p sifr_frontend -p sifr_lowering --lib -- -D warnings`
- formatting, HIR maintainability, file-size, and diff guards passed
- create-PR gate ran once on `8e5f21e487a028bf6b4b5941269ffce4cfde66bf`; all preceding guards passed, then the known separately owned Rust-interop matrix inputs stopped it. It was not rerun.

### DIFF
--- a/crates/sifr_frontend/src/class_method_exports.rs
+++ b/crates/sifr_frontend/src/class_method_exports.rs
@@ -102,24 +102,31 @@ fn adapted_handler_method(
         .cloned()
         .map(Type::TypeVar)
         .collect::<Vec<_>>();
-    let bindings = crate::handler_ancestry::bindings_for_owner(
+    let ancestry = crate::handler_ancestry::resolve(
         module_name,
         owner,
         &root_args,
         callable_owner,
         lowering,
         external_defs,
-    );
-    if bindings.is_none() {
-        return adapted_imported_parent_method(
-            owner,
-            &root_args,
-            handler,
-            local_classes,
-            external_defs,
-        );
-    }
-    let bindings = bindings.unwrap_or_default();
+    )?;
+    let bindings = match ancestry {
+        crate::handler_ancestry::HandlerAncestry::Owner(bindings) => bindings,
+        crate::handler_ancestry::HandlerAncestry::ImportedBoundary {
+            module,
+            name,
+            bindings,
+        } => {
+            return adapted_imported_boundary_method(
+                &module,
+                &name,
+                &bindings,
+                handler,
+                local_classes,
+                external_defs,
+            );
+        }
+    };
     let (source_module, source_name) = callable_owner.rsplit_once('.')?;
     let hir_name = if handler.callable.symbol == "__init__" {
         "new"
@@ -183,57 +190,28 @@ fn adapted_handler_method(
     Some(method)
 }
 
-fn adapted_imported_parent_method(
-    owner: &HirClass,
-    root_args: &[Type],
+fn adapted_imported_boundary_method(
+    boundary_module: &str,
+    boundary_name: &str,
+    boundary_bindings: &HashMap<String, Type>,
     handler: &AdapterHandlerPlan,
     local_classes: &HashMap<String, String>,
     external_defs: &ExternalDefs,
 ) -> Option<StructuralMethodExport> {
-    let root_bindings = owner
-        .type_params
-        .iter()
-        .cloned()
-        .zip(root_args.iter().cloned())
-        .collect::<HashMap<_, _>>();
-    let Type::Class {
-        identity,
-        name,
-        type_args,
-        ..
-    } = owner.parent_type.as_ref()?.resolve_alias()
-    else {
-        return None;
-    };
-    let parent_identity = identity.as_deref().unwrap_or(name);
-    let (parent_module, parent_name) = parent_identity.rsplit_once('.')?;
-    let parent_params = external_defs
-        .class_type_params
-        .get(parent_module)
-        .and_then(|classes| classes.get(parent_name))?;
-    let parent_bindings = parent_params
-        .iter()
-        .cloned()
-        .zip(
-            type_args
-                .iter()
-                .map(|argument| substitute_type_vars(argument, &root_bindings)),
-        )
-        .collect::<HashMap<_, _>>();
     let mut method = external_defs
-        .structural_methods_for(parent_module)?
-        .get(parent_name)?
+        .structural_methods_for(boundary_module)?
+        .get(boundary_name)?
         .iter()
         .find(|method| method.handler_target.as_ref() == Some(&handler.callable))?
         .clone();
     for parameter in &mut method.params {
         parameter.ty = canonicalize_user_export_type(
-            &substitute_type_vars(&parameter.ty, &parent_bindings),
+            &substitute_type_vars(&parameter.ty, boundary_bindings),
             local_classes,
         );
     }
     method.return_type = canonicalize_user_export_type(
-        &substitute_type_vars(&method.return_type, &parent_bindings),
+        &substitute_type_vars(&method.return_type, boundary_bindings),
         local_classes,
     );
     Some(method)
--- a/crates/sifr_frontend/src/class_method_exports.rs
+++ b/crates/sifr_frontend/src/class_method_exports.rs
@@ -1,7 +1,8 @@
 use sifr_lowering::{
-    canonicalize_user_export_type, DeclarationMetadataTargetKind, ExternalDefs, HirClass,
-    MethodKind, StructuralMethodExport,
+    canonicalize_user_export_type, substitute_type_vars, AdapterHandlerPlan,
+    DeclarationMetadataTargetKind, ExternalDefs, HirClass, MethodKind, StructuralMethodExport,
 };
+use sifr_type_system::Type;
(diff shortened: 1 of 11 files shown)


yaseralnajjar:
Reviewed exact SHA `8e5f21e487a028bf6b4b5941269ffce4cfde66bf` (HEAD verified clean, base `2700b807…`), diff scope matches the declared paths.

What I checked against the criteria:
- `crates/sifr_frontend/src/handler_ancestry.rs:252` — `bindings_for_owner` walks the local ancestor chain, substituting each level's `type_args` through the accumulated bindings, with a `visiting` cycle guard, and terminates on an imported owner by zipping `external_defs.class_type_params`. Replaces the former single-hop `inherited_handler_bindings` / `imported_handler_bindings` in `structural_shape/methods.rs`; both call sites (local and imported handler detail) now use it with `root_args` derived from the concrete bindings.
- `class_method_exports.rs:97` — adapted handler contracts are exported with `handler_target` and signatures substituted relative to the selected owner's type params (identity map at the root, ancestor-composed above it), plus an imported-direct-parent fallback at `:192`. Removing the `names_by_class.is_empty()` early return is required so handler-only classes still export; it does not widen ordinary method export.
- `structural_shape/methods.rs:458+` — imported reconstruction filters the duplicate plain entry via `own_handler_names` (keyed on the exact `source_identity` from `structural_shape.rs:288`), emits handlers ordered by `declaration_order` with a stable sort, restores descriptor/target/order and `callable.module`-scoped method and parameter metadata, and forces `origin = None`, so no foreign `SourceOriginId` leaks into another declaration's registry.
- Determinism and package neutrality preserved: no package-policy branching, no new global state, ordering derived from export order plus `declaration_order`.
- New tests cover local transitive ancestry (`structural_shape/tests.rs:510`), imported-transitive-through-generic-parent and fully-imported concrete reconstruction (`structural_shape_import_tests.rs:176`, `:889`), asserting concrete param/result types, descriptor, order, metadata, `origin == None`, and absence of `param:T`/`param:U` in the canonical identity. Docs in `architecture.md` / `const_specialization.md` match the implemented rule.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_frontend/src/handler_ancestry.rs:335` and the fallbacks at `class_method_exports.rs:192` / `structural_shape/methods.rs:532` compose only *local chain → direct imported parent*. A chain with a local intermediate and a multi-hop imported tail (`Child → localMid → importedParent → importedGrand`, handler owned by `importedGrand`) makes `walk_local` return `None`, and the fallback restarts from the root class's own `parent_type` (the local `Mid`), whose module has no `class_type_params` entry — so the handler degrades to the `ShapeMethod` placeholder (`kind: "inherited"`, empty params, `checked_handler` result) instead of a substituted contract. Not a regression (the prior single-hop code leaked unsubstituted `TypeVar`s in the same shape) and outside the two coverage cases this item enumerates; the fix is to have the fallback bind at the deepest local class reached rather than at the root.
- suggestion: `structural_shape/methods.rs:335` still passes empty declaration metadata for the local-class-with-imported-handler path, so that path restores concrete types but not the handler's method/parameter metadata. Pre-existing behavior, unchanged by this candidate; only the fully-imported path is covered by the criterion.
- infrastructure issue: the create-PR gate on this SHA stopped on two separately owned Rust-interop matrix inputs after all guards and generated-demo freshness passed. Owned elsewhere; no in-scope evidence gap for this item.

yaseralnajjar:
HEAD is `4ebffb076` (matches the stated candidate), tree clean. Reviewed `2700b8078..4ebffb076` with focus on `8e5f21e48..4ebffb076` (4 files: `handler_ancestry.rs`, `class_method_exports.rs`, `structural_shape/methods.rs`, `structural_shape_import_tests.rs`). Independently re-ran `cargo test -p sifr_frontend`: 103 passed.

## Verdict: SATISFIED

**Blocking findings:** none.

### Mechanism check
`handler_ancestry::resolve` (`handler_ancestry.rs:16-114`) now returns `Owner(bindings)` or `ImportedBoundary { module, name, bindings }` instead of `Option<bindings>` collapsing every non-owner ancestry to `None`. The boundary is produced only at the point where the local parent chain terminates in an imported class (`:99-113`), after `parent_args` has been substituted through every accumulated local hop (`:73-76`), so the bindings are the imported boundary class's own type params mapped to fully concrete arguments. Both consumers then look up the boundary's already-exported adapted contract by exact `handler_target` equality (`class_method_exports.rs:201-206`, `methods.rs:423-427`) — a match there means the boundary genuinely inherits that handler, and a non-match falls through to `None`/placeholder exactly as before. That is a sound narrowing: no path that previously resolved can now fail, and the boundary export was itself generated with the owner's params as `TypeVar`s (`class_method_exports.rs:99-104`), so a single substitution pass is the correct and complete one.

Duplicated ancestry-walking logic in both `adapted_imported_parent_method` and `imported_handler_detail` is gone; the two call sites now share one resolver. No stale references to `bindings_for_owner`/`adapted_imported_parent_method` remain. All touched files are well under 900 lines (max 737).

### Prior suggestions
1. **Deepest-local + multi-hop imported tail — resolved.** `imported_transitive_generic_handler_uses_the_concrete_child_argument` (`structural_shape_import_tests.rs:265-408`) is now `Child -> Mid[int]` (local, generic) `-> models.Parent[U]` (imported) `-> models.Grand[list[U]]` (imported owner). The boundary is captured at `Mid`'s imported parent, and the asserted param/result are `list[int]` (`:395-402`), not the `checked_handler` placeholder. Under the pre-remediation code the fallback read `class.parent_type` (local `Mid`) and looked `Mid` up in `external_defs.class_type_params["consumer"]`, which is absent — so this test genuinely fails without the fix.
2. **Imported-handler declaration metadata — resolved.** `methods.rs:327-349` now sources metadata from `external_defs.declaration_metadata[handler.callable.module]` keyed on `{owner_name}.{declared_name}`, which is exactly the convention already used by `described_exported_methods` (`:502-524`) and produced by `structural_method_map` (`class_method_exports.rs:238`) — reuse of the established pattern rather than a parallel one. Asserted at `:404-405`.

### Follow-up findings (later item, non-blocking)
- **suggestion:** `structural_shape_import_tests.rs:45` adds `@metadata("parameter", "value", "fixture.role", "input")` to the `imported_generic_parent_handler_uses_the_concrete_child_argument` fixture, but that test asserts no metadata (`:115-123`). The direct owner-is-imported-parent case therefore carries an unexercised fixture decorator; either assert it there or drop it.
- **pre-existing issue:** the local-detail branch maps `ImportedBoundary` (and `None`) to empty bindings (`methods.rs:289-291`), behaviorally identical to the previous `unwrap_or_default()`. If a plan ever names a local handler owner that is not on the class's local ancestry chain, unsubstituted `TypeVar`s leak into the described shape. Not introduced here and not reachable from any current fixture.
- **pre-existing issue:** `adapted_imported_boundary_method` re-applies `canonicalize_user_export_type` with the *consumer's* `local_classes` over types the boundary module already canonicalized (`class_method_exports.rs:207-216`), carried over unchanged from `adapted_imported_parent_method`. Benign today; a cross-module class-name collision is the theoretical exposure.
- **infrastructure issue:** the create-PR gate was not rerun on this exact candidate. Its one allowed run on `8e5f21e48` cleared every preceding guard and stopped only at the separately owned Rust-interop matrix inputs; the remediation touches only `sifr_frontend` shape/export code plus its tests, and the frontend suite plus focused tests pass on `4ebffb076`. No third review needed on this basis.

yaseralnajjar:
Create-PR gate evidence for initial candidate `8e5f21e487a028bf6b4b5941269ffce4cfde66bf`:

The private target exceeded 20 GiB, was verified idle, and was cleaned before the gate. The gate ran exactly once. Cache setup, HIR maintainability, the 900-line guard, generated-demo freshness, dependency-direction and ownership guards, sysroot/resource checks, stdlib intrinsic/reachability/manifest/bootstrap checks, driver maintainability, and verification self-tests passed. The Rust-interop area then stopped at the separately owned missing generated-type-import negative source and empty method-slot schema placeholder. It was not rerun after the focused remediation, per the gate cap.

yaseralnajjar:
Merge-gate evidence for exact final reviewed candidate `4ebffb07665217e49e702c5a7e30162f2c395bd0`:

The merge gate ran exactly once. Cache setup, HIR maintainability, the 900-line guard, generated-demo freshness, dependency-direction and ownership guards, sysroot/resource checks, stdlib intrinsic/reachability/manifest/bootstrap checks, driver maintainability, and verification self-tests passed. The Rust-interop area then stopped at the same separately owned missing generated-type-import negative source and empty method-slot schema placeholder. It was not rerun.

Final remediation review: SATISFIED, no blocking findings: https://github.com/sifr-lang/sifr/pull/3334#issuecomment-5344743657

PR #3338, opened 2026-08-19
Respect nested generic scopes in adapter fields

yaseralnajjar:
## Summary
- add declaration-aware type substitution for nested nominal scopes
- use local and exported class parameter ownership for adapted field plans
- add local and imported same-name generic scope regressions
- document the nested scope contract

## Validation
- `cargo test -p sifr_lowering --lib --no-fail-fast`: 1004 passed, 1 ignored
- `cargo test -p sifr_frontend --lib --no-fail-fast`: 107 passed
- `cargo test -p sifr_driver tests::adapter_defaults --no-fail-fast`: 15 passed
- `cargo fmt --check`
- create-PR gate (single run): stopped only at the externally owned missing Rust-interop negative fixture and empty schema-program placeholder

### DIFF
--- a/crates/sifr_frontend/src/structural_shape.rs
+++ b/crates/sifr_frontend/src/structural_shape.rs
@@ -370,6 +370,7 @@ fn describe_class(
         type_args,
         fields,
         field_plans,
+        lowering,
         external_defs,
     );
     let described_fields = effective_fields
--- a/crates/sifr_frontend/src/structural_shape/generic_fields.rs
+++ b/crates/sifr_frontend/src/structural_shape/generic_fields.rs
@@ -1,6 +1,9 @@
 //! Concrete field types selected by finalized adapted generic declarations.
 
-use sifr_lowering::{substitute_type_vars, AdapterFieldPlan, ExternalDefs, HirClass};
+use sifr_lowering::{
+    substitute_type_vars_with_class_scopes, AdapterFieldPlan, ExternalDefs, HirClass,
+    LoweringResult,
+};
 use sifr_type_system::Type;
 use std::collections::HashMap;
 
@@ -12,6 +15,7 @@ pub(super) fn effective_fields(
     type_args: &[Type],
     declared_fields: &[(String, Type)],
     field_plans: Option<&[AdapterFieldPlan]>,
+    lowering: &LoweringResult,
     external_defs: &ExternalDefs,
 ) -> Vec<(String, Type)> {
     let Some(field_plans) = field_plans else {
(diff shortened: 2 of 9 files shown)


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_lowering/src/lower/scoped_type_substitution.rs:7` now duplicates the whole variant match of `substitute_type_vars` (`crates/sifr_lowering/src/lower/mod_context.rs:677`) — ~120 lines that must stay in sync across two files, and a missed variant in either one silently degrades to `_ => ty.clone()`. The old function is behaviorally identical to the new one called with a resolver that always returns an empty parameter list (no params ⇒ no removal, no rebinding), so it could delegate instead of re-listing every variant. Variant coverage was verified equal in this diff.
- pre-existing issue: both substitution functions clone `parent_class` without substituting it (`scoped_type_substitution.rs:141`, `mod_context.rs:830`), so a generic ancestor's arguments recorded on the nominal type stay unspecialized. Unchanged by this diff.
- infrastructure issue: `crates/sifr_frontend/src/structural_shape/generic_fields.rs:83` — the third resolution branch compares `source_module == module_name` and then searches `lowering.module.classes` by bare name, but `module_name` here is the *owner's* source module while `lowering` is the *describing* module's result. For a local owner these coincide (and this is the branch the local test actually exercises, since `lower_module` leaves `identity` unset); for an imported owner they diverge, so a nested identity that is absent from both `lowering` and `external_defs.class_type_params` can pick up a same-named consumer class's parameters. Not a regression — the base unconditionally captured with the outer bindings — and unreachable while export collection is complete, but it is the one name-based path remaining under the "no name-collision fallback" criterion. Also, when nothing resolves (`Vec::new()`, e.g. a nested class from a module the consumer never imports), no parameters are removed, so an outer same-name binding still reaches the nested fields; same as base behavior, worth recording in the owning issue rather than fixing here.
- suggestion: the three new tests build `Type::Class` with `type_args: [Str]` while `fields` still hold `TypeVar("T")` — a state `annotation_resolution.rs:459-476` does not produce for source-level `Inner[str]`, since it substitutes the inner declaration's parameters at annotation time. The scope rule is therefore locked at the substitution-function and shape-derivation layers, not from Sifr source through to a described shape. This matches how neighboring tests in both files construct `ClassAdapterSelection`, so it is consistent, not a defect.
- pre-existing issue: the `Type::Alias` arm substitutes `body` with the outer bindings and does not rebind alias parameters. Out of scope (the item covers nominal declarations), and alias bodies are already substituted during annotation resolution.
- infrastructure issue: the two create-PR gate stops (missing `negative/package_generated_type_import_rejected.sifr`, placeholder `examples/method_slot_runtime/src/schema_program.sifr`) are externally owned interop matrix inputs, as stated in the evidence; not attributable to this item.

No `pre_v1` paths appear in the diff, and the architecture (`internal_docs/architecture.md:899`) and const-specialization (`internal_docs/const_specialization.md:153`) additions state the scope rule and the same-name non-capture guarantee.

yaseralnajjar:
Create-PR gate evidence for exact candidate `5a370712ecac97ab34cef90409f11f3173fa3a3a` (single permitted run): setup and all preceding guardrails passed. The gate then stopped in the externally owned Rust interop matrix because `negative/package_generated_type_import_rejected.sifr` is missing and `examples/method_slot_runtime/src/schema_program.sifr` remains an empty placeholder. The gate was not rerun and those inputs were not changed.

yaseralnajjar:
Merge-gate evidence for exact reviewed candidate `5a370712ecac97ab34cef90409f11f3173fa3a3a` (single permitted run): setup and all preceding guardrails passed. The gate stopped at the same externally owned Rust interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and placeholder `examples/method_slot_runtime/src/schema_program.sifr`. It was not rerun.

PR #3340, opened 2026-08-19
Unify generic substitution authority

yaseralnajjar:
## Summary
- make the legacy type substitution API delegate to the single recursive visitor
- distinguish local declaration authority from imported exported metadata
- keep unresolved nested scopes symbolic instead of capturing outer bindings
- add a consumer name-collision regression and document fail-closed behavior

## Validation
- `cargo test -p sifr_lowering --lib --no-fail-fast`: 1005 passed, 1 ignored
- `cargo test -p sifr_frontend --lib --no-fail-fast`: 107 passed
- `cargo test -p sifr_driver tests::adapter_defaults --no-fail-fast`: 15 passed
- workspace Clippy, formatting, HIR maintainability, file-size guardrail, diff hygiene
- create-PR gate (single run): stopped only at the externally owned missing Rust-interop negative fixture and empty schema-program placeholder


yaseralnajjar:
**Verdict: SATISFIED**

Checked the diff against each criterion:

- **Single visitor / one variant match** — `mod_context.rs`'s legacy 160-line match is gone; `substitute_type_vars` (`scoped_type_substitution.rs:7`) delegates with `&|_, _| Some(Vec::new())`. I compared the removed arms against the surviving visitor arm-by-arm: identical variant coverage (including `JoinSet`, `Alias`, `make_union` for `Union`, `_ => clone`). With `Some(vec![])`, `nested_bindings` is an unmodified clone of `bindings`, so `Type::Class` fields/methods substitute exactly as before — behavior preserved. Re-export move (`mod.rs:173`) keeps `sifr_lowering::substitute_type_vars` intact for all ~20 existing call sites.
- **Local vs imported owner authority** — `local_class.is_some()` threads through as `owner_is_local`; the bare-name branch (`generic_fields.rs:92`) now requires it. Confirmed at `structural_shape.rs:291-300` that `local_class` is `Some` only when the nested identity equals `module_name.local_name`, so `owner_is_local && source_module == module_name` is genuinely the local-declaration case, while an imported owner can no longer reach a consumer-local class of the same name.
- **Fail-closed unresolved scope** — `Option` return plus `nested_bindings.clear()` (`scoped_type_substitution.rs:104`). `type_args` are computed with the outer `recurse` at line 96, before the clear, so nested arguments still specialize in the outer scope while nested fields/methods stay symbolic. Matches the criterion exactly.
- **Tests** — the unit test pair covers resolved shadowing (`Str` wins over outer `Int`) and unresolved protection (`T` stays symbolic). The frontend addition removes `models.Inner` from exported metadata *and* adds a colliding consumer-local `Inner[Left, Right]`; the assertion (`TypeParameter("T")`) distinguishes new behavior from the old fallback, which would have yielded `int`.
- **Docs** — both `architecture.md:901` and `const_specialization.md:155` state the authority and fail-closed rule.
- No `parent_class`/alias-scope work and no `pre_v1` changes leaked in.

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr_codegen/src/structural_record_fields.rs:42` — `substitute_structural_type` is a third, nested-scope-unaware substitution visitor (it also builds `Type::Union` directly instead of via `make_union`). Same-name outer capture is still possible on that path. Outside this item's changed paths; worth an entry in the owning issue.
- suggestion: `substitute_type_vars` now clones the bindings map at every `Type::Class` node (`scoped_type_substitution.rs:97`) even though its resolver always returns an empty parameter list. A guard that skips the clone when `declared_params` is empty would remove a per-class allocation from a hot lowering path.
- suggestion: neither new test exercises the "nested type arguments still specialize in the outer scope" half of the unresolved rule — both unresolved cases use concrete or empty `type_args`. A single `type_args: [TypeVar("T")]` case asserting `Int` in the args while the field stays `T` would pin that guarantee. The unresolved frontend case likewise doesn't re-assert the sibling `direct: T` field still resolves to `int`.
- infrastructure issue: the create-PR gate's two stopping inputs (missing `negative/package_generated_type_import_rejected.sifr`, placeholder `examples/method_slot_runtime/src/schema_program.sifr`) are externally owned interop matrix inputs, as stated — not treated as item defects.

yaseralnajjar:
Create-PR gate evidence for exact candidate `a1ae3080d2cd1be2aa5c923c63486080e041f561` (single permitted run): setup and all preceding guardrails passed. The gate then stopped in the externally owned Rust interop matrix because `negative/package_generated_type_import_rejected.sifr` is missing and `examples/method_slot_runtime/src/schema_program.sifr` remains an empty placeholder. The gate was not rerun and those inputs were not changed.

yaseralnajjar:
Merge-gate evidence for exact reviewed candidate `a1ae3080d2cd1be2aa5c923c63486080e041f561` (single permitted run): setup and all preceding guardrails passed. The gate stopped at the same externally owned Rust interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and placeholder `examples/method_slot_runtime/src/schema_program.sifr`. It was not rerun.

PR #3348, opened 2026-08-19
Reject transitive generic union topology changes

yaseralnajjar:
## Summary
- select generic parent templates by canonical identity
- follow nested generic declarations in the shared union-topology check
- reject transitive representation changes before Rust code generation

## Validation
- cargo test -p sifr_type_system --no-fail-fast
- cargo test -p sifr_codegen --no-fail-fast
- cargo fmt --all -- --check
- maintainability, file-size, and diff guards
- create-PR gate ran once on e2ff698497de3b902f81d1ad451fc2b97fda2a82; all preceding checks passed before the separately owned Rust-interop matrix stop

### DIFF
--- a/crates/sifr_codegen/src/structural_impl_codegen.rs
+++ b/crates/sifr_codegen/src/structural_impl_codegen.rs
@@ -236,7 +236,7 @@ fn structural_parent_supported(
     let Some(parent) = structural_class_candidate(identity.as_deref(), name, modules) else {
         return false;
     };
-    generic_arguments_preserve_union_structure(parent, type_args)
+    generic_arguments_preserve_union_structure(parent, type_args, modules)
         && parent.parent_class.is_none()
         && !parent.methods.iter().any(|method| method.name == "new")
         && !parent.is_error_type
@@ -310,15 +310,15 @@ fn structural_type_supported(
             ..
         } => {
             let key = identity.as_deref().unwrap_or(name);
+            if visiting.contains(key) {
+                return true;
+            }
             let Some(candidate) = structural_class_candidate(identity.as_deref(), name, modules)
             else {
                 return false;
             };
-            if !generic_arguments_preserve_union_structure(candidate, type_args) {
+            if !generic_arguments_preserve_union_structure(candidate, type_args, modules) {
                 return false;
-            }
-            if visiting.contains(key) {
-                return true;
             }
             if structural_mapped_opaque_supported(candidate) {
                 return true;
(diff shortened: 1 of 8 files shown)


yaseralnajjar:
Exact candidate SHA: e2ff698497de3b902f81d1ad451fc2b97fda2a82

Verdict: SATISFIED

Blocking findings:
None.

Reviewed against the diff itself (`git diff 9247fd98..e2ff6984`), spot-checking the pieces the criteria hinge on:

- Canonical parent selection: `crates/sifr_lowering/src/lower/generic_parent_representation.rs:52-72` now resolves the template through `class_template`/`class_candidate`, matching `Type::Class { identity }` across `class_types.values()` and only falling back to the consumer-local name when the concrete parent carries no identity. The new `canonical_identity_selects_the_parent_and_finds_transitive_hazards` test pins the colliding-local case.
- Nested rebinding: `crates/sifr_type_system/src/substitution.rs` `preserves_union_structure` Class arm substitutes the occurrence's `type_args` under the outer bindings, then zips them onto the nested declaration's own `type_params` before descending into `member_types`, with a `visiting` set inserted/removed around the descent (no cross-sibling suppression).
- Transitive rejection before codegen: driver test `transitive_nested_generic_union_parent_is_rejected_before_rust_codegen` asserts `CLASS_INVALID_BASE` with the topology message; lowering reaches it via `class_type_collection.rs:317-324`, which also removes the prior `is_some_and` fail-open on a missing template.
- Structural parity: `structural_impl_codegen.rs` routes both parent and nested-type checks through the same `substitution_preserves_union_structure_with_class_scopes`, with `generic_union_scope` supplying HIR `type_params` + field/param/return members.
- Fast-path reorder: moving `visiting.contains(key)` ahead of the candidate lookup in `structural_type_supported` is safe — a key is only in `visiting` after a successful lookup with identical `(identity, name)` inputs, and the shallow re-entrant union check it skips is now subsumed by the recursive scope-aware check performed on first visit.
- Docs: `internal_docs/architecture.md:911-916` states the canonical-identity selection and transitive rule.

Follow-up findings:
- suggestion: `class_union_scope` (`generic_parent_representation.rs`) derives `type_params` by `filter_map`-ing `type_args` down to `TypeVar`s, then the type-system side zips that filtered list against the full concrete argument list. For a template entry whose `type_args` are not all type vars (e.g. `[Str, TypeVar("U")]`), the zip misaligns and binds `U` to the wrong argument. The codegen side avoids this by using HIR `class.type_params` directly; consider carrying declared params instead of recovering them from `type_args`.
- suggestion: `class_candidate` uses `class_types.values().find(...)` and `structural_class_candidate` rescans `modules` on every nested class encountered; with deep nesting this is a repeated linear scan per node. An identity-keyed index would remove the quadratic factor if generic nesting depth grows.
- infrastructure issue: the create-PR gate stopped on the two externally owned Rust-interop matrix inputs (`negative/package_generated_type_import_rejected.sifr` missing, `examples/method_slot_runtime/src/schema_program.sifr` empty), so the gate has no green end-to-end run on this candidate. Out of scope per the item, but it remains unverified past that point.

yaseralnajjar:
Create-PR gate evidence for exact candidate e2ff698497de3b902f81d1ad451fc2b97fda2a82: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the separately owned Rust-interop matrix because negative/package_generated_type_import_rejected.sifr is missing and examples/method_slot_runtime/src/schema_program.sifr has an empty placeholder body. Per scope, neither input was inspected or changed, and the gate was not rerun.

yaseralnajjar:
Merge-gate evidence for exact candidate e2ff698497de3b902f81d1ad451fc2b97fda2a82: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the same separately owned Rust-interop matrix inputs: missing negative/package_generated_type_import_rejected.sifr and empty examples/method_slot_runtime/src/schema_program.sifr. Per scope, the gate was not rerun.

PR #3350, opened 2026-08-19
Use declared generic parameters for topology checks

yaseralnajjar:
## Summary
- bind generic parent checks from declared class-parameter metadata
- carry canonical candidate keys through nested class resolution
- reject nested class argument arity mismatches in the shared predicate

## Validation
- all affected type-system, lowering, codegen, frontend, and driver suites
- create-PR gate ran once on 6b93320b8eccfd2b54074098859b162dfb476cea and stopped only at the separately owned Rust-interop inputs after all preceding guards passed

### DIFF
--- a/crates/sifr_lowering/src/lower/classes/class_type_collection.rs
+++ b/crates/sifr_lowering/src/lower/classes/class_type_collection.rs
@@ -319,6 +319,7 @@ pub(in crate::lower) fn collect_class_type(
         {
             if !super::super::generic_parent_representation::preserves_union_structure(
                 &ctx.class_types,
+                &ctx.class_declared_type_params,
                 parent_name,
                 &parent_ty,
             ) {
--- a/crates/sifr_lowering/src/lower/generic_parent_representation.rs
+++ b/crates/sifr_lowering/src/lower/generic_parent_representation.rs
@@ -5,18 +5,16 @@ use std::collections::HashMap;
 
 pub(super) fn preserves_union_structure(
     class_types: &HashMap<String, Type>,
+    class_type_params: &HashMap<String, Vec<String>>,
     parent_name: &str,
     concrete: &Type,
 ) -> bool {
-    let Some(template) = class_template(class_types, concrete, parent_name) else {
+    let Some((template_name, template)) = class_template(class_types, concrete, parent_name) else {
         return true;
     };
     let (
         Type::Class {
-            type_args: template_args,
-            fields,
-            methods,
-            ..
+            fields, methods, ..
         },
         Type::Class {
             type_args: concrete_args,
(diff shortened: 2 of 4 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate 6b93320b8eccfd2b54074098859b162dfb476cea: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the separately owned Rust-interop matrix inputs. The gate was not rerun.

yaseralnajjar:
Exact candidate SHA: 6b93320b8eccfd2b54074098859b162dfb476cea

Reviewed the exact diff and traced the binding-authority paths.

**What I verified**

- `generic_parent_representation.rs:8-58` — `class_template`/`class_candidate` now return the selected local declaration key alongside the type, and both the top-level bindings and every nested `class_union_scope` draw `type_params` from `ctx.class_declared_type_params` for *that* key. Stored template `type_args` no longer participate in binding, so a misleading concrete stored argument (new test `declared_parameters_are_the_parent_binding_authority`) can't suppress the hazard. Positional binding of all declared params replaces the old `filter_map(TypeVar)`, which is what closed the hole.
- Key/metadata consistency: every `class_types.insert` site for imported classes (`imports.rs:372`, `mod_impl.rs:492/669`, `private_stdlib_imports.rs:119/143`) registers `class_declared_type_params` under the *same* local key, so the identity-selected key resolves to real parameter metadata for both imported and local parents.
- Arity rejects can't fire on currently valid programs: `annotation_resolution.rs:449-476` already validates subscripted class arity against the same `class_declared_type_params`, and the class types built in `class_type_collection.rs:378` / `887` derive `type_args` from declared params. Bare generic parents (`class Child(Parent)`) keep arity 1:1 via the declared-param TypeVars, and non-generic parents are 0:0.
- `substitution.rs:257` rejects nested arity mismatch before building `nested_bindings`, and correctly removes the `visiting` key before returning, so the recursion guard isn't leaked.
- Codegen's `structural_impl_codegen.rs:374` already fed declared `HirClass::type_params` into the scope, so the shared predicate stays aligned; snapshot-bearing codegen/driver suites pass on the candidate.
- Architecture doc records the invariant (`internal_docs/architecture.md:917-918`).

Verdict: SATISFIED

Blocking findings:
None.

Follow-up findings:
- suggestion: `crates/sifr_lowering/src/lower/generic_parent_representation.rs:78` — identity lookup uses `class_types.iter().find_map`, so when two local keys alias the same canonical identity the selected declaration key is HashMap-order dependent. Today all such keys carry identical parameter metadata, so behavior is stable, but the selection is now metadata-bearing; a deterministic tie-break (e.g. prefer the passed `name`, then sorted key) would make it order-independent by construction. Nondeterministic selection predates this diff.
- suggestion: `crates/sifr_lowering/src/lower/classes/class_type_collection.rs:325` — if the new arity reject ever fires for a parent, the diagnostic still reads "generic parent arguments change a declared union's member topology", which would misdescribe an arity mismatch. Unreachable for valid input today (annotation resolution errors first), so this is only about message fidelity if that ordering changes.
- infrastructure issue: the single create-PR gate stopped at two externally owned Rust-interop matrix inputs and was not rerun; that belongs to the owning issue, not this item.

yaseralnajjar:
Merge-gate evidence for exact candidate 6b93320b8eccfd2b54074098859b162dfb476cea: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the same separately owned Rust-interop matrix inputs. The gate was not rerun.

PR #3352, opened 2026-08-19
Require exact generic ancestor bindings

yaseralnajjar:
## Summary
- require exact parameter and argument arity while walking generic handler ancestry
- preserve explicit symbolic arguments for valid unspecialized exports
- pin generic alias argument and body specialization agreement

## Validation
- 139 type-system tests
- 109 frontend tests
- 540 driver tests with 76 ignored
- create-PR gate ran once on e4e458766e26b03a6fcb99f29c82c4b535ebe206 and stopped only at the separately owned Rust-interop inputs after all preceding guards passed

### DIFF
--- a/crates/sifr_frontend/src/handler_ancestry.rs
+++ b/crates/sifr_frontend/src/handler_ancestry.rs
@@ -21,12 +21,7 @@ pub(crate) fn resolve(
     lowering: &LoweringResult,
     external_defs: &ExternalDefs,
 ) -> Option<HandlerAncestry> {
-    let bindings = root
-        .type_params
-        .iter()
-        .cloned()
-        .zip(root_type_args.iter().cloned())
-        .collect();
+    let bindings = bind_exact(&root.type_params, root_type_args.iter().cloned())?;
     walk_local(
         module_name,
         root,
@@ -80,12 +75,7 @@ fn walk_local(
                 && parent_identity == format!("{module_name}.{}", candidate.name))
     });
     if let Some(parent) = local_parent {
-        let parent_bindings = parent
-            .type_params
-            .iter()
-            .cloned()
-            .zip(parent_args)
-            .collect();
+        let parent_bindings = bind_exact(&parent.type_params, parent_args)?;
         return walk_local(
             module_name,
             parent,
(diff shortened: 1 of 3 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate e4e458766e26b03a6fcb99f29c82c4b535ebe206: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the separately owned Rust-interop matrix inputs and was not rerun.

yaseralnajjar:
Exact candidate SHA: e4e458766e26b03a6fcb99f29c82c4b535ebe206

Verdict: SATISFIED

Blocking findings:
- None.

Notes from verification: `bind_exact` (crates/sifr_frontend/src/handler_ancestry.rs:105) is applied at the root, the local parent edge, and the imported parent edge, so all three use the same exact-arity rule and no truncating `zip` remains. All three `resolve` call sites (`class_method_exports.rs:100`, `structural_shape/methods.rs:246`, and the `root_args` it forwards to `imported_handler_detail`) build root arguments from `class.type_params`, mapping unbound parameters to `Type::TypeVar`, so valid unspecialized exports stay symbolic and complete rather than arity-mismatched. The alias arm at `crates/sifr_type_system/src/substitution.rs:77` applies the same `recurse` (same outer bindings) to `type_args` and `body`, which is exactly what the new test pins.

Follow-up findings:
- pre-existing issue: `crates/sifr_frontend/src/structural_shape/methods.rs:290` maps a `None`/`ImportedBoundary` ancestry to `HashMap::new()`. With the stricter arity rule, an arity-mismatched edge now yields empty inherited bindings instead of partial ones — correct per the criterion, but the silent empty-map fallback still hides the malformed input rather than surfacing it.
- suggestion: the new unit tests cover the root edge only (missing root args, symbolic local root-to-parent). A case exercising the imported edge via `ExternalDefs::class_type_params` with mismatched arity would pin the "local and imported use the same rule" criterion directly rather than by shared-helper inspection.
- infrastructure issue: the single create-PR gate stopped at two externally owned Rust-interop inputs and was not rerun; that gap is outside this item's ownership.

yaseralnajjar:
Merge-gate evidence for exact candidate e4e458766e26b03a6fcb99f29c82c4b535ebe206: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the same separately owned Rust-interop matrix inputs and was not rerun.

PR #3354, opened 2026-08-19
Fail closed on unresolved handler ancestry

yaseralnajjar:
## Summary
- remove partial inherited-signature substitution when ancestry resolution fails
- retain an explicit opaque checked-handler contract for unresolved metadata
- add direct imported arity-mismatch coverage
- split ancestry regressions into their own test module

## Validation
- 111 frontend tests
- 540 driver tests with 76 ignored
- create-PR gate ran once on 15d8bda3c50d664c37881864ff8535484c787052 and stopped only at the separately owned Rust-interop inputs after all preceding guards passed

### DIFF
--- a/crates/sifr_frontend/src/handler_ancestry.rs
+++ b/crates/sifr_frontend/src/handler_ancestry.rs
@@ -115,7 +115,8 @@ fn bind_exact(
 #[cfg(test)]
 mod tests {
     use super::{resolve, HandlerAncestry};
-    use sifr_lowering::{lower_module, ExternalDefs};
+    use crate::{collect_module_exports, compile_module_hir, FrontendDiagnosticStyle};
+    use sifr_lowering::{lower_module, ExternalDefs, LoweringResult};
     use sifr_syntax::parse_module_suite;
     use sifr_type_system::Type;
 
@@ -126,6 +127,17 @@ mod tests {
         )
         .expect("fixture parses");
         lower_module(&parsed).expect("fixture lowers")
+    }
+
+    fn compile(module: &str, source: &str, external_defs: &ExternalDefs) -> LoweringResult {
+        let parsed = parse_module_suite(source, None).expect("fixture parses");
+        compile_module_hir(
+            module,
+            &parsed,
+            external_defs,
+            FrontendDiagnosticStyle::Bare,
+        )
+        .expect("fixture compiles")
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate 15d8bda3c50d664c37881864ff8535484c787052: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the separately owned Rust-interop matrix inputs and was not rerun. The phase issue already records this external blocker and its ownership boundary.

yaseralnajjar:
Exact candidate SHA: 15d8bda3c50d664c37881864ff8535484c787052

## Review

Verified the diff against the acceptance criteria.

**Key checks:**
- `methods.rs:257-284` — the resolve call moved inside the `local_detail` chain and now gates it: only `HandlerAncestry::Owner(bindings)` produces a local detail; the previous `_ => HashMap::new()` silent-empty fallback is gone. The `owner` binding used for `owner_identity` is the unwrapped `callable_owner`, which is exactly what `callable_owner.unwrap_or(&local_owner)` resolved to on this branch (local_detail only exists when `callable_owner` is `Some`), so successful ancestry is unchanged.
- Fall-through on failure reaches `imported_handler_detail` (which re-resolves and fails or finds no matching external export for a local owner) and lands in the pre-existing `else` arm at `methods.rs:357-373`, emitting `ShapeNode::Other("checked_handler")` with empty params; `target` and `descriptor` are then reassigned at `methods.rs:375-377`, so the checked-handler contract is opaque but identity-preserving.
- `ancestry_tests.rs` clears `Parent[int]`'s args so `bind_exact(["T"], [])` fails, asserting the opaque contract plus preserved `target`.
- `handler_ancestry.rs:179-205` pushes an extra `"V"` param onto imported `models.Base` metadata against a one-arg `Base[U]` base clause, exercising `bind_exact` at the imported boundary (`handler_ancestry.rs:94`) → `None`.
- `structural_shape.rs` is 896 lines with the new tests carved into `structural_shape/ancestry_tests.rs`; docs record the durable rule at `internal_docs/architecture.md:899-900`.

Verdict: SATISFIED

Blocking findings:
None.

Follow-up findings:
- infrastructure issue: the create-PR gate stopped at two externally owned Rust-interop inputs and was not rerun; the diff contains no record of that blocker in its owning issue, which Work Boundaries requires for out-of-scope failures.
- suggestion: on failed local ancestry, `described_handler` resolves ancestry, then `imported_handler_detail` resolves the same `(class, root_args, owner)` a second time before falling through. Passing the first result through would avoid the duplicate walk.
- suggestion: `unresolved_handler_ancestry_uses_an_opaque_checked_contract` asserts the preserved `target` but not the preserved `descriptor`, which is also part of the contract the scope names; adding a `descriptor` assertion would lock the whole contract.

yaseralnajjar:
Merge-gate evidence for exact candidate 15d8bda3c50d664c37881864ff8535484c787052: the gate ran once. Generated-demo freshness and every preceding guard passed. It stopped at the same separately owned Rust-interop matrix inputs and was not rerun.

PR #3356, opened 2026-08-19
fix: resolve adapter parents by canonical identity

yaseralnajjar:
## Summary

- select inherited adapter plans by canonical class identity
- prevent a colliding local bare name from redirecting imported inheritance
- make duplicate canonical generic-parent candidates deterministic

## Validation

- create-PR gate ran once on a0ba94e5e2416de1cce716ca56758658218f39fc; all preceding checks passed, then the two separately owned Rust-interop inputs stopped the lane

### DIFF
--- a/crates/sifr_frontend/src/early_adapters.rs
+++ b/crates/sifr_frontend/src/early_adapters.rs
@@ -673,12 +673,12 @@ fn normalized_field_types(
             _ => None,
         })
         .or_else(|| parent_name.map(|name| format!("{module_name}.{name}")));
-    let parent = parent_name.and_then(|name| {
+    let parent = parent_name.and_then(|_| {
         inheritance::parent_selection(
+            module_name,
             result,
             external_defs,
             parent_identity.as_deref().unwrap_or(""),
-            name,
         )
     });
     let type_args = match class.parent_type.as_ref() {
--- a/crates/sifr_frontend/src/early_adapters/handler_plans.rs
+++ b/crates/sifr_frontend/src/early_adapters/handler_plans.rs
@@ -132,7 +132,7 @@ pub(super) fn inherited_handler_plans(
             _ => None,
         })
         .unwrap_or_else(|| format!("{module_name}.{parent_name}"));
-    inheritance::parent_selection(result, external_defs, &parent_identity, parent_name)
+    inheritance::parent_selection(module_name, result, external_defs, &parent_identity)
         .map(|parent| parent.handler_plans.clone())
         .unwrap_or_default()
 }
--- a/crates/sifr_frontend/src/early_adapters/inheritance.rs
+++ b/crates/sifr_frontend/src/early_adapters/inheritance.rs
@@ -56,7 +56,7 @@ pub(super) fn declaration_value(
     } else {
         return Err("adapted data parent shape is unavailable");
     };
-    let parent_selection = parent_selection(result, external_defs, &parent_identity, parent_name);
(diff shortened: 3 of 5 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate `a0ba94e5e2416de1cce716ca56758658218f39fc`: `scripts/run_all_tests.sh --profile create-pr` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

yaseralnajjar:
Exact candidate SHA: `a0ba94e5e2416de1cce716ca56758658218f39fc`

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr_frontend/src/early_adapters/inheritance.rs:219-227` — `parent_bindings` still resolves parent type params by bare `parent_name` against `result.module.classes`, falling back to `external_defs.class_type_params` only when the local lookup yields no params. In the exact collision this item fixes (local `Parent` shadowing imported `models.Parent`), plan selection is now canonical while the type-var bindings applied to those plans in `declaration_value` and `normalized_field_types` can still come from the colliding local class, so `substitute_type_vars` may bind the wrong parameter names. This was wrong before the change too (both sides were bare-name), so it is not a regression, but it is the remaining half of the same ambiguity and belongs in the follow-up item.
- pre-existing issue: `crates/sifr_frontend/src/early_adapters.rs:672` and `crates/sifr_frontend/src/early_adapters/handler_plans.rs:130` — `identity.as_deref().unwrap_or(name)` can yield a bare, module-less parent identity. Under the new canonical matcher such a value matches no local class (the matcher always compares against `{module}.{name}`) and `rsplit_once('.')` returns `None`, so `parent_selection` returns `None` where the old bare-name path returned the local selection. In the frontend path `parent_type` identities are populated (`declared_class_identity` with `current_module_name` always `Some`), so I could not construct a reachable case, but the `unwrap_or(name)` branch is now silently dead-or-fail-closed rather than meaningful; canonicalizing or removing it would make the rule explicit.
- suggestion: `crates/sifr_lowering/src/lower/generic_parent_representation.rs:79-84` — the new "prefer the requested key" branch has no test; `duplicate_canonical_candidates_use_a_deterministic_key` deliberately passes `"Missing"` and only exercises the `min_by_key` path. A second case asserting the requested key wins over a lexicographically smaller matching alias would lock in the stated preference order.
- suggestion: `crates/sifr_frontend/src/early_adapters.rs:676-683` — `parent_name.and_then(|_| ...)` combined with `parent_identity.as_deref().unwrap_or("")` reads as if an empty identity is possible; since `parent_identity` is `Some` whenever `parent_name` is `Some`, `parent_identity.as_deref().and_then(...)` would express the invariant directly.
- suggestion: `internal_docs/architecture.md:896-897` — the durable rule records canonical selection and deterministic key order but omits the "requested key wins when it is a matching canonical candidate" half of the acceptance criterion.

yaseralnajjar:
Merge gate evidence for exact reviewed candidate `a0ba94e5e2416de1cce716ca56758658218f39fc`: `scripts/run_all_tests.sh` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the same two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

PR #3358, opened 2026-08-19
fix: bind adapter parents by canonical identity

yaseralnajjar:
## Summary

- bind inherited adapter type parameters by canonical class identity
- canonicalize missing local parent identities and remove bare-name field and method fallbacks
- pin requested-key precedence for duplicate canonical generic-parent candidates

## Validation

- create-PR gate ran once on 76f76c81f1d9b8f9acb1c985f4fd6bb34a394dad; all preceding checks passed, then the two separately owned Rust-interop inputs stopped the lane

### DIFF
--- a/crates/sifr_frontend/src/early_adapters.rs
+++ b/crates/sifr_frontend/src/early_adapters.rs
@@ -667,33 +667,26 @@ fn normalized_field_types(
         .parent_type
         .as_ref()
         .and_then(|parent| match parent {
-            Type::Class { identity, name, .. } => {
-                Some(identity.as_deref().unwrap_or(name).to_string())
-            }
+            Type::Class { identity, name, .. } => Some(inheritance::canonical_parent_identity(
+                module_name,
+                identity.as_deref(),
+                name,
+            )),
             _ => None,
         })
         .or_else(|| parent_name.map(|name| format!("{module_name}.{name}")));
-    let parent = parent_name.and_then(|_| {
-        inheritance::parent_selection(
-            module_name,
-            result,
-            external_defs,
-            parent_identity.as_deref().unwrap_or(""),
-        )
+    let parent = parent_identity.as_deref().and_then(|identity| {
+        inheritance::parent_selection(module_name, result, external_defs, identity)
     });
     let type_args = match class.parent_type.as_ref() {
         Some(Type::Class { type_args, .. }) => type_args.as_slice(),
         _ => &[],
     };
-    let bindings = parent_name.map_or_else(HashMap::new, |name| {
-        inheritance::parent_bindings(
-            result,
-            external_defs,
-            parent_identity.as_deref().unwrap_or(""),
-            name,
-            type_args,
-        )
-    });
+    let bindings = parent_identity
+        .as_deref()
+        .map_or_else(HashMap::new, |identity| {
+            inheritance::parent_bindings(module_name, result, external_defs, identity, type_args)
+        });
     expected
         .iter()
         .map(|(identity, _)| {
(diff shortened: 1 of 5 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate `76f76c81f1d9b8f9acb1c985f4fd6bb34a394dad`: `scripts/run_all_tests.sh --profile create-pr` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

yaseralnajjar:
Exact candidate SHA: `76f76c81f1d9b8f9acb1c985f4fd6bb34a394dad`

Verdict: SATISFIED

Blocking findings:
- none

Reviewed against the exact diff and call sites:

- Binding identity authority: `parent_bindings` (`crates/sifr_frontend/src/early_adapters/inheritance.rs:216-223`), `local_parent_fields` (`:188-191`), and `parent_method_contracts` (`:267-270`) all now match local candidates by canonical identity instead of bare name, so a colliding local class can no longer supply the empty-field or empty-method fallback for an imported parent. External params still resolve via the `module.owner` split of the canonical identity (`:231-237`).
- Canonicalization of a missing identity to the current module is centralized in `canonical_parent_identity` (`:242-250`) and used at both remaining callers (`early_adapters.rs:670-674`, `early_adapters/handler_plans.rs:129-133`).
- The meaningless `unwrap_or("")` identity branches at `normalized_field_types` are gone; `parent` and `bindings` are now gated on a present canonical identity (`early_adapters.rs:678-689`). This widens computation to the case where `parent_type` is a class but `data_parent` is `None`; in that case `expected` carries no non-local field identities, so `parent`/`bindings` are unused — behavior-neutral, not a regression.
- Requested-key precedence is pinned by `requested_canonical_candidate_wins_over_key_order` (`generic_parent_representation.rs:246-257`), which exercises the requested-key branch at `:79-86` against the `min_by_key` deterministic-order branch at `:87-93`; the pre-existing determinism test still holds.
- Architecture wording (`internal_docs/architecture.md:896-898`) now states plan *and* binding identity authority plus requested-key-then-deterministic-order precedence.
- No compatibility shim or fallback path was introduced; nothing package-specific was added.

Follow-up findings:
- suggestion: `parent_selection` (`inheritance.rs:341-345`) and `fallback_parent_shape` (`:150-154`) still inline the same `identity.unwrap_or(module.name)` canonicalization by hand instead of calling the new `canonical_parent_identity` helper. Consolidating removes the risk of the three copies drifting.
- pre-existing issue: `fallback_parent_shape` (`:143-156`) still resolves the parent shape by bare local name and returns before consulting `external_defs`, so a colliding local class can shadow an imported parent on the path where `class.parent_type` is not a `Type::Class`. No identity is available at that point, so it is outside this item's criterion, but it is the last bare-name parent resolution in the module.
- suggestion: the new tests cover `parent_bindings` and the canonicalization helper directly; `local_parent_fields` and `parent_method_contracts` identity matching are only covered indirectly by the suite. A focused test for the empty-method fallback under a name collision would pin that half of the criterion explicitly.

yaseralnajjar:
Merge gate evidence for exact reviewed candidate `76f76c81f1d9b8f9acb1c985f4fd6bb34a394dad`: `scripts/run_all_tests.sh` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the same two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

PR #3360, opened 2026-08-19
fix: fail closed without canonical adapter parent

yaseralnajjar:
## Summary

- remove identityless parent-shape recovery from adapter declaration reconstruction
- reuse the canonical parent identity helper at local selection
- add direct empty imported-parent method collision coverage

## Validation

- create-PR gate ran once on c98a882108961e6fa83a7e9875fe1bb3c87eee34; all preceding checks passed, then the two separately owned Rust-interop inputs stopped the lane

### DIFF
--- a/crates/sifr_frontend/src/early_adapters/inheritance.rs
+++ b/crates/sifr_frontend/src/early_adapters/inheritance.rs
@@ -32,9 +32,9 @@ pub(super) fn declaration_value(
     else {
         return Ok(value);
     };
-    let Some(parent_name) = selection.data_parent.as_deref() else {
+    if selection.data_parent.is_none() {
         return Ok(value);
-    };
+    }
     let (parent_identity, mut parent_fields) = if let Some(Type::Class {
         identity,
         name,
@@ -50,10 +50,6 @@ pub(super) fn declaration_value(
             fields.clone()
         };
         (parent_identity, fields)
-    } else if let Some(shape) =
-        fallback_parent_shape(module_name, result, external_defs, parent_name)
-    {
-        shape
     } else {
         return Err("adapted data parent shape is unavailable");
     };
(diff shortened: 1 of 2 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate `c98a882108961e6fa83a7e9875fe1bb3c87eee34`: `scripts/run_all_tests.sh --profile create-pr` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

yaseralnajjar:
Exact candidate SHA: `c98a882108961e6fa83a7e9875fe1bb3c87eee34`

Verdict: SATISFIED

Blocking findings:
- none

Notes on what I verified: `fallback_parent_shape` is fully removed with no remaining callers (`crates/sifr_frontend/src/early_adapters/inheritance.rs:37-55` now returns the existing `"adapted data parent shape is unavailable"` error for any non-`Type::Class` parent, and both call sites — `early_adapters.rs:744` and `adapter_input.rs:37` — propagate it as a diagnostic rather than swallowing it). The `selection.data_parent` early return is preserved, so the no-data-parent path is unchanged, and the canonical local/imported paths still flow through `canonical_parent_identity` + `parent_selection`. `parent_selection` (line ~291) now uses the shared helper instead of the inlined `unwrap_or_else` formatting, which is behavior-identical. The new test pins the collision case correctly: local `Parent` canonicalizes to `consumer.Parent`, the child's parent identity is forced to `models.Parent` with cleared methods, so `parent_method_contracts` must return empty — the assertion would fail if the bare-name lookup were reintroduced. The architecture note states the fail-closed requirement in the durable contract section that already covers canonical-identity parent selection.

Follow-up findings:
- pre-existing issue: `handler_plans.rs:136` and `early_adapters.rs:677` still synthesize a parent identity as `{module_name}.{parent_name}` when `parent_type` is not a class. These are narrower than the removed helper (local-module-qualified only, so an imported parent cannot be redirected by a bare local name) and are now effectively unreachable for non-class parents because `declaration_value` fails closed first in the same pipeline. Still worth folding into the canonical-identity contract in a later item so the fail-closed rule is uniform rather than relying on ordering.
- suggestion: `declaration_value` matches `Some(Type::Class {..})` without `resolve_alias()`, whereas the deleted external branch did resolve aliases. If an adapted data parent can ever lower to an alias wrapping a class, it now fails closed rather than resolving. No test in the suite exercises that shape, so this is speculative — a targeted check would confirm whether it's a real gap or an impossible state.

yaseralnajjar:
Merge gate evidence for exact reviewed candidate `c98a882108961e6fa83a7e9875fe1bb3c87eee34`: `scripts/run_all_tests.sh` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the same two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

PR #3362, opened 2026-08-19
fix: require class identity for adapter parents

yaseralnajjar:
## Summary

- remove unreachable parent identity synthesis outside real class parent types
- fail closed in inherited handler lookup when no class identity exists
- pin that a type alias cannot supply class-parent identity

## Validation

- create-PR gate ran once on bc4a92c5eba3e9150a7cd9a4f506929d456f41f7; all preceding checks passed, then the two separately owned Rust-interop inputs stopped the lane

### DIFF
--- a/crates/sifr_frontend/src/early_adapters.rs
+++ b/crates/sifr_frontend/src/early_adapters.rs
@@ -662,19 +662,14 @@ fn normalized_field_types(
     expected: &[(String, String)],
 ) -> Result<Vec<(String, Type)>, String> {
     let local_prefix = format!("{module_name}.{}.", selection.owner);
-    let parent_name = selection.data_parent.as_deref();
-    let parent_identity = class
-        .parent_type
-        .as_ref()
-        .and_then(|parent| match parent {
-            Type::Class { identity, name, .. } => Some(inheritance::canonical_parent_identity(
-                module_name,
-                identity.as_deref(),
-                name,
-            )),
-            _ => None,
-        })
-        .or_else(|| parent_name.map(|name| format!("{module_name}.{name}")));
+    let parent_identity = class.parent_type.as_ref().and_then(|parent| match parent {
+        Type::Class { identity, name, .. } => Some(inheritance::canonical_parent_identity(
+            module_name,
+            identity.as_deref(),
+            name,
+        )),
+        _ => None,
+    });
     let parent = parent_identity.as_deref().and_then(|identity| {
         inheritance::parent_selection(module_name, result, external_defs, identity)
     });
--- a/crates/sifr_frontend/src/early_adapters/handler_plans.rs
+++ b/crates/sifr_frontend/src/early_adapters/handler_plans.rs
@@ -116,10 +116,10 @@ pub(super) fn inherited_handler_plans(
     external_defs: &ExternalDefs,
     selection: &ClassAdapterSelection,
 ) -> Vec<AdapterHandlerPlan> {
-    let Some(parent_name) = selection.data_parent.as_deref() else {
+    if selection.data_parent.is_none() {
         return Vec::new();
-    };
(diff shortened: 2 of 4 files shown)


yaseralnajjar:
Create-PR gate evidence for exact candidate `bc4a92c5eba3e9150a7cd9a4f506929d456f41f7`: `scripts/run_all_tests.sh --profile create-pr` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

yaseralnajjar:
Exact candidate SHA: `bc4a92c5eba3e9150a7cd9a4f506929d456f41f7`

Verdict: SATISFIED

Blocking findings:
- none

Notes on what I verified (not new requirements):
- `crates/sifr_frontend/src/early_adapters.rs:663-670` — the `.or_else(|| parent_name.map(...))` synthesized identity is gone; `parent_identity` is now `Some` only for a `Type::Class` parent, and the downstream `parent_selection` / `parent_bindings` calls already handle `None` (`and_then` / `map_or_else`), so canonical class parents behave exactly as before. Unreachable-with-data-parent-but-non-class because `expected_field_contracts` → `inheritance::declaration_value` errors first (`inheritance.rs:52`, "adapted data parent shape is unavailable").
- `crates/sifr_frontend/src/early_adapters/handler_plans.rs:119-136` — the `unwrap_or_else(|| format!("{module_name}.{parent_name}"))` synthesis is gone; both remaining exits (`data_parent` absent, or parent not a `Type::Class`) return no inherited plans. Both entry points (`adapter_input.rs:62` and `handler_plans.rs:21`) sit behind `run_one`'s `adapter_input` call at `early_adapters.rs:89`, which fails closed via `declaration_value`, so no silent-empty-inheritance path is reachable for a declared data parent. No fallback added.
- `grep` confirms no other synthesized parent identity remains in `early_adapters*`; `inheritance.rs:200` (`canonical_parent_identity`) still canonicalizes a missing identity from a real class name, which is the pre-existing intended behavior pinned by `missing_parent_identity_is_canonicalized_to_the_current_module`.
- `crates/sifr_lowering/src/lower/type_alias_tests.rs:35-43` pins the alias boundary at lowering with the exact message, `DiagnosticCode::CLASS_INVALID_BASE`, and the `ParentAlias` range, using `expect_err` so rejection precedes any adapter reconstruction. I re-ran just this test: 1 passed.
- `internal_docs/architecture.md:899-900` states class-base lowering rejects a type alias as data-parent identity authority, in the adapter-inheritance paragraph.

Follow-up findings:
- suggestion: the fail-closed guarantee for `inherited_handler_plans` is positional — it depends on `adapter_input`/`validate_fields` invoking `declaration_value` before or alongside the handler path. Returning `Vec::new()` there is correct today but would become a silent drop if a future caller reaches handler inheritance without the declaration reconstruction. A one-line comment at `handler_plans.rs:119` naming that dependency (or having it mirror the `Err` shape) would keep the invariant local.
- infrastructure issue: the create-PR gate stopped on two externally owned Rust-interop fixture inputs; per the evidence these are unrelated to this item and belong in their owning issue.

yaseralnajjar:
Merge gate evidence for exact reviewed candidate `bc4a92c5eba3e9150a7cd9a4f506929d456f41f7`: `scripts/run_all_tests.sh` ran once. Generated-demo freshness and every preceding guard passed. The lane then stopped at the same two separately owned Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

PR #3364, opened 2026-08-19
docs: classify remaining compiler deferrals

yaseralnajjar:
## Summary

- classify every recorded compiler deferral as complete, certification, terminal, or a bounded item
- order remaining compiler mechanism work as items 18 through 26
- keep pre_v1 and the separately owned Rust-interop gate inputs out of scope

## Validation

- git diff --check
- documentation-only item; compiler gates do not apply


yaseralnajjar:
Initial candidate SHA: `c7b0f7b8023d679d5c68ad3c8c838e82d329c88b`

Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md:2481` (the `M11 compiler prerequisites` audit row) — criterion "Audit every compiler-owned M12 deferral already recorded by the phase" / "Group remaining work into bounded mechanism items". Reason: the M11 compiler prerequisite 4 deferral at line 1792, "M12 must localize provider-module references in nonliteral attached defaults" (recorded as the separate mechanism defect found by that item's second review, which is why no third review ran), is not classified anywhere in the new audit. The M11 row's ten entries map exactly onto the prerequisite 2, 3, 5, 6, and 7 deferrals (independent negatives / direct attached codegen / stale installed-helper paths; duplicate decorator scans / concrete-owner assertions; hash predicates / flattened parent defaults / recursive boxed parents; structural-bound prescans; `StdlibFeature` inventory); prerequisite 4's surviving M12 clause has no counterpart there or in any other row, and no merged M12 item 1–16 record claims it — `grep -rn -i "nonliteral|non-literal" crates` shows no compiler work on that path either. A known mechanism defect therefore silently drops out of the phase's remaining-work set. Correction: add the nonliteral attached-default provider-module localization to the M11 row as owned by item 26 (or as its own bounded item ordered with the other attached-API/codegen work), classified as a bounded remaining item rather than terminal.

Follow-up findings:
- pre-existing issue: the M11 package deferral at line 1976 embeds one compiler-owned clause, "resolve the compiler's nested direct specialization result". Item 1's record (line 2015) reports direct local and imported nested concrete generic-plan regressions, so it is plausibly already closed, but the audit never states that. A one-clause "complete via item 1" note in the M8-M10 or M12 row would remove the ambiguity.
- suggestion: three smaller prerequisite clauses are absorbed only implicitly — prerequisite 2's conditional "align negative metadata if core-language gains a manifest-driven harness" (line 1728), prerequisite 3's "split `rust_interop_structural_tests.rs` before that file reaches 900 lines" (line 1758), and prerequisite 5's "add direct plain-parent structural coverage" (line 1835). The M6/M11 rows treat file headroom and coverage generically; naming these as certification or terminal would make the row exhaustive.
- suggestion: item 2's residual notes (lines 2076-2079, impossible local handler owner; re-canonicalizing imported boundary types against colliding consumer-local names) are summarized by the M12 row as "assertions, comments, performance ideas, or certification work". Items 12 and 13-15 do appear to have closed them, so the classification holds, but "closed by items 12 and 13-15" is more accurate than the current wording.

Non-blocking confirmations: item 18 is the sole selected next action; the `pre_v1` and separately owned Rust-interop exclusions are preserved; the ordering places the M5 handler-descriptor item first; no terminal classification covers a stated correctness defect.

yaseralnajjar:
Final candidate SHA: `239f6b6bf03205f803ca2377c21a15937122c7e5`

Inspected the final diff (base → final) and cross-checked each audit row against the recorded M0–M12 deferrals in the phase record.

Verification notes:
- `plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md:2481` now names "nonliteral attached-default provider localization" in item 26, matching the M11 prerequisite‑4 defect at line 1786 ("M12 must localize provider-module references in nonliteral attached defaults"). Prior blocking finding corrected, and it is kept as a bounded item, not reclassified.
- The same row now also names "direct plain-parent structural coverage" (prereq 5, line 1834), states the conditional clause ("Negative metadata alignment is conditional on a future manifest-driven core harness", prereq 2 line 1726) and the guard-owned split ("The 900-line guard owns the structural-test split", prereq 3 line 1758).
- `:2482` names item 1 (nested local/imported concrete generic-plan specialization, line 2006) and items 12 and 13–15 (fail-closed structural handler ancestry, line 2331; canonical parent selection and collision coverage, lines 2359–2440), which map to item 2's deferrals at lines 2082–2085.
- Remaining M11 prereq 1–7 and M12 item 1–16 deferrals are each accounted for as complete, certification, terminal, or bounded; `pre_v1` and separately owned Rust-interop exclusions are preserved at `:2487`. Only the phase record changed; nothing else regressed.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: prerequisite 7's second deferral ("parse the inline dependency tables in the exact-pin test", line 1907) is only implicitly folded into item 26's "retained `StdlibFeature` inventory"; naming it explicitly when item 26 is implemented would keep the trace exact.
- suggestion: the audit selects item 18 as next but the record never states what item 17 is (presumably this audit itself); one sentence identifying item 17 would remove the numbering gap.

PR #3366, opened 2026-08-19
Close M5 handler descriptor contract gaps

yaseralnajjar:
## Summary
- reject declared-owned field and temporary receivers without an owned local binding
- restrict `Self` annotations to ordinary class methods and pin static-method descriptor stacking
- make inherited-handler shape validation local to handler planning
- split decorator-order and Python-cleanup validation along responsibility boundaries

## Validation
- affected lowering, frontend, and driver suites
- formatting, maintainability, diff hygiene, and 900-line guard
- create-PR gate attempted once on exact candidate `5c44c248d9a0144fb0c95151349b06061c45167b`; all preceding guards passed, then the out-of-scope Rust-interop matrix stopped on its known missing negative fixture and empty placeholder class

### DIFF
--- a/crates/sifr_driver/src/tests/handler_descriptors.rs
+++ b/crates/sifr_driver/src/tests/handler_descriptors.rs
@@ -203,6 +203,29 @@ class Invalid(HandlerContract):
 }
 
 #[test]
+fn staticmethod_descriptor_must_be_the_inner_adjacent_decorator() {
+    let errors = compile_errors(
+        r#"
+from fixture.handlers import HandlerContract, handler
+
+class Invalid(HandlerContract):
+    value: str
+
+    @handler("wrong-order")
+    @staticmethod
+    def parse(own value: str) -> str:
+        return value
+"#,
+    );
+    assert!(errors.iter().any(|error| {
+        error.code == sifr_diagnostics::DiagnosticCode::META_MALFORMED_DECLARATION.code()
+            && error
+                .message
+                .contains("directly above the method with @staticmethod as the outer decorator")
+    }));
+}
+
+#[test]
 fn owned_handler_requires_exact_self_output_at_the_descriptor() {
     let errors = compile_errors(
         r#"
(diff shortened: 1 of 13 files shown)


yaseralnajjar:
Exact candidate SHA: `5c44c248d9a0144fb0c95151349b06061c45167b`

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_lowering/src/lower/typing_and_functions/annotation_resolution.rs:17-19` — the `unreachable!` turns a paired-state invariant (`self_annotation_available` ⇒ `current_class.is_some()`) into a compiler panic. The invariant holds today (all four assignment sites in `class_body_lowering.rs:104/250/442` and `class_type_collection.rs:522` set both together, and no `continue`/`return` escapes between set and restore), but falling through to `invalid_type_annotation` would be a strictly safer failure mode than a panic if a future edit breaks the pairing.
- suggestion: `crates/sifr_lowering/src/lower/classes/class_body_lowering.rs:105` and `:251` — the enum and newtype paths set `self_annotation_available = true` unconditionally, so `Self` in a `@staticmethod`/`@classmethod` annotation on those class shapes is rejected only because `collect_class_type` (`class_type_collection.rs:523-525`) already errored. The outcome is correct (compilation fails), but the rule is not local to those paths the way it is at `:443`.
- suggestion: `crates/sifr_frontend/src/typed_descriptors/method_decorator_ordering.rs:25-250` — when a method carries both `@classmethod` and `@staticmethod` plus a descriptor, both diagnostics fire for the same malformed stack. One message per stack would read better; the established `@classmethod` behavior is otherwise preserved verbatim.
- suggestion: `crates/sifr_frontend/src/typed_descriptors/method_decorator_ordering.rs:37-41` — the staticmethod rule additionally requires exactly two decorators (`index + 1 != decorator_list.len()`), so `@staticmethod` + descriptor + any third decorator is rejected. That mirrors the `@classmethod` rule and matches the criterion, but only the swapped-order case is tested; the "descriptor not last" case is untested.
- infrastructure issue: the evidence covers library suites plus the create-PR gate up to the two separately owned Rust-interop matrix inputs, so the E2E pass suite did not run against the new field/temporary owned-receiver rejection (`method_calls.rs:352-358`), which is a behavioral tightening. I scanned every `.sifr` source in the repo for `own self` methods invoked on a non-identifier receiver (field, subscript, or chained call) and found none, and the existing `@staticmethod` + descriptor fixtures (`verification/areas/core_language/fixtures/typed_declaration_descriptors/src/app.sifr:18`, `verification/areas/rust_interop/fixtures/method_slot_table/examples/method_slot_runtime/src/main.sifr:20/58`) all match the newly pinned valid stack — but the suite itself was not exercised.
- pre-existing issue: consuming an owned-receiver method through a borrowed `self` (`self.finish()` inside `def m(self)`) still takes the `HirExpr::Name` branch and is handled by receiver-convention propagation in `method_receiver_analysis.rs` rather than by a diagnostic. Unchanged by this diff and outside the field/temporary target of the criterion.

yaseralnajjar:
Exact candidate SHA: `5c44c248d9a0144fb0c95151349b06061c45167b`

Gate evidence:
- Create-PR gate ran once. It passed every preceding guard, including generated-demo freshness and the 3,191-file size guard. It stopped at the unchanged, out-of-scope Rust-interop matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr` placeholder class.
- Merge gate ran once on the same SHA. It passed every preceding guard and stopped at the same two inputs.
- Neither gate was rerun. The `pre_v1` session, worktree, processes, artifacts, failures, and fixes remain wholly out of scope.

PR #3368, opened 2026-08-20
fix structural mapping contracts

yaseralnajjar:
## Summary
- keep same-path structural mapping probes enforceable and align malformed local marking with exported metadata
- track stdlib structural implementation origin explicitly and cover late multi-module emission
- document and test mapped-value thread-safety backstops while restoring source-file headroom

## Validation
- cargo test -q -p sifr_runtime --features structural
- cargo test -q -p sifr_lowering -p sifr_codegen -p sifr_driver -- --skip test_e2e_pass
- cargo fmt --check
- git diff --check

Exact candidate: c631ca82d97773e223b4176001de5df0f07384ed

### DIFF
--- a/crates/sifr_codegen/src/lib.rs
+++ b/crates/sifr_codegen/src/lib.rs
@@ -160,7 +160,9 @@ pub use rust_interop_plan::{
 };
 mod rust_ir;
 pub use rust_ir::*;
+mod stdlib_codegen_metadata;
 mod stdlib_filter;
+pub use stdlib_codegen_metadata::StdlibCode;
 mod stdlib_import_signatures;
 mod stdlib_rust_source;
 pub use stdlib_rust_source::StdlibRustSource;
--- a/crates/sifr_codegen/src/lib_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests.rs
@@ -83,6 +83,8 @@ mod string_concat_codegen_tests;
 #[cfg(test)]
 mod structural_impl_demand_codegen_tests;
 #[cfg(test)]
+mod structural_stdlib_impl_codegen_tests;
+#[cfg(test)]
 mod structured_condition_codegen_tests;
 #[cfg(test)]
 mod structured_intrinsic_codegen_tests;
--- a/crates/sifr_codegen/src/lib_codegen_tests/structural_stdlib_impl_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/structural_stdlib_impl_codegen_tests.rs
@@ -1,0 +1,148 @@
+use sifr_ir::{
+    HirClass, HirClassKind, HirFunction, HirImport, HirModule, MethodKind,
+    RustInteropAbiRequirements, RustInteropDeclaration, RustInteropDecoratorKind,
+    RustInteropEffect,
+};
+use sifr_type_system::Type;
+use std::collections::{HashMap, HashSet};
+
+const STRUCTURAL_IMPL_TARGET: &str = "impl ::sifr_runtime::interop::structural::StructuralType";
+
+#[test]
+fn multi_module_stdlib_structural_impl_is_emitted_for_each_imported_nominal() {
+    let mut alpha = module(Vec::new(), Vec::new());
+    alpha.imports.push(json_import());
+    let mut zeta = module(Vec::new(), Vec::new());
+    zeta.imports.push(json_import());
+    let main = module(vec![structural_function()], Vec::new());
+
+    let generated = crate::generate_rust_multi_with_metadata(
+        &[("zeta", &zeta), ("main", &main), ("alpha", &alpha)],
+        &json_stdlib(),
+    );
+    let count = generated
+        .rust_files
+        .values()
+        .map(|source| source.matches(STRUCTURAL_IMPL_TARGET).count())
+        .sum::<usize>();
+
+    assert_eq!(count, 2, "{:?}", generated.rust_files);
+    assert!(generated.rust_files["alpha"].contains(STRUCTURAL_IMPL_TARGET));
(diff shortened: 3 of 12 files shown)


yaseralnajjar:
## Review

Verified the diff against each criterion; one criterion fails.

**Criterion 5 detail.** `MappedValue<T, M>` stores `_mapping: PhantomData<fn() -> M>` (`crates/sifr_runtime/src/interop/structural/mapped.rs:45`). Function-pointer types implement `Send`/`Sync` unconditionally, so `PhantomData<fn() -> M>` is `Send + Sync` for every `M`, including a `!Send` marker. I confirmed this with a standalone `rustc` check: `MappedValue<String, M>` with `struct M(*mut u8)` satisfies `Send + Sync`. So the actual behavior is "`Send`/`Sync` iff `T` is", not "iff both `T` and the mapping marker are". The added test (`String` + `StringMapping`, both `Send + Sync`) passes under either reading, so it does not check the claim the doc makes.

Other criteria: (1) the `mapping != rust_path` filter is gone and the same-path obligation is emitted, with a direct unit test; (2) `has_valid_rust_opaque_structural_mapping_syntax` requires `Expr::Attribute`, which matches the ≥2-segment requirement in `parse_target_path`, so a bare-`Name` mapping no longer marks the class Structural — consistent with the HIR-based export marking in `rust_class_exports.rs:19`; (3) the `identity.starts_with("sifr.")` heuristic is replaced by `StructuralRecordOrigin`, with a negative test for a project class carrying a `sifr.`-prefixed identity; (4) `emitted_targets` is per-module and the new multi-module test asserts one impl per generated module (each module inlines its own stdlib preamble, so the duplicate impls are in distinct module scopes); (6) `lib_modules_and_codegen.rs` was at 899 lines on base, so the `StdlibCode` extraction was size-required and responsibility-aligned.

Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_runtime/src/interop/structural/mapped.rs:17-19`, criterion 5. The documented `Send`/`Sync` behavior is incorrect: because the marker is held as `PhantomData<fn() -> M>`, auto traits ignore `M` entirely, so `MappedValue<T, M>` is `Send`/`Sync` exactly when `T` is — the "only when both the stored value and mapping marker do" claim is false, and the new positive-only test at `mapped.rs:228-233` cannot detect the difference. Correction: state that the marker is erased behind `PhantomData<fn() -> M>` so `Send`/`Sync` follow `T` alone, and extend the compile-time check with a deliberately `!Send`/`!Sync` mapping marker (e.g. one holding a raw pointer) asserting `MappedValue<String, ThatMarker>: Send + Sync`, which pins the real behavior. If the intent was instead for the marker to constrain thread safety, change the field to `PhantomData<M>` and keep the current wording.

Follow-up findings:
- suggestion: the Send/Sync paragraph is appended to the `StructuralMapping` trait doc comment, so rustdoc renders it on the trait rather than on `MappedValue`; it belongs on the `MappedValue` struct doc at `mapped.rs:39-43`.
- suggestion: `has_valid_rust_opaque_structural_mapping_syntax` uses an `Expr::Attribute` shape proxy rather than the real validity rule. Attribute forms that still fail `parse_target_path` (`structural=rust.Foo`, `structural=make().Foo`) mark the class Structural locally while the HIR declaration is dropped, diverging from export marking. Harmless today because those forms already emit a malformed-decorator error, but reusing the segment collection/validation would remove the divergence.
- suggestion: `structural_impl_codegen/stdlib_implementations.rs` holds a single 8-line function while the parent file sits at 858/900 lines, so this extraction was not size-forced; inlining it back would keep the target-string logic next to its only caller.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_lowering/src/lower/rust_interop.rs:36` — local structural marking now validates the `structural=` path shape, but it still marks the class `Structural` when a *different* argument of the same `@rust.opaque(...)` call is malformed (e.g. `close="x"`), which drops the whole declaration from exported HIR. No unsoundness (the malformed diagnostic already fails compilation), but the marking/HIR divergence class from criterion 2 is not fully closed; worth a later item that gates marking on successful declaration parse.
- suggestion: `crates/sifr_driver/src/build/rust_interop_probe_tests.rs:36` — the same-path obligation is pinned at the probe-source-text level only; a later item could add end-to-end coverage that a same-path mapping without a `StructuralMapping` impl surfaces through the structured probe diagnostic rather than raw rustc output.
- pre-existing issue: `crates/sifr_codegen/src/structural_impl_codegen.rs` (858 lines) and `crates/sifr_codegen/src/lib_modules_and_codegen.rs` (866 lines) sit close to the 900-line guardrail after this item's extractions; both will need a responsibility split before further growth.

yaseralnajjar:
Gate evidence for exact candidate `5fcaabe5798ac0e7a867194a1194da6f1eda731b`

- Create-PR gate: ran exactly once with `scripts/run_all_tests.sh --profile create-pr`. It passed every preceding guard and stopped at the Rust-interop matrix after 191.92s.
- Merge gate: ran exactly once with `scripts/run_all_tests.sh`. It passed every preceding guard and stopped at the same Rust-interop matrix after 127.77s.
- Both stops report the same separately owned inputs: missing `negative/package_generated_type_import_rejected.sifr` evidence and the empty placeholder body in `examples/method_slot_runtime/src/schema_program.sifr`.
- Neither gate was rerun. These inputs and all `pre_v1` work remain out of scope and untouched.

In-scope exact-candidate evidence is green: runtime 80/80; codegen 1059/1059; driver 543 passed with 76 ignored; lowering 1011 passed with 1 ignored; workspace Clippy; formatting; HIR maintainability; 900-line file-size guard; and diff hygiene.

PR #3371, opened 2026-08-20
Close M7/M7b attached API audit gaps

### LINKED ISSUE #3370: Audit M7/M7b attached API identity and generic-call metadata
Close item 20 of the active static class-adapter phase.

Scope:
- require exact canonical identity for imported attached-API sets
- reject attached API collisions with fields and methods
- make canonical and module-less owner binding keys explicit
- filter private attached declarations symmetrically
- validate the real data parent before StaticProgram eligibility
- canonicalize remaining generic-call metadata lookups
- document intentionally weaker emitted Rust generic bounds

Phase record: plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md

yaseralnajjar:
Closes #3370.

Closes item 20 of the active static class-adapter phase.

- require exact canonical identity for imported attached-API sets
- reject attached API collisions with fields and methods
- make canonical/module-less owner binding keys explicit
- filter private declarations symmetrically
- validate concrete real data-parent eligibility
- canonicalize generic-call metadata lookups
- document emitted Rust generic-bound authority

Validation on `78d510dacfcf0abae4b792de1859f5c32b905eb4`:
- `cargo test -p sifr_runtime`: 80 passed
- `cargo test -p sifr_codegen --lib`: 1,062 passed
- `cargo test -p sifr_frontend`: 117 passed
- `cargo test -p sifr_lowering`: 1,014 passed, 1 ignored
- `cargo test -p sifr_driver`: 546 passed, 76 ignored
- workspace clippy, fmt, HIR maintainability, file-size, and diff checks passed

### DIFF
--- a/crates/sifr_codegen/src/function_emitter/sifr_int_analysis.rs
+++ b/crates/sifr_codegen/src/function_emitter/sifr_int_analysis.rs
@@ -244,7 +244,8 @@ pub(super) fn hir_expr_returns_sifr_int_result(
         HirExpr::BinOp { op, ty, .. } => {
             matches!(op.as_str(), "//" | "%") && is_result_int_type(ty)
         }
-        HirExpr::Call { func, .. } => result_function_returns.contains(func),
+        HirExpr::Call { func, .. } | HirExpr::GenericCall { func, .. } => result_function_returns
+            .contains(crate::stmt_support_emitter::canonical_plain_call_name_for_ir(func)),
         HirExpr::MethodCall { object, method, .. } => {
             hir_expr_class_name(object).is_some_and(|class_name| {
                 result_method_returns.contains(&result_method_key(&class_name, method))
@@ -444,10 +445,12 @@ pub(super) fn collect_sifr_int_call_arg_function_params(
     let mut discovered: HashMap<String, HashSet<usize>> = HashMap::new();
     let mut on_stmt = |_stmt: &HirStmt| {};
     let mut on_expr = |expr: &HirExpr| {
-        let HirExpr::Call { func, args, .. } = expr else {
+        let (HirExpr::Call { func, args, .. } | HirExpr::GenericCall { func, args, .. }) = expr
+        else {
             return;
         };
-        let Some(params) = module_function_params.get(func) else {
+        let canonical = crate::stmt_support_emitter::canonical_plain_call_name_for_ir(func);
+        let Some(params) = module_function_params.get(canonical) else {
             return;
         };
         for (idx, arg) in args.iter().enumerate() {
@@ -464,7 +467,10 @@ pub(super) fn collect_sifr_int_call_arg_function_params(
                 module_sifr_int_bindings,
                 function_sifr_int_returns,
             ) {
-                discovered.entry(func.clone()).or_default().insert(idx);
+                discovered
+                    .entry(canonical.to_string())
+                    .or_default()
+                    .insert(idx);
             }
         }
     };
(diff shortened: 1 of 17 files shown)


yaseralnajjar:
**Verdict: SATISFIED**

Blocking findings:
- None.

Follow-up findings:
- pre-existing issue: `crates/sifr_codegen/src/method_call_emitter.rs:53` (`is_generator_call`), `crates/sifr_codegen/src/python_interop_plan.rs:273` (`raw_call_names.contains(func)`), and `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_method_and_question_mark.rs:346` (`imported_project_functions.contains(func)`) still match `HirExpr::Call` only and key on the raw `func`, without `canonical_plain_call_name_for_ir`. These are generator detection, Python-interop raw-call detection, and imported-project error conversion — outside the attached-API metadata surfaces this item canonicalized, but they are the same class of lookup and would silently miss a turbofished or `GenericCall` form.
- pre-existing issue: `crates/sifr_lowering/src/lower/descriptor_declarations.rs:248-253` registers `attached_api_set_bindings` for imported names via `attached_api_sets.get(module).contains_key(original)`, i.e. map-key presence, not `ExternalDefs::contains_attached_api_set`. It cannot diverge today because `descriptor_exports.rs:115-119,188-193` keys that map by the declaring module's own local sets with no re-export aliasing, so the stored identity always matches the key. Worth routing through the new accessor for uniformity.
- suggestion: `crates/sifr_lowering/src/lower/classes/class_type_collection.rs:320` inserts `parent_ty` into `class_data_parent_types` before the `if let Type::Class { .. } = parent_ty` check, so a non-class data parent (already diagnosed as `parent type '...' is not a class`) is recorded and would be evaluated by `supports_structural_bridge_type_inner` as a plain type. Compilation already fails on that path, so this is cosmetic ordering only.
- suggestion: `binding_for_owner` (`attached_api_surfaces.rs:249`) probes `surface_name`, canonical identity, then emitted name unconditionally, while insertion in `apply` gates the canonical key on the tail match. The lookup is a superset of the insert set, which is safe given keys are per-module, but a comment noting the asymmetry would keep the invariant legible.

Notes on what I verified beyond the diff: `contains_attached_api_set` is used at both former map-key sites (`early_adapters.rs:446`, `attached_api_declarations.rs:156`); the private-function filter is now applied symmetrically to `ctx.attached_apis` and `ctx.externals.attached_apis` in both `provisional_declarations` and `declarations_for_set` (the external side was already filtered at export time by `should_export_stdlib_declaration`, so the local side was the real asymmetry); the `type_bounds.rs:283-291` rewrite is strictly a tightening — `adapted_class_bindings.contains_key(name)` is still required, and `parent_class` is `Some` only when a real data-parent `Type::Class` was collected, so the `imported_structural_identity_inputs` fallback is reachable only for imported owners; and `func_signatures`/`nested_fn_captures`/`module_function_params`/`result_function_returns`/`function_sifr_int_returns` have no remaining non-canonical `func` lookups in `sifr_codegen`.

yaseralnajjar:
Exact-SHA validation evidence for `78d510dacfcf0abae4b792de1859f5c32b905eb4`:

- Candidate suites passed: runtime 80; codegen 1,062; frontend 117; lowering 1,014 passed / 1 ignored; driver 546 passed / 76 ignored.
- Workspace clippy, formatting, HIR maintainability, unified file-size, and diff checks passed.
- Opus review: SATISFIED with no blockers: https://github.com/sifr-lang/sifr/pull/3371#issuecomment-5355489839
- The single create-PR gate and single merge gate each passed all preceding guardrails and stopped only at the same externally owned `pre_v1` matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and the empty placeholder in `examples/method_slot_runtime/src/schema_program.sifr`. Per phase scope, these inputs were not inspected, changed, or rerun.
- No item-20-specific gate failure occurred.

PR #3395, opened 2026-08-20
Canonicalize generic-call and attached-set lookups

### LINKED ISSUE #3394: Canonicalize remaining generic-call and attached-set lookups
Phase item 28 from plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md.

Scope:
- Canonicalize generator-function lookup for generic/turbofished calls.
- Canonicalize the raw-coroutine-loop plan lookup for generic/turbofished calls.
- Canonicalize imported-project question-mark conversion lookup for generic/turbofished calls.
- Register imported attached-API set bindings only when the stored declaration has the exact canonical module and symbol identity.

Acceptance:
- Plain calls retain current behavior and GenericCall uses the same canonical function identity in all three codegen consumers.
- A map entry whose stored attached-set identity disagrees with its module key is not registered as an erased attached-set binding.
- Focused codegen and lowering regressions pass.

Out of scope: the item-20 non-class-parent ordering and safe-lookup asymmetry suggestions; all work and failures owned by the separate pre_v1 session 01a01407-6090-7a53-8ebc-c1c8f8cf0eec.

yaseralnajjar:
Closes #3394.

## Summary
- canonicalize generator, raw-coroutine, and imported-project call lookups across plain and generic HIR calls
- require exact stored identity before registering an imported attached-API set binding
- add focused regressions for every corrected lookup

## Validation
- `python3 scripts/check_sifr_driver_maintainability_guardrails.py`
- `git diff --check`

### DIFF
--- a/crates/sifr_codegen/src/method_call_emitter.rs
+++ b/crates/sifr_codegen/src/method_call_emitter.rs
@@ -50,10 +50,32 @@ impl RustEmitter {
 
     /// Check if an expression is a call to a generator function
     pub(crate) fn is_generator_call(&self, expr: &HirExpr) -> bool {
-        if let HirExpr::Call { func, .. } = expr {
-            self.generator_functions.contains(func)
-        } else {
-            false
-        }
+        matches!(
+            expr,
+            HirExpr::Call { func, .. } | HirExpr::GenericCall { func, .. }
+                if self.generator_functions.contains(
+                    crate::stmt_support_emitter::canonical_plain_call_name_for_ir(func)
+                )
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn generic_generator_calls_use_the_canonical_function_name() {
+        let mut emitter = RustEmitter::new();
+        emitter.generator_functions.insert("produce".to_string());
+        let expression = HirExpr::GenericCall {
+            func: "produce::<i64>".to_string(),
+            type_args: vec![Type::Int],
+            args: Vec::new(),
+            mutable_arg_places: Vec::new(),
+            ty: Type::Iterator(Box::new(Type::Int)),
+        };
+
+        assert!(emitter.is_generator_call(&expression));
     }
 }
--- a/crates/sifr_codegen/src/python_interop_plan.rs
+++ b/crates/sifr_codegen/src/python_interop_plan.rs
@@ -270,8 +270,13 @@ fn module_requires_raw_coroutine_loop(module: &HirModule) -> bool {
     }
     let mut required = false;
     let mut inspect = |expression: &sifr_ir::HirExpr| {
-        if matches!(expression, sifr_ir::HirExpr::Call { func, .. } if raw_call_names.contains(func))
-        {
+        if matches!(
+            expression,
+            sifr_ir::HirExpr::Call { func, .. } | sifr_ir::HirExpr::GenericCall { func, .. }
+                if raw_call_names.contains(
+                    crate::stmt_support_emitter::canonical_plain_call_name_for_ir(func)
+                )
+        ) {
             required = true;
         }
     };
--- a/crates/sifr_codegen/src/python_interop_plan_tests.rs
+++ b/crates/sifr_codegen/src/python_interop_plan_tests.rs
@@ -164,6 +164,32 @@ fn aliased_raw_coroutine_call_requires_the_owned_async_loop() {
 }
 
 #[test]
+fn generic_raw_coroutine_call_requires_the_owned_async_loop() {
+    let mut module = module_with_functions(vec![function(
+        "main",
+        vec![HirStmt::Expr {
+            expr: HirExpr::GenericCall {
+                mutable_arg_places: Vec::new(),
+                func: "run_coroutine_blocking::<Object>".to_string(),
+                type_args: vec![Type::Any],
+                args: Vec::new(),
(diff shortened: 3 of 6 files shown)


yaseralnajjar:
## Verdict: SATISFIED

No blocking in-scope mechanism defect. Findings per criterion, verified against the committed diff (`2997d549b`) and surrounding authorities.

**1. Generator-function lookup** — `method_call_emitter.rs:52-59` now matches `HirExpr::Call | HirExpr::GenericCall` with the guard applied to both arms and resolves through `canonical_plain_call_name_for_ir` (`expr_call_metadata.rs:19`, strips the `::<…>` suffix). `is_generator_call` has exactly one production consumer (`iterator_lowering.rs:325`), where recognition only short-circuits an already-iterator source — no downstream path assumes the `Call` shape. `generator_functions` is populated with plain names (`union_type_helpers.rs:115`, `generator_bodies.rs:307`, `lib_modules_and_codegen.rs:303`), so canonicalization is the correct key normalization, not a widening.

**2. Raw-coroutine-loop planning** — `python_interop_plan.rs:270-281` adds the `GenericCall` arm under the same canonical guard. `raw_call_names` is still built from `sifr.python` imports with the pre-existing original→alias resolution, so aliases retained by imports flow through unchanged; `BTreeSet<String>::contains(&str)` is valid via `Borrow`. Both class-method and nested-function traversals share the single `inspect` closure, so the corrected predicate covers every walked site.

**3. Imported-project question-mark conversion** — the inline `matches!` is replaced by `is_imported_project_call_for_ir` (`stmt_expr_method_and_question_mark.rs:8-19`) covering `Call | GenericCall | PythonCall` through the canonical name. `imported_project_functions` is populated with alias-local plain names (`module_prescan.rs:55-67`), so the key space matches. Grep confirms this is the only imported-project question-mark decision site — no second uncorrected path, no fallback branch retained.

**4. Attached-API set registration identity** — `descriptor_declarations.rs:247-256` now calls the pre-existing `ExternalDefs::contains_attached_api_set` (`external_defs.rs:197-202`), which requires `declaration.identity == requested identity`, not just map-key presence. This is the same authority already used by `attached_api_declarations::finalize` and `early_adapters::attached_api_set_exists`, so the change removes a divergent lookup rather than adding one. I confirmed `attached_api_sets` has a single writer (`descriptor_exports.rs:190` via `replace_module_entry`, keyed by the declaring `module_name` with identities stamped from `ctx.current_module_name` at `attached_api_declarations.rs:59-62`) and that sets — unlike markers/selections — are not re-keyed through the import re-export loop. There is therefore no legitimate flow where the key and `identity.module` diverge, so the strictening cannot reject a real import. `original` is moved at its last use in the loop body; no later use.

**5. Focused regressions** — one test per corrected mechanism: `generic_generator_calls_use_the_canonical_function_name`, `generic_raw_coroutine_call_requires_the_owned_async_loop`, `generic_imported_project_calls_use_the_canonical_function_name`, and `attached_api_set_import_requires_the_stored_canonical_identity`. The last one discriminates the fix: with a stored identity of `actual.Api` under key `declared`, the base code would register the binding and reject `class Child(Api)` via the `attached_api_set_use` malformed diagnostic (`descriptor_declarations.rs:317-324`), while the candidate leaves the imported class intact. No new authority, alias table, or fallback branch is introduced; the remaining hunks are pure `rustfmt` reflow.

### Non-blocking suggestions
- The three codegen regressions assert on helpers/planning predicates rather than emitted output; a `run`/`emit`-level fixture for a generic imported-project `?` conversion would pin the actual conversion emission, not just the predicate.
- Criterion 2's alias dimension is covered only by the pre-existing plain-`Call` alias test; a combined generic-plus-alias case (`run_coroutine_blocking as rcb` called as `rcb::<Object>`) would close the cross-product.
- Three near-identical `Call | GenericCall [| PythonCall]`-plus-canonical-name matches now exist across two crates' modules, and `canonical_plain_call_name_for_ir` is reached by fully-qualified path from 8 files. A single `canonical_call_name_of(&HirExpr) -> Option<&str>` accessor would make future lookup sites correct by construction.

### Pre-existing / out-of-scope observations
- `descriptor_exports::erase_marker_imports` (`crates/sifr_frontend/src/descriptor_exports.rs:9-20`) still decides attached-API-set import erasure by `items.contains_key(name)`, i.e. map-key presence alone — the asymmetry criterion 4 removes from the lowering registration site remains at this frontend erasure site. Not part of item 28's stated criterion (which names registration), so recording rather than blocking.
- Item-20 non-class-parent ordering and safe-lookup asymmetry, and everything owned by pre_v1 session `01a01407-…`, were excluded as directed. No files were modified during this review.

yaseralnajjar:
Exact candidate: `2997d549b2f2a13c90edcc185cb7de625dc0fa9d`

Review: Opus returned `SATISFIED` with no blocking in-scope finding: https://github.com/sifr-lang/sifr/pull/3395#issuecomment-5358588538

Validation:
- `cargo test -p sifr_codegen`: 1,078 passed.
- `cargo test -p sifr_lowering`: 1,020 passed, 1 ignored.
- Workspace Clippy, formatting, HIR and driver maintainability, the 900-line file-size guard, and diff hygiene passed.

External inputs, not inspected or changed here:
- missing `negative/package_generated_type_import_rejected.sifr`
- empty `examples/method_slot_runtime/src/schema_program.sifr`

All `pre_v1` session work remains out of scope.

PR #3398, opened 2026-08-20
Reject unsafe optional callback inputs

### LINKED ISSUE #3397: Reject optional iterable elements for required map and filter callbacks
Phase item 29 from plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md.

Scope:
- Determine whether an optional iterable element can reach a non-optional registry callback parameter.
- Reject incompatible map() and filter() callback input types during lowering.
- Make registry callback codegen fail closed instead of generating a data-dependent option expect for an unadapted optional value.
- Preserve valid owned class upcasts in the shared callback argument adapter.

Acceptance:
- map() and filter() report SIFR-TYPE-0002 when an iterable element type is not assignable to the matching callback parameter.
- Valid callback inputs retain current behavior, including subclass-to-parent conversion.
- The registry callback helper does not emit a data-dependent option unwrap for incompatible optional input.
- Focused lowering and codegen regressions pass.

Out of scope: all work and failures owned by the separate pre_v1 session 01a01407-6090-7a53-8ebc-c1c8f8cf0eec.

yaseralnajjar:
Closes #3397.

## Summary
- type-check `map()` and `filter()` iterable element types against their callback parameters
- reject malformed optional-to-required registry callback HIR instead of generating a data-dependent option unwrap
- preserve and test owned subclass-to-parent callback conversion
- add exact diagnostics for optional iterable elements passed to required callback parameters

## Validation
- both maintainability guards
- 900-line file-size guard
- `git diff --check`
- direct `sifr check` probe reports `SIFR-TYPE-0002` at the iterable argument

### DIFF
--- a/crates/sifr_codegen/src/intrinsic_method_emitters/registry_helpers.rs
+++ b/crates/sifr_codegen/src/intrinsic_method_emitters/registry_helpers.rs
@@ -565,10 +565,7 @@ pub(super) fn registry_call_callable_with_owned_args(
                 args: vec![lowered_arg],
             };
         } else if !param_is_option && arg_is_option && !option_value_adapted {
-            lowered_arg = RustEmitter::force_unwrap_option_expr_for_ir(
-                lowered_arg,
-                "compiler-verified callable argument should be Some",
-            );
+            return None;
         }
 
         if matches!(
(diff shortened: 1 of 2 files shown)


yaseralnajjar:
**SATISFIED** — no blocking in-scope mechanism defect.

**Criteria**
1. `validate_callable_input_types` (`expression_functional_builtins.rs:25`) runs in both `lower_map_call:203` and `lower_filter_call:281`, after arity/signature resolution and before HIR construction. `HirIteratorOp::Map`/`Filter` are constructed only at `:216`/`:302`, so the gate has no bypass. Index arithmetic `args[index + 1]` is bounded by the preceding arity equality checks.
2. Diagnostic is `DiagnosticCode::TYPE_MISMATCH` = `SIFR-TYPE-0002` (`sifr_diagnostics/src/codes/registry.rs:37`), anchored at the iterable argument, phrased consistently with the parallel-map check (`parallel_calls.rs:176`).
3. `registry_call_callable_with_owned_args` (`registry_helpers.rs:567`) returns `None` in place of the force unwrap; all three call sites (`builtin_core_methods.rs:626,652,689`) propagate with `?`. No data-dependent unwrap remains on this path.
4. Some-wrap, `flatten_option_value_for_target`, and `consuming_value_upcast_for_ir` branches are untouched; the union-representation test still holds and the new class-upcast test pins the `Into<Parent>` path.
5. Checking delegates to `Type::is_assignable_to`, the single authority (protocols, classes, unions, literals all handled there); no new fallback branch.

**Non-blocking**
- New lowering tests assert code + message but not the range, which is the part criterion 2 is about; consider a primary-span assertion (pattern at `exact_int_and_fixed_width.rs:708`).
- No e2e/diagnostics fixture for the new rejection, though the repo pairs them for `map_callable_arity_mismatch`.
- Validation list omits the e2e pass suite / merge gate — the place over-rejection would surface. Surveyed `.sifr` fixtures all use concrete `list[int]` callbacks, so risk looks low.
- Corner: `int | None` element with an `Any`/`Unknown`-typed callback parameter is still assignable, so lowering permits it while codegen now bails; the `IteratorCall` fallback (`recursive_exprs.rs:52-60`) would emit a raw `map(...)` that rustc rejects. Fail-closed and better than the old unwrap, but the error surfaces as generated-Rust breakage rather than a diagnostic.

**Pre-existing / out of scope**
- The same "assignable ⇒ Some" assumption still drives force unwraps at `plain_call_args.rs:208`, `registry_method_arg_conventions.rs:66`, `call_args_and_returns.rs:147`, `field_and_stdlib_rewrites.rs:78`, guarded by the same predicate in `regular_calls.rs:144` — a follow-up item.
- `Child | None` → `Parent | None` takes the both-option path, unchanged by this diff.
- Session 01a01407 work excluded.

Full write-up: `/Users/yaseralnajjar/.claude/plans/review-sifr-phase-item-swirling-zebra.md`.

yaseralnajjar:
Exact candidate: `7a9c25505fbb24937f277ae624126ed5ad68f0d3`

Review: Opus returned `SATISFIED` with no blocking in-scope finding: https://github.com/sifr-lang/sifr/pull/3398#issuecomment-5358845681

Validation:
- `cargo test -p sifr_lowering`: 1,022 passed, 1 ignored.
- `cargo test -p sifr_codegen`: 1,080 passed.
- A direct `sifr check` probe reports `SIFR-TYPE-0002` at the incompatible iterable argument.
- Workspace Clippy, formatting, both maintainability guards, the 900-line file-size guard, and diff hygiene passed.

External inputs were not inspected or changed. All `pre_v1` session work remains out of scope.

PR #3401, opened 2026-08-20
Reserve structural construction field temporaries

### LINKED ISSUE #3400: Reserve structural construction field temporaries
Phase item 30 from plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md.

Scope:
- Stop using user field names as local bindings during structural record construction.
- Use reserved per-field temporaries and explicit record-field initializers.
- Preserve inherited-parent constructor argument order and phantom-marker initialization.
- Add package-neutral source and codegen coverage where an earlier field is named like a later default factory.

Acceptance:
- An earlier field named `description` cannot shadow a later `default_factory=description` call.
- Generated record initialization maps every source field explicitly from its reserved temporary.
- Inherited fields still feed the parent constructor in declaration order.
- The static-class-adapter fixture builds and runs with the collision case.

Out of scope: all work and failures owned by the separate pre_v1 session 01a01407-6090-7a53-8ebc-c1c8f8cf0eec.

yaseralnajjar:
Closes #3400.

## Summary
- construct structural fields into one reserved tuple binding so no per-field local can shadow any default expression
- initialize record fields explicitly from indexed compiler temporaries
- retain inherited parent-constructor argument order and escaped Rust field names
- certify the real package-neutral fixture with a `description` field before `default_factory=description`

## Validation
- `cargo test -p sifr_codegen`: 1,080 passed
- `cargo test -p sifr_driver adapter_defaults`: 15 passed
- static-class-adapter fixture built and ran, printing `attached-contract`
- both maintainability guards
- 900-line file-size guard
- `git diff --check`

### DIFF
--- a/crates/sifr_codegen/src/lib_codegen_tests/structural_impl_demand_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/structural_impl_demand_codegen_tests.rs
@@ -49,11 +49,12 @@ fn structural_construction_uses_checked_defaults_for_missing_fields() {
 fn structural_construction_checks_required_fields_before_default_factories() {
     let mut payload = payload_class();
     payload.fields = vec![
+        ("description".to_string(), Type::Str),
         ("names".to_string(), Type::List(Box::new(Type::Str))),
         ("required".to_string(), Type::Int),
     ];
     payload.field_defaults = vec![(
-        0,
+        1,
         HirExpr::Call {
             func: "description".to_string(),
             args: Vec::new(),
@@ -61,18 +62,27 @@ fn structural_construction_checks_required_fields_before_default_factories() {
             ty: Type::List(Box::new(Type::Str)),
         },
     )];
-    payload.field_default_identities = vec![(0, "callable[main.description]".to_string())];
+    payload.field_default_identities = vec![(1, "callable[main.description]".to_string())];
     let module = module(vec![structural_function()], vec![payload]);
 
     let generated = generate_rust(&module);
     let required_check = generated
-        .find("if __sifr_child_nodes[1].is_none()")
+        .find("if __sifr_child_nodes[2].is_none()")
         .unwrap_or_else(|| panic!("required-field precheck must be emitted: {generated}"));
     let factory_call = generated
         .find("None => description()")
         .expect("checked default factory must remain callable");
 
     assert!(required_check < factory_call, "{generated}");
+    let first_field = generated
+        .find("let (__sifr_field_0, __sifr_field_1, __sifr_field_2,) =")
+        .expect("field construction must use reserved tuple temporaries");
+    assert!(first_field < factory_call, "{generated}");
+    assert!(!generated.contains("let description ="), "{generated}");
+    assert!(
+        generated.contains("description: __sifr_field_0, names: __sifr_field_1"),
+        "record initialization must map reserved temporaries explicitly: {generated}"
+    );
     assert!(
         generated.contains("let __sifr_description = __sifr_source.node(__sifr_node)?;"),
         "generated locals must not shadow default callables: {generated}"
@@ -387,8 +397,8 @@ fn structural_impls_escape_rust_keyword_field_identifiers() {
     let generated = generate_rust_multi(&[("models", &models), ("api", &api)]);
     let models_rust = generated.get("models").expect("models module is generated");
 
-    assert!(models_rust.contains("let r#type ="));
-    assert!(models_rust.contains("Ok(Self { r#type })"));
+    assert!(models_rust.contains("let (__sifr_field_0,) ="));
+    assert!(models_rust.contains("Ok(Self { r#type: __sifr_field_0 })"));
     assert!(models_rust.contains("&self.r#type"));
     assert!(models_rust.contains("RecordField"));
     assert!(models_rust.contains("\"type\""));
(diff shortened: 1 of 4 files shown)


yaseralnajjar:
## Verdict: **SATISFIED**

Inspected at the exact candidate SHA `e048c8fc5` (worktree clean, HEAD = candidate), diff vs `eea372f99`, plus the surrounding authorities: `structural_impl_codegen.rs:532-690`, `structural_record_fields.rs`, `render_expr_and_blocks.rs:661` (`render_identifier`), `structural_identity_codegen.rs` default-identity path, and `class_declarations.rs:230-270` (default-index convention).

### Criterion-by-criterion

1. **No shadowing of later defaults by earlier field names** — the only locals the construct body now introduces are `__sifr_description`, `__sifr_child_nodes`, `__sifr_edge`, `__sifr_field_index`, `__sifr_child`, and `__sifr_field_N`. No user-derived identifier is bound anywhere in the body, so `default_factory=description` resolves to the module function regardless of a `description` field. Confirmed by `structural_impl_codegen.rs:576-591` and the negative assertion `!generated.contains("let description =")`.
2. **Single tuple binding, defaults evaluated on the RHS** — `field_initializers` are bare `match` expressions joined into one `let (…,) = (…,);` (`:580-591`). No `__sifr_field_N` is in scope while any match arm is evaluated. `required_prechecks` still precede `field_values` in the body template (`:651`), so the arity check keeps firing before any factory call (asserted by `required_check < factory_call` and `first_field < factory_call`).
3. **Explicit source-name mapping incl. keywords** — record initializers are `{render_identifier(name)}: __sifr_field_{index}` (`:624-635`), so `r#type: __sifr_field_0`; the struct-decl side uses the same renderer, and the keyword test now asserts exactly that pair.
4. **Parent order / local fields / phantom** — `structural_record_fields` places parent fields first in the parent's `Type::Class` field order, so `inherited_values` (`:608-613`) is index-ordered and identical to the previous name-ordered filter; `<Parent>::new(__sifr_field_0)` in both inheritance tests. Non-inherited fields and the `__sifr_type_marker` push are unchanged in shape and position.
5. **Empty / single field** — empty: `field_initializers` empty → `field_values` is `""` and the `fields.is_empty()` `collect_children` branch plus `Ok(Self { })` are untouched. Single field: the trailing comma makes a genuine 1-tuple/1-element pattern (`let (__sifr_field_0,) = (match … ,);`), asserted by the keyword test.
6. **Fixture** — `app.sifr:51-52` adds a real `description: str = contract_field(default="field")` field *before* `descriptions: list[str] = contract_field(default_factory=description)`, with the module-level `def description()` at `:15`. That is precisely the collision the old sequential-`let` codegen would mis-resolve, and `:115` asserts the value. Reported to build a release binary and run.

Mutable-reborrow soundness of `(f(__sifr_source, …)?, f(__sifr_source, …)?, …)` holds: each reborrow's region ends with its call because `structural_construct_at` returns an owned `Result<Self, StructuralContractError>` unrelated to `S`'s lifetime — and the fixture's 6-field release build is real rustc evidence of that.

I also checked the default-index convention for a mismatch risk: the frontend records `class_field_defaults` with `parent_field_count + own_index` (`class_declarations.rs:250-252`), which matches the inherited-first index space used by `structural_construct_impl`. No misalignment introduced or exposed.

No blocking in-scope mechanism defect found.

### Non-blocking suggestions
- Temporary naming diverges from the established constructor path, which uses `__sifr_field_init_N` with an explicit type annotation (`Self { label: __sifr_field_init_0 }`). Aligning the prefix (or noting the deliberate split) would make the two field-materialization sites read as one convention.
- Test coverage for the new ordering is single-inherited-field only (`<Parent>::new(__sifr_field_0)`). A two-inherited-field case would pin the declaration-order contract of criterion 4 against future reordering; today it is only guaranteed structurally.
- No assertion pins the zero-field case to "no tuple binding emitted"; it is covered only indirectly by existing empty-record tests.
- The emitted tuple is one long line; a `\n`-separated tuple literal would keep `emit` output readable without changing semantics.

### Pre-existing / out-of-scope observations
- The `__sifr` prefix is not reserved by the frontend — `grep` finds no rejection of `__sifr*` user identifiers, and `render_identifier` passes user names through unmangled. A user symbol literally named `__sifr_field_0` (or `__sifr_description`, `__sifr_source`) would reintroduce the same shadowing class. This is a codebase-wide convention hazard, not something this diff creates or worsens; worth a frontend reservation as its own item.
- If `parent_type` is `Some` while `parent_class` is `None`, inherited temporaries would be bound with no consumer and the struct literal would miss those fields. Unchanged by this diff (the previous code had the identical shape), and the `__`-prefixed names actually suppress the unused-variable warning the old user-named locals would have produced.

yaseralnajjar:
Exact candidate: `e048c8fc53d914ab85e8b48b0a77be79907c057c`

Review: Opus returned `SATISFIED` with no blocking in-scope finding: https://github.com/sifr-lang/sifr/pull/3401#issuecomment-5359119093

Validation:
- `cargo test -p sifr_codegen`: 1,080 passed.
- `cargo test -p sifr_driver adapter_defaults`: 15 passed.
- The package-neutral static-class-adapter fixture built a release binary and ran successfully with the collision, printing `attached-contract`.
- Workspace Clippy, formatting, both maintainability guards, the 900-line file-size guard, and diff hygiene passed.
- The private target crossed 20 GiB, so it was confirmed unused and cleaned before the long gates; no other target was touched.

External inputs were not inspected or changed. All `pre_v1` session work remains out of scope.

PR #3404, opened 2026-08-20
Isolate ordinary if branch cache state

### LINKED ISSUE #3403: Isolate string-cache state across ordinary if branches
Phase item 31. Make branch-local string-cache declarations transactional in the ordinary non-union if lowering path. A local declared and indexed in a branch must not suppress the cache declaration for a later same-named local outside the branch. Add focused code-generation regression coverage. This item excludes the already-closed speculative union mechanism and all pre_v1 work.

yaseralnajjar:
Closes #3403.

Makes string-character cache bookkeeping transactional for every lowered if branch, including ordinary non-union then/elif/else bodies. A focused source regression verifies that a branch-local cache cannot suppress a later same-named cache declaration in its enclosing block.

Validation before freeze:
- cargo test -p sifr_codegen (1081 passed)
- cargo clippy -p sifr_codegen -- -D warnings
- cargo fmt --check
- git diff --check
- touched first-party files remain below 900 lines

### DIFF
--- a/crates/sifr_codegen/src/lib_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests.rs
@@ -79,6 +79,8 @@ mod receiver_codegen_tests;
 #[cfg(test)]
 mod recursive_node_codegen_tests;
 #[cfg(test)]
+mod string_char_cache_branch_codegen_tests;
+#[cfg(test)]
 mod string_concat_codegen_tests;
 #[cfg(test)]
 mod structural_impl_demand_codegen_tests;
--- a/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_branch_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_branch_codegen_tests.rs
@@ -24,3 +24,28 @@ def count_text(enabled: bool, flag: bool) -> int:
         "branch and outer declarations need separate caches:\n{generated}"
     );
 }
+
+#[test]
+fn walrus_if_branch_string_cache_does_not_suppress_later_declaration() {
+    let generated = generate_rust_from_source(
+        r#"
+def count_text(enabled: bool) -> int:
+    selected: int = 0
+    if enabled:
+        if (value := 1) > 0:
+            text: str = "left"
+            selected = len(text)
+        text: str = "right"
+        selected += len(text)
+    return selected
+"#,
+    );
+
+    assert_eq!(
+        generated
+            .matches("let __sifr_chars_text: Vec<char> = text.chars().collect")
+            .count(),
+        2,
+        "walrus branch and outer declarations need separate caches:\n{generated}"
+    );
+}
--- a/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_branch_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_branch_codegen_tests.rs
@@ -1,0 +1,26 @@
+use super::generate_rust_from_source;
+
+#[test]
+fn branch_local_string_cache_does_not_suppress_later_declaration() {
+    let generated = generate_rust_from_source(
+        r#"
+def count_text(enabled: bool, flag: bool) -> int:
+    selected: int = 0
+    if enabled:
+        if flag:
+            text: str = "left"
+            selected = len(text)
+        text: str = "right"
+        selected += len(text)
+    return selected
+"#,
+    );
+
+    assert_eq!(
+        generated
+            .matches("let __sifr_chars_text: Vec<char> = text.chars().collect")
+            .count(),
+        2,
+        "branch and outer declarations need separate caches:\n{generated}"
+    );
+}
--- a/crates/sifr_codegen/src/stmt_support_emitter/if_condition_lowering.rs
+++ b/crates/sifr_codegen/src/stmt_support_emitter/if_condition_lowering.rs
@@ -582,7 +582,7 @@ impl RustEmitter {
         }))
     }
 
-    fn try_lower_if_branch_for_ir(
+    pub(crate) fn try_lower_if_branch_for_ir(
         &mut self,
         body: &[HirStmt],
     ) -> Result<Option<Vec<RustStmt>>, crate::CodegenError> {
--- a/crates/sifr_codegen/src/stmt_support_emitter/if_condition_lowering.rs
+++ b/crates/sifr_codegen/src/stmt_support_emitter/if_condition_lowering.rs
@@ -358,8 +358,7 @@ impl RustEmitter {
                                     .collect::<Vec<_>>()
                             })
                             .unwrap_or_default();
-                        let Some(lowered_else_body) =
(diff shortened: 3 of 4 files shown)


yaseralnajjar:
**NOT SATISFIED**

## What the commit does (verified)

`154aa35` renames `try_lower_speculative_branch_for_ir` → `try_lower_if_branch_for_ir` (body unchanged — save/restore of `string_char_cache_vars` on success, decline, and error: `crates/sifr_codegen/src/stmt_support_emitter/if_condition_lowering.rs:585-604`) and routes the ordinary path's branch bodies through it: else body (`:489`) and every then/elif body via `try_lower_if_clause_for_ir` (`:515`). The union path (`:361`, `:402`) keeps the same helper plus its outer transaction (`:345`/`:415`), so the rename is behavior-neutral there.

Evidence gathered:
- New test is a real regression guard: at base `1245ed4c` (source tree copied to `/tmp`, unmodified repo) `branch_local_string_cache_does_not_suppress_later_declaration` **fails** (`left: 1, right: 2`, emitting `selected += __sifr_chars_text.len()` with no in-scope declaration); at `154aa35` it passes.
- `cargo test -p sifr_codegen`: 1081 passed, 0 failed. `cargo clippy --workspace -- -D warnings`: clean. `cargo fmt --check`: clean. `check_hir_maintainability_guardrails.py`: PASS. Touched files 764 / 100 / 26 lines — under 900.
- End-to-end emits at HEAD are correct for then-nested (`9`), elif and else branches, and the always-exits/`let-else` fast path (`crates/.../if_condition_lowering.rs:420-486`, verified by running the programs).
- Suppression mechanism confirmed: `string_char_cache_init_stmt_for_local` returns `None` when `string_char_cache_vars.contains_key(name)` (`crates/sifr_codegen/src/string_char_cache.rs:664-671`), so a leaked entry yields a *use without a declaration*, i.e. non-compiling Rust — not merely a missed optimization.

## Blocking finding (in scope)

**Ordinary non-union `if` with a walrus condition is still not transactional** — `crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs:242`.

`try_lower_structured_if_stmt` is the top-level statement entry (`crates/sifr_codegen/src/lib_emitter_state.rs:796`) and only delegates to the fixed `try_lower_if_stmt_for_ir` at `structured_return_if_while.rs:425`. The earlier walrus special case (no elif, no else) lowers `then_body` with a raw `try_lower_stmt_block_for_ir` and no save/restore, so the exact item-31 property fails for a plain non-union `if`:

```python
if (n := pick(v)) > 0:
    text: str = "left"
    s = len(text)
text: str = "right"
s += len(text)
```
At `154aa35` this emits the branch-scoped `let __sifr_chars_text` only, and `sifr run` fails with `error[E0425]: cannot find value \`__sifr_chars_text\` in this scope` (SIFR-BUILD-0005). Same defect class, same statement kind; the fix is to route that body through the new `try_lower_if_branch_for_ir` (needs `pub(crate)`), and the same decline path (`return Ok(false)` at `:244`) must not leak state either.

## Newly discovered, out of scope — record as later items

1. **Statement-level `isinstance` union path leaks branch cache state** — `structured_return_if_while.rs:297, :337, :370, :388` use bare `try_lower_stmt_block_for_ir`. Because branches are lowered in reverse (`:330`), the last branch's entry suppresses the earlier branch's declaration: `if isinstance(value, int): text: str = "int"; s = len(text)` emits `s = __sifr_chars_text.len()` with no declaration in that arm. This is a *different* path from the closed item-23 mechanism in `if_condition_lowering.rs`, and it shadows that fix for statement contexts. Not item 31's non-union scope.
2. **Loop bodies leak the same state** — a `for` body declaring a cached local suppresses the post-loop declaration (`loops_try_finally.rs:40/121`, `async_with_and_for.rs:56/95/267`), again producing `__sifr_chars_text` uses with no in-scope declaration. Loop lowering is outside the `if` scope.

Both reproduce at `154aa35` and at the base, so neither is a regression from this commit.

## Non-blocking notes

- `if_condition_lowering.rs:425` (always-exits fast path) restores state on decline (`:426`) and on each success (`:444`, `:454`, `:478`, `:485`) but propagates `Err` via `?` without restoring — unobservable today since a codegen error aborts, but inconsistent with the helper's error handling.
- Ordering inside that fast path is inconsistent (case at `:429`/`:461` builds patterns before restore, `:451` after); harmless because `option_binding_pattern_for_ir`/`option_binding_value_expr_for_ir` do not read cache state (`condition_lowering.rs:129-157`).
- Test adequacy: only the nested-then shape is asserted. else/elif and `let-else` shapes are correct but unguarded; one added assertion per shape would lock them in.
- The active issue (`plans/issues/active/ad-hoc-static-class-adapters-and-pydantic-ergonomics.md:2919`) still lists item 31 as the next action; per the item-30 precedent (`daffd6bb2`) the record lands in a follow-up PR, so this is expected rather than a violation.
- `if_condition_lowering.rs` is at 764/900 lines; the added helper leaves little headroom.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- pre-existing issue: `crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs:297,337,370,388` — the statement-level `isinstance`-union branches still lower with `try_lower_stmt_block_for_ir`, so a cached local inside a union arm can suppress a later same-named declaration. Out of item-31 scope (non-union ordinary if), matches the previously recorded later-item finding.
- pre-existing issue: loop/while/for and with/try bodies (`stmt_block.rs:719,727,794`, `loops_try_finally.rs`, `async_with_and_for.rs`, `python_context/*`) still leak `string_char_cache_vars` out of the body; only the `For` target key is snapshotted (`stmt_block.rs:783`). Later-item work, not an item-31 regression.
- suggestion: `if_condition_lowering.rs:425` (the `codegen_body_always_exits` fast path) keeps a hand-rolled snapshot/restore around `try_lower_stmt_block_for_ir` with five restore sites; routing it through `try_lower_if_branch_for_ir` would remove the duplicated bookkeeping and the risk of a future early-return skipping a restore.
- suggestion: the union path (`if_condition_lowering.rs:344,412`) now wraps an outer snapshot/restore around branch calls that already restore internally; the outer transaction is redundant, though harmless since both restore the same snapshot.
- suggestion: `if_condition_lowering.rs:585-600` — the three match arms each do the identical restore; a single restore after capturing the result (or a scope guard) reads cleaner. Test assertion messages at lines ~701 and ~723 still say "speculative lowering" after the rename.

yaseralnajjar:
Item 31 exact-candidate gate evidence for `be9652442ea740386fe1b9fceadbb47aa88464e3`:

- Create-PR gate: run once. All setup and guardrails passed; terminal stop only in the Rust-interop matrix on the two externally owned inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty-placeholder `examples/method_slot_runtime/src/schema_program.sifr`. Compatibility matrix 40/40, wall time 117.66s, warm budget passed, no swaps.
- Merge gate: run once. Same preceding guards passed and the same two external inputs stopped the matrix. Compatibility matrix 40/40, wall time 119.64s, warm budget passed, no swaps.

PR #3407, opened 2026-08-20
Isolate statement-level union branch cache state

### LINKED ISSUE #3406: Isolate statement-level union branch cache state
Phase item 32. Make every statement-level isinstance-union branch transactional for string-character cache bookkeeping. A cached local inside one union arm must not suppress a same-named declaration in another arm or the enclosing block. Add focused source regression coverage. Ordinary non-union if state is closed in item 31. All pre_v1 work is out of scope.

yaseralnajjar:
Closes #3406.

Routes every statement-level isinstance-union body through the shared transactional if-branch lowering helper. The source regression proves separate cache declarations for narrowed then, elif, and remaining else scopes.

Validation before freeze:
- cargo test -p sifr_codegen (1083 passed)
- cargo fmt --check
- git diff --check
- touched files remain below 900 lines

### DIFF
--- a/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_branch_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_branch_codegen_tests.rs
@@ -49,3 +49,31 @@ def count_text(enabled: bool) -> int:
         "walrus branch and outer declarations need separate caches:\n{generated}"
     );
 }
+
+#[test]
+fn statement_union_branches_use_separate_string_caches() {
+    let generated = generate_rust_from_source(
+        r#"
+def count_text(value: int | str | float) -> int:
+    selected: int = 0
+    if isinstance(value, int):
+        text: str = "int"
+        selected = len(text)
+    elif isinstance(value, str):
+        text: str = "string"
+        selected = len(text)
+    else:
+        text: str = "float"
+        selected = len(text)
+    return selected
+"#,
+    );
+
+    assert_eq!(
+        generated
+            .matches("let __sifr_chars_text: Vec<char> = text.chars().collect")
+            .count(),
+        3,
+        "each union branch needs its own cache declaration:\n{generated}"
+    );
+}
--- a/crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs
+++ b/crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs
@@ -294,7 +294,7 @@ impl RustEmitter {
                                 .collect::<Vec<_>>()
                         })
                         .unwrap_or_default();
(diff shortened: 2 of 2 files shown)


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- pre-existing issue: the union-chain lowering in `crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs:263-353` is a near-duplicate of `if_condition_lowering.rs:324-417`. Item 32 had to fix the same defect twice, and the two copies still differ (the expression-level copy wraps the whole chain in a snapshot/restore at `if_condition_lowering.rs:345/415`; the statement-level copy relies solely on per-branch restore). Both are behaviorally equivalent today — nothing between the branch lowerings mutates `string_char_cache_vars`, since `resolve_union_enum_name` and `collect_mutated_vars` are non-mutating — but the duplication is drift-prone.
- pre-existing issue: `try_lower_stmt_block_for_ir` is still called without cache-state isolation for loop bodies/else (`loops_try_finally.rs:40,48,121,140,270,519`), `while` bodies (`stmt_block.rs:719,727`), `with`/async blocks (`async_with_and_for.rs`, `python_context/*`), and `finalbody` (`loops_try_finally.rs:577`). Same class of leak, outside item 32's scope.
- suggestion: the new test only exercises the multi-arm `if/elif/else` chain path (`structured_return_if_while.rs:266-352`). The elif-less union path that lowers to a `match` (`:356-422`, including the changed calls at `:370` and `:388`) is not covered by a dedicated regression, nor is the "union arm followed by a same-named declaration in the enclosing block" shape that the acceptance text calls out (item-31 tests cover that shape only for ordinary ifs). Two small additional sources would close both gaps.

Verification notes: the new test is discriminating — at base, the `else` arm was lowered first and leaked `text` into `string_char_cache_vars`, suppressing the two remaining arms (1 declaration instead of 3). `try_lower_if_branch_for_ir` (`if_condition_lowering.rs:585-604`) restores the snapshot on success, decline, and error, so all five statement-level call sites now roll back on every outcome, and the enclosing block's state is unchanged after the chain. Restoring after a *successful* arm is correct here because each arm lowers into a real Rust block, so an arm-local cache is not in scope afterwards.

yaseralnajjar:
Item 32 exact-candidate gate evidence for `53e56dfaf7ae18a6d141131a171a9fc0b9db8a03`:

- Create-PR gate: run once. All setup and guardrails passed; terminal stop only in the Rust-interop matrix on the two externally owned inputs. Compatibility matrix 40/40, wall time 138.72s, warm budget passed, no swaps.
- Merge gate: run once. The same preceding guards passed and the same two external inputs stopped the matrix. Compatibility matrix 40/40, wall time 125.80s, warm budget passed, no swaps.

PR #3410, opened 2026-08-20
Isolate control-flow body cache state

### LINKED ISSUE #3409: Isolate non-if control-flow body cache state
Phase item 33. Audit loop, with, try, finalbody, and async nested-body lowering for string-character cache state. Any body emitted into a Rust lexical scope must restore string_char_cache_vars after success, decline, or error. Add focused source evidence for every reachable body mechanism. If the audit finds distinct mechanisms that cannot be closed as one bounded change, record later items. All pre_v1 work is out of scope.

yaseralnajjar:
Closes #3409.

Adds one shared scoped statement-block transaction for string-character cache state and applies it across loops, loop else bodies, with/async-with bodies, async-for bodies, try closures, handlers, final bodies, and Python contexts. For-loop target cache setup is included in the same rollback boundary. Focused source regressions cover while/for/loop-else, native with, try/handler/finally, async with, and async for.

Validation before freeze:
- cargo test -p sifr_codegen (1087 passed)
- cargo fmt --check
- maintainability and diff checks
- file-size guardrails: 3202 files pass; touched files remain below 900 lines

### DIFF
--- a/crates/sifr_codegen/src/lib_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests.rs
@@ -81,6 +81,8 @@ mod recursive_node_codegen_tests;
 #[cfg(test)]
 mod string_char_cache_branch_codegen_tests;
 #[cfg(test)]
+mod string_char_cache_control_flow_codegen_tests;
+#[cfg(test)]
 mod string_concat_codegen_tests;
 #[cfg(test)]
 mod structural_impl_demand_codegen_tests;
--- a/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_control_flow_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/string_char_cache_control_flow_codegen_tests.rs
@@ -1,0 +1,159 @@
+use super::generate_rust_from_source;
+
+fn cache_declaration_count(generated: &str, name: &str) -> usize {
+    generated
+        .matches(&format!(
+            "let __sifr_chars_{name}: Vec<char> = {name}.chars().collect"
+        ))
+        .count()
+}
+
+#[test]
+fn loop_bodies_do_not_suppress_enclosing_string_caches() {
+    let generated = generate_rust_from_source(
+        r#"
+def count_text(enabled: bool) -> int:
+    total: int = 0
+    if enabled:
+        while total < 1:
+            while_text: str = "while"
+            total += len(while_text)
+        while_text: str = "after while"
+        total += len(while_text)
+
+        while total < 0:
+            pass
+        else:
+            while_else_text: str = "while else"
+            total += len(while_else_text)
+        while_else_text: str = "after while else"
+        total += len(while_else_text)
+
+        for value in [1]:
+            for_text: str = "for"
+            total += value + len(for_text)
+        for_text: str = "after for"
+        total += len(for_text)
+
+    for value in [1]:
+        total += value
+    else:
+        for_else_text: str = "for else"
+        total += len(for_else_text)
+    if enabled:
+        for_else_text: str = "after for else"
+        total += len(for_else_text)
+    return total
+"#,
+    );
+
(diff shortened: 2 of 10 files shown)


yaseralnajjar:
**Verdict: SATISFIED**

**Blocking findings:**
- None.

Basis (verified at the exact candidate):
- Call-site inventory is complete: `try_lower_stmt_block_for_ir_inner` is the only block-lowering entry point, and all 20 reachable body call sites now either go through `try_lower_scoped_stmt_block_for_ir` (`stmt_block_helpers.rs:505`) or use an explicit snapshot that subsumes the for-target init. The only remaining bare call is `if_condition_lowering.rs:425` (if-then LetElse speculation), which is if state closed in items 31/32 and carries its own snapshot/restore.
- Restoration covers success, decline (`Ok(None)`), and error: the helper restores before returning `result`, and both for-loop paths (`loops_try_finally.rs:116-124`, `stmt_block.rs:788-799`) restore before `?` propagation — the previous `stmt_block.rs` code leaked the map on the error path, so this is a strict improvement.
- For-target sequencing is correct: the snapshot is taken after iterable lowering (so iterable-scope caches persist) and before `string_char_cache_init_stmt_for_loop_target`, which is the only mutator on that path via `build_string_char_cache_init_stmt` (`string_char_cache.rs:701-704`). The target-init statement is still inserted at index 0 of the loop body, inside the Rust `for` braces.
- Rollback is sound because the map is insert-only within a scope (only `string_char_cache.rs:36` and `:703` mutate it), so restoring cannot resurrect a stale binding; it can only force a fresh re-init in the enclosing scope.
- Control-flow stack interactions preserved: `loop_else_stack` push/pop, `active_timeout_durations`, `python_context_envelope_depth`, and `try_closure_*` save/restore ordering are unchanged relative to base; the cache restore is inserted inside those windows without reordering them.
- `HirStmt::Match` arms are not reachable in this emitter (no `Match` branch in the dispatch chain — the block declines), and `AsyncFor` is always constructed with `else_body: None` (`sifr_lowering/src/lower/async_for.rs:233`), so those are not uncovered bodies.
- Tests discriminate: `cache_declaration_count(...) == 2` fails before the fix (the enclosing re-declaration would reuse the inner block's out-of-scope cache name) for while, while/else, for, for/else, native with, try body/handler/finally, async with, and async for.
- Guardrails: working tree clean; largest touched source `stmt_block.rs` at 889 lines, under the 900-line limit.

**Follow-up findings:**
- suggestion: no direct test covers the Python-context `with` bodies (`python_context/sync.rs:15`, `python_context/async_context.rs:75`) or the user-defined `async with` body (`async_with_and_for.rs:95`); those sites are only covered transitively by the shared helper. Adding one Python-context case would close the coverage gap for a distinct emission shape.
- pre-existing issue: `if_condition_lowering.rs:425` restores the speculative snapshot on every success and decline path but not on the `?` error path. Harmless today because a `CodegenError` aborts the run, and it is if-scope owned by items 31/32 — no action needed for item 33.
- pre-existing issue: `loops_try_finally.rs:576` appends the lowered `finalbody` inline into the enclosing statement list rather than a nested Rust block, so the scoped rollback can cause a later enclosing statement to emit a second `let __sifr_chars_x` shadowing the finalbody's. Legal Rust and semantically correct (fresh collect), but slightly redundant output; a mechanism-level tidy-up for a later item, not a defect in this one.

yaseralnajjar:
Item 33 exact-candidate gate evidence for `b7062d3500b3c24a6adae7f97197aab12defee06`:

- Create-PR gate: run once. All setup and guardrails passed; terminal stop only in the Rust-interop matrix on the two externally owned inputs. Compatibility matrix 40/40, wall time 163.44s, warm budget passed, no swaps.
- Merge gate: run once. The same preceding guards passed and the same two external inputs stopped the matrix. Compatibility matrix 40/40, wall time 147.81s, warm budget passed, no swaps.

PR #3413, opened 2026-08-20
Certify installed static class adapters

### LINKED ISSUE #3412: Certify installed static-class adapter fixture with sysroot runtime paths
M12 certification exposed an in-scope boundary defect: `sysroot_release:boundary-equivalence` copies the package-neutral `static_class_adapter` fixture outside the source tree, but its Cargo manifest retains a repository-relative `sifr_runtime` path. The installed compiler then resolves `/private/var/folders/.../crates/sifr_runtime` instead of the installed sysroot.

Scope: bind each temporary certification copy to the exact source or installed sysroot runtime crate, keep the repository fixture unchanged, add focused fail-closed helper coverage, and rerun source/installed boundary certification.

This does not own the two separately recorded Rust-interop gate inputs or any `pre_v1` worktree/session state.

yaseralnajjar:
Closes #3412

Completes the installed/source static-class-adapter certification boundary: binds temporary package fixtures to the selected sysroot runtime, packages and validates `sifr_structural_identity`, keeps toolchain copies independent, and makes the cache edit probe assertion-safe.

Exact reviewed candidate: `0d90ecb6eef2c25d7dcc0bc32ea80649bd52b07c`.

### DIFF
--- a/crates/sifr/tests/sysroot_cli.rs
+++ b/crates/sifr/tests/sysroot_cli.rs
@@ -144,7 +144,7 @@ fn write_skeleton(root: &Path) {
     std::fs::write(
         root.join("Cargo.toml"),
         r#"[workspace]
-members = ["crates/sifr_runtime", "crates/sifr_stdlib"]
+members = ["crates/sifr_runtime", "crates/sifr_structural_identity", "crates/sifr_stdlib"]
 resolver = "2"
 "#,
     )
@@ -152,6 +152,7 @@ resolver = "2"
     std::fs::write(root.join("Cargo.lock"), "").expect("lockfile");
     std::fs::write(root.join(".cargo/config.toml"), "").expect("cargo config");
     write_minimal_crate(root, "sifr_runtime");
+    write_minimal_crate(root, "sifr_structural_identity");
     write_minimal_crate(root, "sifr_stdlib");
 }
 
--- a/crates/sifr_sysroot/src/layout.rs
+++ b/crates/sifr_sysroot/src/layout.rs
@@ -35,6 +35,8 @@ pub struct SysrootPaths {
     pub stdlib_private_sources: PathBuf,
     pub runtime_crate: PathBuf,
     pub runtime_crate_manifest: PathBuf,
+    pub structural_identity_crate: PathBuf,
+    pub structural_identity_crate_manifest: PathBuf,
     pub stdlib_crate: PathBuf,
     pub stdlib_crate_manifest: PathBuf,
     pub cargo_manifest: PathBuf,
@@ -47,6 +49,7 @@ impl SysrootPaths {
     #[must_use]
     pub fn from_root(root: &Path) -> Self {
         let runtime_crate = root.join("crates").join("sifr_runtime");
+        let structural_identity_crate = root.join("crates").join("sifr_structural_identity");
         let stdlib_crate = root.join("crates").join("sifr_stdlib");
         let stdlib_root = stdlib_source_root(root);
         Self {
@@ -56,6 +59,8 @@ impl SysrootPaths {
             stdlib_root,
             runtime_crate_manifest: runtime_crate.join("Cargo.toml"),
             runtime_crate,
+            structural_identity_crate_manifest: structural_identity_crate.join("Cargo.toml"),
+            structural_identity_crate,
             stdlib_crate_manifest: stdlib_crate.join("Cargo.toml"),
             stdlib_crate,
             cargo_manifest: root.join("Cargo.toml"),
@@ -88,6 +93,12 @@ impl SysrootPaths {
             root,
             &self.runtime_crate_manifest,
             "crates/sifr_runtime/Cargo.toml",
+        )?;
+        require_file(
+            binary_path,
+            root,
+            &self.structural_identity_crate_manifest,
+            "crates/sifr_structural_identity/Cargo.toml",
         )?;
         require_file(
             binary_path,
@@ -187,6 +198,13 @@ fn validate_workspace_manifest(
         ));
     };
     require_workspace_member(binary_path, root, manifest, members, "crates/sifr_runtime")?;
+    require_workspace_member(
+        binary_path,
+        root,
+        manifest,
+        members,
+        "crates/sifr_structural_identity",
+    )?;
     require_workspace_member(binary_path, root, manifest, members, "crates/sifr_stdlib")?;
     if workspace.get("resolver").and_then(toml::Value::as_str) != Some("2") {
         return Err(SysrootError::new(
--- a/crates/sifr_sysroot/src/tests.rs
+++ b/crates/sifr_sysroot/src/tests.rs
@@ -189,6 +189,10 @@ fn layout_validation_checks_all_skeleton_assets() {
         ("stdlib_private_sources", "lib/sifr/stdlib/_sifr"),
         ("vendor", "vendor"),
         ("runtime_manifest", "crates/sifr_runtime/Cargo.toml"),
+        (
+            "structural_identity_manifest",
+            "crates/sifr_structural_identity/Cargo.toml",
+        ),
         ("stdlib_manifest", "crates/sifr_stdlib/Cargo.toml"),
     ] {
         let root = complete_sysroot(label, COMPILER_SIFR_VERSION);
@@ -213,7 +217,7 @@ fn workspace_validation_requires_generated_stdlib_member() {
         root.path.join("Cargo.toml"),
         r#"
 [workspace]
-members = ["crates/sifr_runtime"]
+members = ["crates/sifr_runtime", "crates/sifr_structural_identity"]
 resolver = "2"
 "#,
     )
@@ -232,6 +236,35 @@ resolver = "2"
     assert!(error
         .message
         .contains("workspace member crates/sifr_stdlib"));
+}
+
+#[test]
+fn workspace_validation_requires_structural_identity_member() {
+    let root = complete_sysroot("missing_structural_identity_member", COMPILER_SIFR_VERSION);
+    fs::write(
+        root.path.join("Cargo.toml"),
+        r#"
+[workspace]
+members = ["crates/sifr_runtime", "crates/sifr_stdlib"]
+resolver = "2"
+"#,
+    )
+    .expect("workspace manifest");
+    let input = SysrootResolutionInput {
+        explicit_sysroot: Some(root.path.clone()),
+        env_sysroot: None,
+        current_exe: PathBuf::from("/tool/bin/sifr"),
+        current_dir: root.path.clone(),
+        allow_source_tree_development: false,
+    };
+
+    let error =
+        resolve_sysroot_with(&input).expect_err("missing structural identity member should fail");
+
+    assert_eq!(error.kind, SysrootErrorKind::InvalidWorkspace);
+    assert!(error
+        .message
+        .contains("workspace member crates/sifr_structural_identity"));
 }
 
 #[test]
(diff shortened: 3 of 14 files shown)


yaseralnajjar:
NOT SATISFIED

## Blocking finding

**`crates/sifr/tests/sysroot_cli.rs` regresses: 2 tests fail on the candidate.**

The new fail-closed assets in `crates/sifr_sysroot/src/layout.rs:97-102` and `:201-207` (`crates/sifr_structural_identity/Cargo.toml` + workspace member) were propagated to every skeleton builder *except* the `sifr` CLI integration-test skeleton (`crates/sifr/tests/sysroot_cli.rs:122-155`, which still writes `members = ["crates/sifr_runtime", "crates/sifr_stdlib"]` and only `write_minimal_crate` for those two crates).

Verified by running it on the exact candidate SHA:

```
test explicit_sysroot_missing_vendor_reports_boundary_error ... ok
test print_sysroot_uses_explicit_installed_layout_outside_repo ... FAILED
test print_sysroot_json_reports_identity_and_layout_paths ... FAILED

error[SIFR-BUILD-0002]: Sifr sysroot is missing crates/sifr_structural_identity/Cargo.toml
test result: FAILED. 1 passed; 2 failed
```

Failure scenario: `sifr --sysroot <skeleton> --print sysroot` now exits non-zero because `SysrootPaths::validate` requires the new asset, so both `assert!(output.status.success())` calls panic. This is squarely in the changed responsibility (this item added the requirement) and blocks the PR/merge gate, which runs `cargo test -p sifr`. Fix is one line + one `write_minimal_crate(root, "sifr_structural_identity")`, mirroring what was already done in `crates/sifr_sysroot/src/tests.rs:346,358` and `verification/areas/distribution_release/cases/common.sh`.

The `explicit_sysroot_missing_vendor_reports_boundary_error` case still passes only because `require_dir(vendor)` runs before the new `require_file` — that ordering is fine and intentional.

## What does hold

- **Acceptance 1/2/3**: the structural-identity crate is now packaged (`build_release_artifacts.sh:262`), archive-verified (`verify_release_archive.py` REQUIRED_FILES + DIR_PREFIXES), installer-validated (`generate_version_installer.sh:710`), and layout/workspace-validated. `sifr_runtime`'s `structural` feature resolves `sifr_structural_identity = { workspace = true }` → `crates/sifr_structural_identity` relative to the install root, so the installed archive is now self-contained for it.
- **Per-toolchain binding**: `bind_runtime_dependency` rewrites exactly one `sifr_runtime = { path = "…", features = ["structural"] }` line and returns an error otherwise; the fixture manifest matches the regex exactly (`fixtures/static_class_adapter/Cargo.toml:7`). No fallback path.
- **Copy independence**: `runner.py:235` moves `copytree` inside the per-label loop with a label-suffixed destination, so the `api.sifr` cache-edit probe in one toolchain can't leak into the other. `api.sifr` is restored before the result check, and all copies live in the temp tree — the repo fixture is untouched.
- **Cache probe compatibility without weakening**: `print(AttachedModel.describe())` is the only stdout source, and the driver still pins the literal values (`attached-contract` → `attached-contract-edited`, `attached_api_certification.py:76,98`). The three rewritten asserts (`app.sifr:92,100,101`) compare *distinct* owners (`ImportedAlias`, `ConcreteAttached`, `ImportedConcrete`) against `AttachedModel`, so cross-owner equality is still checked rather than made trivially true. `app.sifr` is not consumed by any snapshot or driver test (`adapter_defaults.rs` includes only `api/contract_types/contract/facade`).
- **Fail-closed coverage**: new `workspace_validation_requires_structural_identity_member` test, the `structural_identity_manifest` entry in the all-assets loop, the archive-rejection spec row, the installer preview-target assertion, and the two Python helper cases (missing/duplicate) all target real negatives.
- I found no other unhandled skeleton builder: `sysroot_interop.rs:696-745` constructs `ResolvedSysroot` directly (no `validate`), and `cargo_manifest.rs` / `rust_interop_probe_tests.rs` use synthetic `/opt/sifr` paths.

## Non-blocking notes

- `attached_api_certification.py:26` passes `replacement` as a `re.subn` template, so backslashes in the path would be reinterpreted as replacement escapes (`json.dumps` `\\` collapses back to `\`). Harmless for POSIX temp/repo paths; `lambda _: replacement` would make it exact.
- `SysrootPaths::structural_identity_crate` (the directory field) has no consumer — only the derived manifest path is used. Symmetric with `runtime_crate`, so fine, but it is currently dead surface.
- Not a defect of this diff: `cargo test -p sifr --test e2e test_e2e_fail` fails on the candidate with a `sifr_lowering/src/cfg.rs:300` ICE on `tests/e2e/fail/protocol_bound_unknown_forwarded_typevar.sifr`. Nothing in the diff touches the compiler; unrelated and out of scope. Same for the pre-existing `src/contract.sifr` formatting issue you already flagged.

yaseralnajjar:
SATISFIED

## Blocker closure

`crates/sifr/tests/sysroot_cli.rs:147,155` now adds `crates/sifr_structural_identity` to the skeleton workspace members and calls `write_minimal_crate(root, "sifr_structural_identity")`. That matches exactly what `SysrootPaths::validate` (`crates/sifr_sysroot/src/layout.rs:97-102`) and `validate_workspace_manifest` (`layout.rs:201-207`) now require, so both previously failing CLI integration tests are satisfied by construction. Reported evidence: `cargo test -p sifr --test sysroot_cli` 3 passed.

## Acceptance criteria

1. **Self-contained archives.** `build_release_artifacts.sh:262` packages `crates/sifr_structural_identity`. `crates/sifr_runtime/Cargo.toml` declares `sifr_structural_identity = { workspace = true, optional = true }` behind `features = ["structural"]`, and the archived root `Cargo.toml` supplies `[workspace.dependencies] sifr_structural_identity = { path = "crates/sifr_structural_identity" }`, so the installed path resolves inside the sysroot. Both content hashes (`sysroot_content_sha256` in the shell script, `is_sysroot_content_path` in `verify_release_archive.py:108-113`) enumerate `crates/` by prefix, so the new crate is hashed on both sides without further change.
2. **Both toolchains exercise the fixture.** `runner.py:199-245` now passes a per-label `runtime_crate` (`install_root/crates/sifr_runtime` vs `REPO_ROOT/crates/sifr_runtime`) and, critically, copies a *fresh* fixture per label (`static_class_adapter-{label}`) instead of sharing one mutated tree — that is what makes the installed and source runs independent rather than the second one inheriting the first's rewritten manifest.
3. **Edit-only-api.sifr.** `app.sifr` no longer pins the literal in the cross-type assertions (`ImportedAlias/ConcreteAttached/ImportedConcrete.describe() == AttachedModel.describe()`), so the single edit in `api.sifr` propagates without tripping an assert. The literal is still pinned exactly once, at the stdout comparison in `attached_api_certification.py` (`attached-contract` → `attached-contract-edited`), so the check did not go slack.
4. **Fail-closed.** `bind_runtime_dependency` returns an error unless `subn` reports exactly one match (missing and duplicate both covered by `test_attached_api_certification.py`), and returns errors on read/write `OSError` rather than continuing. Archive side: new entry in `REQUIRED_FILES`/`REQUIRED_DIR_PREFIXES` plus the broken-archive spec at `artifact_broken_sysroot_archives_rejected.sh:86` — files are checked before directory prefixes (`verify_release_archive.py:166` vs `:168`), so deleting the crate dir yields the asserted *file* message. Installer: `require_extracted_file "crates/sifr_structural_identity/Cargo.toml"`. Sysroot: new `require_file` + `require_workspace_member`, both negatively covered (`layout_validation_checks_all_skeleton_assets` entry and `workspace_validation_requires_structural_identity_member`, whose assertion text matches `require_workspace_member`'s `must include workspace member {expected}`).
5. **Skeletons remain valid.** Swept every skeleton in the tree. Updated: `sysroot_cli.rs`, `sifr_sysroot/src/tests.rs` (`write_complete_sysroot` + the two member-list fixtures), `distribution_release/cases/common.sh::make_mock_sysroot_root`, `qualification_fixture.py::write_synthetic_target`, `artifact_generated_installer_all_preview_targets.sh`. The one remaining skeleton without the crate — `crates/sifr_driver/src/build/sysroot_interop.rs:706-744` — is not affected: it constructs a `ResolvedSysroot` literal directly and never calls `resolve_sysroot_with`/`validate` (no such call exists in that file). The repo root `Cargo.toml` already lists the member, so `--sysroot REPO_ROOT` is unaffected.
6. **No fallback / duplicate authority / special case / pre_v1.** The asset requirement has one executable authority (`layout.rs`); packaging, installer, and archive verifier are enforcement mirrors, not competing resolvers. `bind_runtime_dependency` errors instead of defaulting. Nothing in the compiler special-cases the fixture. No pre_v1 references.

The remediation's second change (`subn(lambda _match: replacement, ...)`) is a genuine correctness fix, not cosmetic: with a string replacement, a Windows-style or backslash-bearing crate path would have been reinterpreted as a backreference/escape. The new test asserts the exact TOML escaping `path = "/tmp/runtime \\ \"exact\""`, which is what `json.dumps` correctly produces for a TOML basic string.

Since the remediation commit touched only `sysroot_cli.rs`, the lambda, and the new unit test — none of which alter the boundary-equivalence code path on POSIX — the initial-candidate boundary-equivalence evidence remains valid for the final candidate.

No new mechanism defect found.

## Non-blocking suggestions

- **Stale internal docs.** `internal_docs/distribution_pipeline.md:150-162`, and `internal_docs/sifr_sysroot_and_stdlib_architecture.md` at the archive-contents block (~990-1002), the distributed workspace manifest shape (~398-404), and the boundary-check asset list (~1076-1082) all still enumerate only `crates/sifr_runtime` and `crates/sifr_stdlib`. Not gate-enforced (the doc checks that read these files cover release-facts markers and schema epochs, not these listings), but the sysroot architecture doc is the prose authority for exactly the two lists this diff changed.
- `SysrootPaths::structural_identity_crate` (`layout.rs:38`) has no consumer; only the `_manifest` sibling is read. Harmless as a public field, but `runtime_crate` at least feeds the driver's dependency plan, so the asymmetry may confuse later readers.
- `bind_runtime_dependency`'s regex `path = "[^"]+"` cannot re-match its own output if the bound path contained a quote. Safe today because each label gets a fresh `copytree`, but it makes the function non-idempotent.
- `shutil.copytree` on the fixture root copies its gitignored `target/` directory (~27 MB locally), and this now happens twice per run instead of once. `ignore=shutil.ignore_patterns("target")` would keep it cheap.

yaseralnajjar:
Create-PR gate evidence for exact candidate `0d90ecb6eef2c25d7dcc0bc32ea80649bd52b07c`: the one permitted `scripts/run_all_tests.sh --profile create-pr` run passed cache setup, generated-demo freshness, all preceding guardrails, Rust-interop tiers, the 40/40 compatibility matrix, stale-draft checks, and stable claims. It stopped only at the two separately owned matrix inputs already recorded out of scope: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. The gate was not rerun.

yaseralnajjar:
Merge gate evidence for exact candidate `0d90ecb6eef2c25d7dcc0bc32ea80649bd52b07c`: the one permitted `scripts/run_all_tests.sh` run passed cache setup, generated-demo freshness, all preceding guardrails, Rust-interop tiers, the 40/40 compatibility matrix, stale-draft checks, and stable claims. It stopped only at the same two separately owned matrix inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`. Neither compiler gate will be rerun.

PR #3419, opened 2026-08-20
Probe mapped opaque types against their value type

### LINKED ISSUE #3418: Probe mapped opaque types against their declared value type
M12 certification item 36. Source-package certification for Pydantic-Sifr merge `a44116e188cc6b45cffb297d57b9084467a39e8f` fails for `Url`, `MultiHostUrl`, and `Pattern`: the Rust probe plans an opaque probe for the `structural=` mapping path and emits `StructuralMapping<MappingType>`, although the declared implementation is `StructuralMapping<OpaqueValueType>`.

Scope: resolve every opaque target path as before, but schedule the concrete opaque type/trait probe only for the `type=` path; cover package-local bridge mappings with distinct value and mapping types. Do not absorb stale package negative expectations, the two external Rust-interop gate inputs, or any pre_v1 state.

yaseralnajjar:
Closes #3418.\n\nOpaque declarations with a distinct `structural=` mapping now resolve every declared Rust path but schedule the concrete opaque probe only for `type=`. Package-local and direct-backend probe planning share the same exact path-role selector.\n\nFocused validation includes the structural bridge runtime regression and the exact Pydantic package check.

### DIFF
--- a/crates/sifr_driver/src/build/rust_interop.rs
+++ b/crates/sifr_driver/src/build/rust_interop.rs
@@ -439,46 +439,14 @@ impl<'a> RustInteropResolver<'a> {
                     sysroot_trust.is_some(),
                 );
                 self.validate_backend_generated_bridge_imports(declaration, backend);
-                let signature =
-                    target_resolution::is_primary_target(&declaration.declaration, path)
-                        .then(|| {
-                            self.signature_contracts
-                                .get(&canonical_target_path)
-                                .cloned()
-                        })
-                        .flatten();
-                let async_thread_affinity = self.async_thread_affinity_for_probe(declaration);
-                let Some(sysroot_runtime_crate) = self.context.sysroot_runtime_crate.clone() else {
-                    self.push_diagnostic(
-                        declaration,
-                        path.span,
-                        DiagnosticCode::RUST_CARGO_METADATA,
-                        "Rust bridge probe requires a resolved Sifr sysroot runtime crate",
-                        vec![("target", path.dotted())],
-                        vec!["Direct Rust bridge probes must use the same resolved sysroot runtime crate as generated Cargo projects.".to_string()],
-                        None,
-                    );
+                if !self.plan_direct_backend_probe(
+                    declaration,
+                    path,
+                    backend,
+                    sysroot_trust.as_ref(),
+                ) {
                     return;
-                };
-                self.pending_direct_probes.push(PendingRustBridgeProbe {
-                    declaration: declaration.clone(),
-                    path: path.clone(),
-                    backend: backend.clone(),
-                    source_prefix: None,
-                    signature,
-                    async_thread_affinity,
-                    zero_copy_obligations: self
-                        .zero_copy_probe_obligations
-                        .get(&canonical_target_path)
-                        .copied()
-                        .unwrap_or((false, false)),
-                    trusted_sysroot: sysroot_trust.is_some(),
-                    sysroot_runtime_crate,
-                    sysroot_vendor_dir: sysroot_trust
-                        .as_ref()
-                        .map(|trust| trust.vendor_dir.clone()),
-                    cargo_resolution: self.cargo_resolution.clone(),
-                });
+                }
                 if let Some(trust) = &sysroot_trust {
                     resolved_sysroot_crate_root(&backend.dependency_name, backend, trust)
                         .unwrap_or_else(|| RustInteropResolvedRoot::DirectCargoDependency {
--- a/crates/sifr_driver/src/build/rust_interop/probe_planning.rs
+++ b/crates/sifr_driver/src/build/rust_interop/probe_planning.rs
@@ -53,6 +53,9 @@ impl RustInteropResolver<'_> {
         sysroot_trust: Option<&SysrootRustInteropTrust>,
     ) -> Option<String> {
         let dependency_name = package_bridge_dependency_name(package);
+        if !super::target_resolution::is_concrete_probe_path(&declaration.declaration, path) {
+            return Some(dependency_name);
+        }
         let signature = super::target_resolution::is_primary_target(&declaration.declaration, path)
             .then(|| {
                 self.signature_contracts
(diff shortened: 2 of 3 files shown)


yaseralnajjar:
# Verdict: SATISFIED

No blocking findings.

## Mechanism verification

**Selector correctness (`crates/sifr_driver/src/build/rust_interop/target_resolution.rs:26-41`)**
`is_concrete_probe_path` returns `true` unconditionally for non-`Opaque` kinds, so non-opaque probing is bit-identical to the parent. For `Opaque`, it matches by pointer identity against the `type=` argument's `RustInteropValue::TargetPath`.

The pointer-identity contract holds on both live call chains: `rust_interop.rs:306` iterates `declaration_paths(&declaration.declaration)`, which returns borrows *into that same declaration* (`target_resolution.rs:5-13`), and passes each borrow through `resolve_path` → planner → `is_concrete_probe_path(&declaration.declaration, path)`. No clone intervenes on either path; `plan_package_bridge_probe` / `plan_direct_backend_probe` are each called from exactly one site (`rust_interop.rs:345`, `rust_interop.rs:442`), and the `declaration.clone()` inside the planners happens only after the selector runs.

**Path-role identity is unambiguous.** Opaque declarations can never carry a positional `target` (`sifr_lowering/src/lower/rust_interop.rs:328-360` rejects positional args for `Opaque`), so `declaration.target` is always `None` and `type=` is the sole probe-eligible path. Duplicate `type=` keys are impossible (Python-level duplicate-keyword rejection), so the `.any()` cannot schedule two concrete probes.

**Equal-valued type/mapping.** `type=Value, structural=Value` yields two *distinct* `&RustTargetPath` referents (separate `arguments` elements), so `ptr::eq` selects only the first — exactly one concrete probe, no duplicate. Value-equality comparison (as used by `is_primary_target`, which compares span+segments) would have been wrong here; the pointer choice is the right one. The added unit test pins this, and the pre-existing end-to-end direct-backend test `package_rust_interop_same_path_mapping_failure_is_structured` (`rust_interop_opaque_contract_tests.rs:48`) still passes and still surfaces the `StructuralMapping` child diagnostic.

**Mapping obligation retained.** The surviving probe is built from `probe.path` = the `type=` path, and `opaque_probe_source` (`rust_interop_probe.rs:238-245`) independently reads the `structural` argument, emitting `type __SifrProbe = Value; type __SifrMapping = Mapping; __sifr_assert_structural_mapping::<__SifrMapping>()` with the bound `M: StructuralMapping<__SifrProbe>` — i.e. `Mapping: StructuralMapping<Value>`. `requires_structural_runtime` (`rust_interop_probe.rs:78-85`) is keyed off the declaration, not the path, so the probe manifest still enables the `structural` runtime feature. This is precisely the defect being fixed: previously the mapping path was probed as `__SifrProbe = Mapping` and demanded the nonsensical `Mapping: StructuralMapping<Mapping>` (fixture `structural_bridge_calls` declares `impl StructuralMapping<Token> for TokenMapping` only).

**Both planners use the same selector**, placed at the same position (first statement, before signature/affinity/sysroot resolution), so package-local bridge roots and direct Cargo backend roots behave identically.

**Non-probed structural paths still get full resolution.** The skip is confined to probe scheduling; `resolve_path` still runs root classification, sysroot-crate restriction, `validate_backend_trust`, `validate_backend_generated_bridge_imports`, generated-bridge-module registration (`uses_bridge_root`), and still pushes a `RustInteropResolvedTarget` for the mapping path. Package-bridge case correctly returns `Some(dependency_name)` (not `None`) so the resolved root is still built.

**Fail-closed preserved.** Missing/invalid `type=` is rejected in `parse_opaque_contract` (`opaque_contract.rs:155-160`, `86-92`), which runs at `rust_interop.rs:287-290` *before* the resolve loop, so the selector is never reached with an absent `type=`. No compatibility or fallback path was introduced.

**Refactor equivalence for the direct-backend extraction** is exact: same `canonical_sifr_target_path`, same argument order to `push_diagnostic`, same field values (`sysroot_trust.map(|t| t.vendor_dir.clone())` on `Option<&_>` ≡ the prior `as_ref().map(...)`), and `false` → caller `return` reproduces the prior early return with no resolved target pushed.

**Checks run:** `cargo test -p sifr_driver target_resolution::` (2 passed), `cargo test -p sifr_driver opaque` (21 passed, 2 pre-existing ignores), `cargo clippy -p sifr_driver --all-targets` (no warnings in touched files; the "items after a test module" warning is pre-existing in `rust_interop_probe_diagnostics.rs:174`), `cargo fmt --check` clean. All touched files well under 900 lines (837/183/168).

## Non-blocking observations

1. **Implicit pointer-identity contract (maintainability, `target_resolution.rs:37`).** `is_concrete_probe_path(&RustInteropDeclaration, &RustTargetPath)` has no type-level tie between the two arguments. If a future caller passes a path from one declaration and a *clone* of that declaration (e.g. `PendingRustBridgeProbe.declaration`), the selector silently returns `false` for every opaque path — a fail-open drop of the concrete probe with no diagnostic. Correct today; bounded hardening: have `declaration_paths` return `(PathRole, &RustTargetPath)` pairs and thread the role into the planners, replacing pointer identity with an explicit role.

2. **`clone=custom(path)` on opaque declarations is no longer probed.** Consistent with the stated requirement ("only the declared `type=` path"), and inconsequential in practice: the custom path is validated for shape in `clone_policy_value` (`opaque_contract.rs:218-223`) but is not carried into `OpaqueContract` or codegen. Previously it was probed as `type __SifrProbe = <fn path>;`, which would itself have been ill-formed. Out of scope for this item; record in the owning issue if `clone=custom` existence checking is wanted.

3. **Test altitude.** The two new tests exercise the helper only. Planner wiring for the *distinct* type/mapping case is covered solely by the E2E fixture `verification/areas/rust_interop/fixtures/structural_bridge_calls` (not run here). A cheap resolver-level assertion on `pending_direct_probes` length/path for a `type=A, structural=B` package-bridge declaration would pin the wiring in the fast suite.

4. The commit does not update the issue record; per `AGENTS.md` that is a post-merge step, not a defect in this diff.

yaseralnajjar:
Exact-SHA local gate evidence for `97a3383ef88a37683dadcf537c070703d5c511e0`:\n\n- Create-PR gate: ran once. All reached guards passed (HIR/file-size, demo freshness, dependency direction, ownership, sysroot resource certification, intrinsic/reachability/manifest/bootstrap guards, driver maintainability, verification self-tests, tiers, 40-row compatibility matrix, stale-draft and stable-claim checks). It stopped only at the excluded shared-state inputs: missing `negative/package_generated_type_import_rejected.sifr` and empty `examples/method_slot_runtime/src/schema_program.sifr`.\n- Merge gate: ran once on the same SHA with the same result and same two excluded inputs.\n\nNeither gate was rerun.

PR #3427, opened 2026-08-22
Item 1: migrate native hash regression vectors to bytes

yaseralnajjar:
## Scope

- replace stale native string hash calls with the canonical byte APIs
- assert complete SHA-256, MD5, and SHA-1 digest byte vectors
- keep the public Sifr hashlib contract unchanged

## Validation

- `CARGO_NET_OFFLINE=true cargo test --locked -p sifr_stdlib --all-features`
- `CARGO_NET_OFFLINE=true cargo test --locked -p sifr_driver stateless_crypto_codegen_tests`
- `cargo fmt --check`
- `git diff --check`

No compiler files changed, so the Sifr create-PR and merge gates do not apply.


yaseralnajjar:
Exact-SHA Claude Opus review

- Candidate: `07c213d8cf9b7aada0e55796469fb72195b5af39`
- Verdict: `SATISFIED`
- Blocking findings: none
- Follow-up: consider full known vectors for the remaining length-only SHA-224, SHA-384, SHA-512, BLAKE2b, and BLAKE2s assertions in a later item.

PR #3429, opened 2026-08-22
Item 2: use canonical performance LSP project roots

yaseralnajjar:
## Scope

- add an explicit project root to every LSP benchmark input
- validate `sifr.toml` and `src/main.sifr` from that root
- initialize the production LSP with that root
- migrate the single-file fixture to the canonical project layout
- remove the sibling-manifest mode and duplicate temporary package scaffold
- update the durable performance contract

## Validation

- performance manifest validation: pass (65 cases)
- performance runner self-test, including flat-layout rejection: pass
- all 18 LSP benchmark scenarios in smoke mode: pass
- TypeScript-Go transfer guardrails: pass
- JSON, Python compile, diff, and file-size checks: pass
- performance rules: manifest, runner, and budget variants pass; the two expected stale trend-hash variants remain assigned to Item 3

No compiler files changed, so Sifr create-PR and merge gates do not apply.


yaseralnajjar:
Exact-SHA Claude Opus review

- Candidate: `ddcecec119601724487adef2e1f888fc65a646de`
- Verdict: `SATISFIED`
- Blocking findings: none
- Follow-ups: add a self-test for a valid project root paired with a non-canonical source path; align explicit/default `src` notation between the two LSP fixtures; retain the current fail-closed spelling check or normalize only if manifest authoring later requires it.

PR #3438, opened 2026-08-22
Preserve open IOError dispatch

yaseralnajjar:
## Summary
- keep all named IOError handlers guarded and require a base or catch-all handler for open-domain coverage
- derive compiler IO subclass handling from the canonical type-system mapping
- refresh the 21 stale demo companions from the corrected compiler

## Validation
- focused IOError pass fixture checks, builds, and runs
- open-base fail fixture reports SIFR-RESULT-0005
- full sifr_lowering and sifr_codegen test suites
- affected-package clippy with warnings denied
- demo emitted-freshness, HIR maintainability, and file-size guardrails

### DIFF
--- a/crates/sifr/tests/e2e/fail/try_ioerror_subclasses_do_not_cover_base.sifr
+++ b/crates/sifr/tests/e2e/fail/try_ioerror_subclasses_do_not_cover_base.sifr
@@ -1,0 +1,22 @@
+# expect-error[col=5]: SIFR-RESULT-0005
+# Test: named IOError subclasses do not cover the open base IOError kind domain.
+
+def load_io() -> Result[str, IOError]:
+    raise IOError("other")
+
+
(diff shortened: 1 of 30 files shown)


yaseralnajjar:
Exact-SHA Opus review for `080fd152e55ca96a8caff682c2049d24ea37fe48`: **APPROVED**.

The review confirmed that Item 5A removes both closed-domain assumptions at their owners, preserves exact kind guards, requires base/catch-all coverage, consolidates the duplicate I/O subclass registries onto `IO_ERROR_KIND_CASES`, and provides direct/runtime evidence without a panic fallback.

Non-blocking notes: the full fail harness currently aborts first on the recorded Item 9 protocol mismatch, direct construction of named I/O subclasses is pre-existing/out of scope, and the remaining all-builtin registries have separate ownership purposes.

PR #3444, opened 2026-08-22
fix(lowering): return CFG invariant diagnostics

yaseralnajjar:
## Summary

- lower one-arm `match` control flow as a complete `goto` instead of an invalid one-target branch
- return CFG invariant failures through lowering as structured internal compiler diagnostics
- promote the minimized CFG crash regression and add exact negative-diagnostic coverage

## Validation

- `cargo test -p sifr_lowering cfg -- --nocapture`
- `cargo test -p sifr_lowering match_diagnostics -- --nocapture`
- `cargo clippy -p sifr_lowering -- -D warnings`
- direct `sifr check` for both non-exhaustive one-arm fixtures
- `cargo test -p sifr test_e2e_fail -- --nocapture` (CFG cases pass; stops at the separately owned Item 9 protocol diagnostic mismatch)
- `uv run --project verification --locked python verification/areas/regression/runner.py --suite fixedbugs --suite crashes`
- `cargo fmt --check`
- `git diff --check`
- first-party source file-size guardrail

### DIFF
--- a/crates/sifr_lowering/src/cfg.rs
+++ b/crates/sifr_lowering/src/cfg.rs
@@ -2,7 +2,8 @@
 
 use crate::{HirExpr, HirStmt};
 use sifr_ir::{
-    CfgBlock, CfgBlockId, CfgBlockLabel, CfgTerminator, ControlFlowGraph, FlowExitEffect, FlowFacts,
+    CfgBlock, CfgBlockId, CfgBlockLabel, CfgInvariantError, CfgTerminator, ControlFlowGraph,
+    FlowExitEffect, FlowFacts,
 };
 use sifr_type_system::Type;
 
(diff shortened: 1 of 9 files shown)


yaseralnajjar:
## Exact-SHA Opus review

Candidate: 7d51849ba8d9672792eaf8671f1377e4e81b74f4
Base: 882947d53e1b49b584973b4e6096b7d0153b79fa

Verdict: SATISFIED

Blocking findings:
- None

Follow-up findings:
- Pre-existing: guarded wildcard arms are incorrectly treated as exhaustive; this belongs to exhaustiveness guard handling, not CFG construction.
- Pre-existing: a two-arm int-or-None match can fail codegen with an unresolved generated union module.
- Suggestion: stop or deduplicate repeated structured CFG diagnostics after the first invariant failure.
- Suggestion: consider propagating CFG invariant errors through the two boolean/list helper callers instead of conservative fallbacks.
- Suggestion: the promoted fixed-bug manifest could additionally lock the diagnostic message; direct unit coverage already locks the exact code and count.

Claude also rechecked all CFG Branch construction sites, all Result call sites, the regression promotion schema, the focused tests, and valid-match smoke cases.

yaseralnajjar:
## Exact-SHA gate evidence

Candidate: 7d51849ba8d9672792eaf8671f1377e4e81b74f4

The create-PR gate and merge gate each ran exactly once on this candidate.

Both gates passed:
- Cargo cache setup
- lowering maintainability guardrail
- generated demo freshness
- source dependency direction and self-test
- submodule ownership and self-test
- sysroot resource certification and self-test
- stdlib intrinsic allowlist, native-adapter reachability, retained-manifest, and bootstrap-ordering guards and self-tests
- driver maintainability
- verification hardening and runner foundation self-tests

Both gates stopped at the same already-owned Rust-interop matrix inputs:
- Linked delivery A: missing negative/package_generated_type_import_rejected.sifr evidence source
- Item 10: examples/method_slot_runtime/src/schema_program.sifr still has an empty placeholder class body

These failures are unchanged external inputs, not Item 8 regressions. Per the phase record, Item 8 does not change or waive either later delivery. The gates were not rerun.

PR #3446, opened 2026-08-22
test(protocol): separate diagnostic fixture intent

yaseralnajjar:
## Summary

- make the forwarded-TypeVar fixture resolve its protocol before testing conformance
- assign the existing unresolved TypeVar bound fixture to the canonical name diagnostic
- strengthen focused lowering tests to require one exact diagnostic family

## Validation

- both focused lowering tests pass
- full sifr_lowering suite: 1,029 passed, 1 ignored
- broad fail suite: all 567 fixtures passed
- normal sifr_lowering Clippy with warnings denied
- cargo fmt --check
- git diff --check

This changes only tests and fixtures. No production compiler file changed, so the Sifr create-PR and merge gates do not apply.

### DIFF
--- a/crates/sifr/tests/e2e/fail/protocol_bound_unknown_forwarded_typevar.sifr
+++ b/crates/sifr/tests/e2e/fail/protocol_bound_unknown_forwarded_typevar.sifr
@@ -1,10 +1,14 @@
 # expect-error[col=12]: SIFR-PROTO-0001
-# Test: unknown protocol bounds are rejected when forwarding TypeVars through generic calls
+# Test: forwarded TypeVars must satisfy a resolved protocol bound
+
+class MissingBound(Protocol):
+    def required(self) -> int:
+        pass
 
 def take_missing[T: MissingBound](x: T) -> T:
     return x
 
-def relay_missing[U: MissingBound](x: U) -> U:
+def relay_missing[U](x: U) -> U:
     return take_missing(x)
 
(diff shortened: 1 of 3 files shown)


yaseralnajjar:
## Exact-SHA Opus review

Candidate: 5e7f78f40c87faca49589b0f062c47f9854091d8
Base: 3930d6e2616fefe6925fa6df4ed8d00acc6ce0a3

Verdict: SATISFIED

Blocking findings:
- None

Verified directly:
- the resolved forwarding fixture emits exactly one SIFR-PROTO-0001 at line 12, column 12
- the unresolved TypeVar-bound fixture emits exactly one SIFR-NAME-0003 at line 5, column 24
- the removed Readable/Closable body was duplicate coverage; the same mechanism remains covered in protocol_diagnostics.rs and protocol_bound_forwarding_non_conforming_typevar.sifr
- no production compiler file changed

Follow-up findings:
- Suggestion: rename protocol_bound_unknown_forwarded_typevar.sifr because the bound now resolves; a rename changes lexical fixture order and is outside this item.
- Suggestion: rename the focused unit test to distinguish an unbounded TypeVar from the separately covered incompatible-bound case.
- Pre-existing: test-target Clippy warnings remain outside this diff.

No remediation review was needed. Sifr create-PR and merge gates do not apply to this tests-and-fixtures-only candidate.

PR #3448, opened 2026-08-22
test(interop): align method-slot verification

yaseralnajjar:
## Summary

- distinguish compiler-required class-adapter marker bodies from real empty placeholder classes
- retain mutation coverage that rejects ordinary empty classes and add coverage that accepts required markers
- update the method-slot driver baseline to all five canonical output lines and final shared count 3

## Root cause

The matrix helper treated every line equal to pass as an empty class placeholder. A class-adapter marker is required by the compiler to be field-less and contain only pass, so the verification rule rejected a valid declaration. Method-slot runtime behavior was already canonical; only the driver expectation remained stale.

## Validation

- rust-interop fixture-matrix self-test: 237 cases passed
- focused ignored method-slot runtime test passed
- focused ignored lifetime/thread/shared-context rejection test passed
- full sifr_driver suite: 556 passed, 76 ignored
- full fixture matrix now stops only at linked delivery A's missing shared-bridge source; the method-slot row passes
- sifr_driver library Clippy passed with warnings denied
- Python syntax compilation for all changed checks
- cargo fmt --check
- git diff --check

A broad driver Clippy probe found one pre-existing Rust 1.94 lint in the unrelated diagnostic-rendering binary. It is outside this diff. No production compiler file changed, so Sifr create-PR and merge gates do not apply.

### DIFF
--- a/crates/sifr_driver/src/tests/package_rust_interop_method_slots.rs
+++ b/crates/sifr_driver/src/tests/package_rust_interop_method_slots.rs
@@ -31,7 +31,7 @@ fn test_method_slot_runtime() {
     );
     assert_eq!(
         run_built_package(&entrypoint),
-        "value-normalized\ninput-receiver\nvalue-no-context\nvalue-shared-2"
+        "value-normalized\ninput-receiver\ninput-serialized\nvalue-no-context\nvalue-shared-3"
     );
     let _ = std::fs::remove_dir_all(package_root);
 }
--- a/verification/areas/rust_interop/checks/_binding_helpers.py
+++ b/verification/areas/rust_interop/checks/_binding_helpers.py
@@ -99,5 +99,47 @@ def is_identifier_or_path_char(char: str) -> bool:
     return char.isalnum() or char in {"_", "."}
 
 
-def contains_empty_pass_body(text: str) -> bool:
-    return any(line.strip() == "pass" for line in text.splitlines())
+def contains_empty_placeholder_class(text: str) -> bool:
+    lines = text.splitlines()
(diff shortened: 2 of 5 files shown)


yaseralnajjar:
## Exact-SHA Opus review

Candidate: c049f2d6e65e76f33efc7f2c890e1df9fb545c0c
Base: bc4a77ae6cd4fe105c85cd5a184eb44a3ae263af

Verdict: SATISFIED

Blocking findings:
- None

Verified directly:
- the compiler requires @class_adapter_marker to be field-less and contain only pass
- SlotContract is therefore a required declaration, not a placeholder
- the fixture already emits normalize, receiver, serialize, no-context, and shared in that order
- context count reaches 3 before the shared-context read
- ordinary empty classes remain rejected while required marker classes are accepted
- every former helper call site uses the semantic placeholder classifier
- slot order and count logic is unchanged

Follow-up findings:
- Suggestion: deliberately decide whether the narrower class-only guard should also reject docstring-plus-pass or multiline class-header forms.
- Suggestion: consider normalizing whitespace before matching the class_adapter_marker decorator.
- Infrastructure: _scenario_checks.py and check_fixture_matrix.py are now 898 and 899 lines; the next change must split by responsibility.
- Pre-existing: the phase's original Item 10 wording describes the disproved fixture edit and must be corrected in the record-only update.
- Pre-existing: linked delivery A and the unrelated diagnostic-rendering Clippy lint remain outside this item.

No remediation review was needed. No production compiler file changed, so Sifr create-PR and merge gates do not apply.

PR #3450, opened 2026-08-22
fix(codegen): unify emitted builtin nominal ownership

yaseralnajjar:
## Summary

- exclude IOError kind aliases from direct project nominal collection
- register builtin definitions emitted by the shared synthetic prelude before relocating module code
- remove duplicate local builtin definitions from 41 generated demo companions

## Root causes

A direct Type::Class for an IOError kind entered the builtin nominal map even though code generation represents all six kinds through IOError and emits no subclass struct. Separately, builtin definitions emitted transitively in the shared synthetic prelude were not registered, so relocation retained duplicate local definitions.

## Validation

- six focused project nominal registry tests passed
- full sifr_codegen suite: 1,103 passed
- sifr_codegen Clippy with warnings denied
- demos/uuid_and_datetime and the UUID/date-time E2E fixture built and ran
- all generated demo companions are fresh
- every shared demo re-export resolves to exactly one struct, enum, or type definition
- no demo exports an IOError kind alias from the project nominal module
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/sifr_codegen/src/project_stdlib_nominals.rs
+++ b/crates/sifr_codegen/src/project_stdlib_nominals.rs
@@ -212,6 +212,7 @@ pub(crate) fn project_stdlib_nominal_plan(
     );
     let probe_name_refs = probe_names.iter().map(String::as_str).collect();
     let shared_source = strip_rust_items_by_name(&generated.rust_source, &probe_name_refs);
+    register_emitted_builtin_nominals(&shared_source, &mut registry);
     register_transitive_stdlib_nominals(&shared_source, stdlib_code, &mut registry);
     let crate_root_candidates = SHARED_MODULE_CRATE_ROOT_NOMINAL_IDENTITIES
         .iter()
@@ -321,6 +322,16 @@ fn register_transitive_stdlib_nominals(
     }
 }
 
+fn register_emitted_builtin_nominals(shared_source: &str, registry: &mut ProjectNominalRegistry) {
+    let defined_names = rust_source_defined_item_names(shared_source);
+    for name in crate::BUILTIN_ERROR_CLASSES {
+        if sifr_type_system::io_error_kind(name).is_some() || !defined_names.contains(*name) {
+            continue;
+        }
+        registry.register_shared((*name).to_string(), class_rust_name(None, name));
+    }
+}
+
 fn collect_function_nominals(
     function: &FunctionType,
     declarations: &mut BTreeMap<String, HashSet<String>>,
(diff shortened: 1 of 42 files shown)


yaseralnajjar:
## Exact-SHA Opus review

Candidate: c11ccebe428ee9ac2c1593d4f8e27ace485394eb
Base: 529573333780f24f4ef8ab1cf6072ab3db2f4316

Verdict: SATISFIED

Blocking findings:
- None

Verified directly:
- all 41 changed demos have exactly one shared definition for every project nominal re-export
- every removed local definition is re-exported from the shared module
- none of the six IOError kind aliases is defined or exported as a project nominal
- the direct Type::Class collection path is closed
- uuid_and_datetime has one ParseError definition and one canonical re-export
- no generated companion outside demos references the shared project nominal module

Follow-up findings:
- Infrastructure: three demos lose an unreachable local ValueError-to-Error conversion with their duplicate Error; broad E2E pass validation should prove no registered fixture depends on it.
- Suggestion: compare defined names with the computed Rust name in register_emitted_builtin_nominals to remove an implicit current-name coupling.
- Suggestion: add a project-codegen integration assertion for a transitively emitted builtin; the refreshed generated companions already lock this property.

No remediation review was needed.

yaseralnajjar:
## Broad E2E classification

The reviewed candidate ran the full E2E pass suite once: 679 fixtures passed and 9 reported build failures.

Exact-base reproduction at 529573333780f24f4ef8ab1cf6072ab3db2f4316 proved all three failure mechanisms are pre-existing:
- cpython_configparser_subset: E0382 partial move of parsed_option_pair reproduces on base
- imported_error_not_catch_all: E0317 try/except return fallthrough reproduces on base
- network_http_tls_loopback_split: E0382 moved server_config, client_tcp, and accepted values reproduce on base

The other six reports share those same generated mechanisms or batch groups. They are not caused by Item 10A and will be recorded as later phase work because no active issue owns them.

Item 10A's nominal ownership evidence remains valid on candidate c11ccebe428ee9ac2c1593d4f8e27ace485394eb. No remediation review is required.

yaseralnajjar:
## Exact-SHA gate evidence

Candidate: c11ccebe428ee9ac2c1593d4f8e27ace485394eb

The create-PR gate and merge gate each ran exactly once on this candidate. Both passed every prerequisite guard, including generated demo freshness and the 237-case Rust-interop matrix self-test. Both stopped only at linked delivery A's missing negative/package_generated_type_import_rejected.sifr source.

Item 10's prior method-slot matrix failure is closed. The gates were not rerun.

Additional E2E evidence: one full pass-suite run completed 679 fixtures and reported 9 build failures. Exact-base probes reproduced all three mechanisms on base SHA 529573333780f24f4ef8ab1cf6072ab3db2f4316. Those failures are not Item 10A regressions and will be recorded as later phase items.

PR #3452, opened 2026-08-22
fix(codegen): filter dead try continuation bindings

yaseralnajjar:
## Item 10B

Preserve ownership across generated try continuations by promoting only bindings referenced after the try statement.

Base: `cbf256719ff6dc1ed800c878ad50390c046fa400`
Candidate: `42cbe6ef798c8047b33289c9779dde94852e0046`

## Validation

- focused sequential try binding tests: 10 passed
- full sifr_codegen suite: 1,107 passed
- config-parser fixture family: builds
- cpython_configparser_subset: runs
- network_http_tls_loopback_split: builds and runs
- demo emitted freshness: pass
- sifr_codegen Clippy with warnings denied: pass
- format, HIR maintainability, file-size, and diff checks: pass

## Review remediation

The first exact-SHA Opus review found that assignment-target-only uses were absent from the liveness query. The query now covers scalar, field, subscript, attribute, nested-subscript, augmented, and tuple-rebind targets. Focused scalar and subscript regressions pass.

## Gate note

The one create-PR gate attempt on the pre-demo-refresh implementation SHA stopped at demo freshness. The 20 generated companions are now refreshed and independently pass freshness. A process exception is required before any second create-PR gate.

### DIFF
--- a/crates/sifr_codegen/src/class_emitter.rs
+++ b/crates/sifr_codegen/src/class_emitter.rs
@@ -424,8 +424,10 @@ impl RustEmitter {
         self.register_local_body_binding_types(&str_func.body);
 
         let mut body = Vec::new();
-        for stmt in &str_func.body {
-            let lowered = self.capture_structured_stmts(|inner| inner.emit_stmt(stmt));
+        for (stmt_index, stmt) in str_func.body.iter().enumerate() {
+            let lowered = self.capture_structured_stmts(|inner| {
+                inner.emit_stmt_with_following(stmt, Some(&str_func.body[stmt_index + 1..]));
+            });
             body.extend(lowered);
         }
 
--- a/crates/sifr_codegen/src/class_method_emitter.rs
+++ b/crates/sifr_codegen/src/class_method_emitter.rs
@@ -121,9 +121,18 @@ impl RustEmitter {
     pub(crate) fn lower_class_stmt_strict(
         &mut self,
         stmt: &HirStmt,
+        context: &str,
+    ) -> Vec<RustStmt> {
+        self.lower_class_stmt_strict_with_following(stmt, None, context)
+    }
+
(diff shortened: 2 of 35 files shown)


yaseralnajjar:
## Exact-SHA Opus remediation review

Candidate: `42cbe6ef798c8047b33289c9779dde94852e0046`
Base: `cbf256719ff6dc1ed800c878ad50390c046fa400`

**Verdict: SATISFIED**

Blocking findings: none.

The review confirmed that the remediation closes assignment-target liveness across scalar, field, subscript, attribute, nested-subscript, augmented, and tuple-rebind targets; block-tail threading is complete across the intended emission paths; and no clones were introduced.

Required later item from the second review: split “needs an enclosing declaration” from “needs a value returned” when a try-local was moved inside the try but is referenced later only as an assignment target. This mechanism reproduces on the exact base and is not an Item 10B regression. No third review will run.

Other follow-ups: prune now-test-only wrappers in later cleanup; note the conservative shared-query effect on with-target naming; cover nested-function default expressions in canonical traversal; retain the pre-existing project-union registry panic as a separate later item.

yaseralnajjar:
## Exact-SHA gate evidence

Candidate: `42cbe6ef798c8047b33289c9779dde94852e0046`

The authorized replacement create-PR gate and the single merge gate each ran once on this exact candidate.

Both gates passed:
- Cargo cache setup
- HIR maintainability
- generated demo freshness
- source-crate dependency direction
- submodule ownership
- sysroot resource certification
- stdlib intrinsic, adapter, manifest, and bootstrap guards
- driver maintainability
- verification hardening and runner self-tests

Both gates then stopped at the same previously assigned linked-delivery-A failure: the Rust-interop matrix requests `negative/package_generated_type_import_rejected.sifr` instead of the checked-in `negative/src/` path. All other Rust-interop suites and self-tests passed. Item 10B does not own that path correction.

PR #3454, opened 2026-08-22
fix(codegen): make raising try control flow total

yaseralnajjar:
## Summary
- classify always-raising `try` bodies with the canonical HIR control-flow effect
- emit an impossible-success arm and total result match instead of a one-sided `if let`
- add focused codegen coverage and refresh the affected generated demo

## Validation
- 1,108 `sifr_codegen` tests passed
- `imported_error_not_catch_all.sifr` builds and runs natively
- demo freshness, affected Clippy, formatting, HIR maintainability, file-size, and diff checks passed
- exact-SHA Opus review: SATISFIED
- create-PR and merge gates passed all preceding checks and stopped only at linked delivery A

### DIFF
--- a/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
@@ -243,3 +243,22 @@ def replace_first() -> Result[int, ProbeError]:
     assert!(generated.contains("let (mut data,) = match __sifr_try_res"));
     syn::parse_file(&generated).expect("subscript-live try binding Rust should parse");
 }
+
+#[test]
+fn always_raising_try_body_uses_total_result_match() {
+    let generated = generate_rust_from_source(&format!(
+        r#"{PROBE_ERROR}
+def classify() -> str:
+    try:
+        raise ProbeError("bad")
+    except ProbeError as error:
+        return error.message
+"#
+    ));
+
+    assert!(generated.contains("match __sifr_try_res"));
+    assert!(generated.contains("Ok(()) =>"));
+    assert!(generated.contains("sifr try/except raising body returned success"));
+    assert!(!generated.contains("if let Err(__sifr_try_err) = __sifr_try_res"));
+    syn::parse_file(&generated).expect("total raising try Rust should parse");
+}
--- a/crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs
+++ b/crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs
@@ -254,6 +254,10 @@ impl RustEmitter {
         // only returns from the closure-backed try body need a carrier.
         let capture_returns =
             self.current_return_type.is_some() && queries::body_contains_return(body);
+        let body_always_raises = matches!(
+            queries::block_control_flow_effect(body),
+            sifr_ir::HirControlFlowEffect::AlwaysRaises
+        );
         let direct_return_capture = capture_returns
             && queries::block_control_flow_effect(body).always_exits()
             && handlers
@@ -360,6 +364,12 @@ impl RustEmitter {
                         .collect(),
                 )],
             })));
+        } else if body_always_raises {
+            closure_body.push(RustStmt::Expr(RustExpr::FormatMacro {
+                name: "unreachable".to_string(),
+                format_str: "sifr try/except raising body fell through".to_string(),
+                args: vec![],
+            }));
         } else if !capture_returns {
             closure_body.push(RustStmt::Return(Some(RustExpr::FnCall {
                 func: Box::new(RustExpr::Ident("Ok".to_string())),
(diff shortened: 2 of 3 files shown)


yaseralnajjar:
SATISFIED

## Verdict rationale

The diff closes the one uncovered shape: a `try` body whose HIR control-flow effect is `AlwaysRaises` with **no** `return` in it, which meant `capture_returns == false` and therefore the old `if let Err(...)` statement form (`crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs:552`) plus an implicit `Ok(())` closure tail. In a value-returning function that fell through as `()`.

Mechanism check:

- **Branch reachability is consistent.** `successful_try_bindings` returns empty whenever the body always exits (`crates/sifr_codegen/src/stmt_support_emitter/try_error_helpers.rs:8`), so when `body_always_raises` holds the bindings list is always empty, `capture_returns` is false (an always-*raising* body has no `return`), `ok_ty` is `Unit`, and the new `Ok(())` pattern at `loops_try_finally.rs:526` matches the declared `Result<(), E>` exactly. The new closure-tail arm (`loops_try_finally.rs:367`) sits after both binding arms, so it can never shadow a binding-transport tail.
- **Rust typing is total.** Both arms diverge (`unreachable!` and `return`), so the `match` statement diverges and coerces in return position; a block ending in a diverging statement has type `!`, matching the pre-existing `direct_return_capture` idiom at `loops_try_finally.rs:378`.
- **Nominal dispatch preserved.** The new branch reuses `lower_try_except_handler_chain_for_ir` unchanged, so a union carrier still lowers to a per-variant `match` on the exact nominal variants (`try_handlers.rs:58-87`), and an unsupported handler set still bails via `Ok(None)` rather than emitting a wrong arm.
- **Async unaffected.** `closure_is_async` is derived after the tail is pushed; `unreachable!` contains no `await`, so async/non-async selection is unchanged, and the tail is valid inside an async body.
- **Fall-through handlers still work.** If a handler body does not exit, the `Err` arm yields `()` and the `Ok` arm still diverges, so void/continuing shapes keep their old semantics (visible in `demos/error_subclasses/emitted.rs:5430`).
- **Panic policy.** No `.unwrap()`/`.expect()`; `unreachable!` is used as a compiler invariant, consistent with the existing return-capture and try/finally emissions. The `AlwaysRaises` analysis is conservative in the safe direction: loops union with a fall-through `else` (`crates/sifr_ir/src/hir_flow.rs:155-169`), Python context managers force `falls_through` (`hir_flow.rs:195`), and nested `try/except` unions handler bodies — so the invariant is not data-dependent.
- **Regression coverage is adequate.** The new focused test pins the shape (total `match`, no `if let Err`), and the item's e2e acceptance fixtures already exercise it end-to-end at rustc level: `crates/sifr/tests/e2e/pass/imported_error_not_catch_all.sifr` (raise-only body, two nominal handlers that return) and `catches_same_basename_errors` / `nested_catches_same_basename_errors` in `crates/sifr/tests/e2e/pass/nominal_identity_alias_paths.sifr` (union carrier). No stale generated baseline remains: a scan of every `demos/*/emitted.rs` found no other `__sifr_try_res` closure with a `return Err(...)` followed by `Ok(())`.

No blocking findings.

## Non-blocking suggestions

1. `crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs:511-541` duplicates the trailing `else` branch; it differs only by the added `Ok(())` arm. Building `arms` conditionally (as the `capture_returns` path already does at :459-497) would remove the fourth copy of the handler-chain lowering and keep the file (838 lines) further from the 900-line cap.
2. The new test only asserts text plus `syn::parse_file`, which cannot catch E0308/E0317 — the exact defect class. The e2e fixtures carry the real proof, so consider an assertion that the emitted `classify` has no trailing implicit unit (or a note pointing at `imported_error_not_catch_all`) so the intent survives refactors of the emitter.
3. Residual, out of this item's mechanism: a conditional handler chain (IOError kind comparisons, `try_handlers.rs:23-38`) is still non-total, so an always-raising body whose only matching handler is kind-conditional would still fall through. No such program exists in the repo today (`raise FileNotFoundError` is unexercised), so this is latent rather than live.
4. Also pre-existing and untouched: a nested `try/finally` inside a non-capturing `try` body computes `can_return_error` from `try_closure_depth` (`loops_try_finally.rs:652`), which is only incremented for return-capturing closures. Inside a non-capturing try closure returning `Err` is legal, yet the emitter picks the `unreachable!("sifr try/finally error propagation in non-Result function")` arm — a reachable panic for `try: try: raise ...: finally: ...` in a non-`Result` function. Worth recording against a later item (10D/10F sibling), not this one.

yaseralnajjar:
Exact candidate: `752d9258c71414d4736c8e4ebb23fdf09886fd3d`

Gate evidence:
- `scripts/run_all_tests.sh --profile create-pr`: every preceding guardrail passed; stopped only at linked delivery A because the Rust-interop matrix names missing `negative/package_generated_type_import_rejected.sifr`.
- `scripts/run_all_tests.sh`: same result on the same SHA; every preceding guardrail passed and the same linked-owner path stopped the matrix.
- The matrix self-test and all nine other Rust-interop variants passed in both profiles.

Focused evidence:
- all 1,108 `sifr_codegen` tests passed
- `imported_error_not_catch_all.sifr` built and ran natively
- demo freshness, affected-crate Clippy, formatting, HIR maintainability, file-size, and diff checks passed

Base classifications:
- `try_union_error_alias`, `try_union_error_channel`, and `nominal_identity_alias_paths` stop earlier at the Item 10F-owned missing imported-union nominal path.
- workspace Clippy exposes the already-recorded unrelated `diagnostic_rendering_harness.rs` lint introduced outside this item.

PR #3456, opened 2026-08-22
Fix moved try binding declaration liveness

yaseralnajjar:
## Summary

- separate declaration liveness from value liveness after sequential try lowering
- emit a typed enclosing declaration without transporting a moved value
- reject value-dependent mutation of moved bindings during lowering
- add focused liveness, code-generation, lowering, and native-run coverage

## Candidate

- base: `65a350884388d76d807bbf5a03dfaeca16840a55`
- candidate: `f87b5f7e03c54d85b0a08b01ecc0130a16379bff`

## Validation

- full `sifr_codegen` and `sifr_lowering` test suites pass
- native moved-binding replacement fixture builds and runs
- generated Rust passes `rustc -D unused-mut`
- affected Clippy, formatting, demo freshness, HIR maintainability, file-size, and diff checks pass
- exact-SHA Opus review: SATISFIED
- create-PR and merge gates each passed preceding guardrails and stopped only at linked delivery A stale Rust-interop evidence path

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_moved_binding_reinitialized.sifr
+++ b/crates/sifr/tests/e2e/pass/try_moved_binding_reinitialized.sifr
@@ -1,0 +1,29 @@
+class ProbeError(Error):
+    message: str
+
+
+def load_text() -> Result[str, ProbeError]:
+    return "old"
+
+
+def consume(own value: str) -> None:
+    assert value == "old"
+
+
+def replace_text() -> Result[str, ProbeError]:
+    try:
+        text: str = load_text()
+        consume(text)
+    except ProbeError as error:
+        raise error
+    text = "new"
+    return text
+
+
+def main() -> None:
+    try:
+        result: str = replace_text()
+    except ProbeError:
+        assert False
+        return
+    assert result == "new"
--- a/crates/sifr_codegen/src/expr_render_helpers/field_and_stdlib_rewrites.rs
+++ b/crates/sifr_codegen/src/expr_render_helpers/field_and_stdlib_rewrites.rs
@@ -551,6 +551,19 @@ impl RustEmitter {
                     value,
                 }
             }
+            crate::RustStmt::LetDecl {
+                mutable,
+                name,
+                mut ty,
+            } => {
+                if is_plain_i64_storage_type(Some(&ty)) && self.is_forced_sifr_int_local(&name) {
+                    ty = crate::RustType::Named("SifrInt".to_string());
+                    self.sifr_int_local_bindings
+                        .borrow_mut()
+                        .insert(name.clone());
+                }
+                crate::RustStmt::LetDecl { mutable, name, ty }
+            }
             crate::RustStmt::LetPattern { pattern, value } => crate::RustStmt::LetPattern {
                 pattern,
                 value: self.rewrite_stdlib_constant_idents_in_expr(value),
--- a/crates/sifr_codegen/src/hir_analysis/queries.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries.rs
@@ -1,5 +1,7 @@
 mod queries_impl;
 pub(crate) use queries_impl::*;
+mod value_liveness;
+pub(crate) use value_liveness::*;
 #[cfg(test)]
 mod tests;
 #[cfg(test)]
--- a/crates/sifr_codegen/src/hir_analysis/queries/value_liveness.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries/value_liveness.rs
@@ -1,0 +1,603 @@
+use super::expr_references_var;
+use crate::hir_analysis::traversal::{self, TraversalConfig};
+use sifr_ir::{
+    HirAsyncWithKind, HirExpr, HirPattern, HirStmt, HirTupleTargetBinding, MutableArgumentTarget,
+};
+use sifr_type_system::ReceiverConvention;
+use std::cell::Cell;
+
+pub(crate) fn stmts_require_var_value_at_entry_including_nested_functions(
+    stmts: &[HirStmt],
+    var_name: &str,
+) -> bool {
+    block_requires_value(stmts, var_name, false)
+}
+
+pub(crate) fn declaration_only_binding_needs_mutability(stmts: &[HirStmt], var_name: &str) -> bool {
+    max_reassignments_on_path(stmts, var_name) > 1 || stmts_mutate_existing_value(stmts, var_name)
+}
+
+fn max_reassignments_on_path(stmts: &[HirStmt], var_name: &str) -> usize {
+    stmts.iter().fold(0, |total, stmt| {
+        (total + stmt_max_reassignments(stmt, var_name)).min(2)
+    })
+}
+
+fn stmt_max_reassignments(stmt: &HirStmt, var_name: &str) -> usize {
+    match stmt {
+        HirStmt::Let { name, .. }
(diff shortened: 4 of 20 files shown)


yaseralnajjar:
# SATISFIED

## Evidence

**Diff scope verified at exact SHAs** (`git rev-parse HEAD` = `f87b5f7e0`; base `65a350884`): 20 files, 829/21 lines. Nothing outside the item's semantic owners (codegen try lowering + liveness query, lowering ownership check, tests, fixture).

**Declaration liveness is distinct from value liveness.** `successful_try_binding_candidates` now produces the raw candidate set, and the two filters partition it (`crates/sifr_codegen/src/stmt_support_emitter/try_error_helpers.rs:4-40`): transported = `stmts_require_var_value_at_entry_including_nested_functions`; declaration-only = referenced **and not** value-required. The sets are disjoint by construction, so no name gets both a `LetDecl` and a tuple slot. The new backward value-liveness walk (`hir_analysis/queries/value_liveness.rs:223-425`) kills the incoming value only when every continuing path rebinds before reading — `if/elif/else` requires all arms (`:250-263`), missing `else` keeps the value live, `TupleUnpack`/`StarUnpack` rebinding is honored (`:298-322`), and nested-function bodies are shadow-checked (`:400-406`). Unit tests cover exactly these three cases (`:498-540`).

**Moved values are not transported.** Generated Rust for the new fixture (`cargo run -p sifr -- emit crates/sifr/tests/e2e/pass/try_moved_binding_reinitialized.sifr`):

```rust
fn replace_text() -> Result<String, ProbeError> {
    let text: String;
    let __sifr_try_res: Result<(), ProbeError> = (|| {
        let text: String = load_text()?;
        consume(text);
        Ok(())
    })();
    if let Err(__sifr_try_err) = __sifr_try_res { ... }
    text = "new".to_string();
    Ok(text)
}
```

The closure returns `Result<(), _>` — no `Ok((text,))`, no clone, no default, no `MaybeUninit`, no sentinel. Deferred initialization only.

**Scalar replacement gets a valid enclosing typed declaration.** `LetDecl { mutable, name, ty }` is a first-class IR node (`rust_ir.rs:92`), emitted before the try closure with the emitter's resolved type (`loops_try_finally.rs:266-272, 416-425`), rendered as `let [mut ]name: T;` (`render/render_core.rs:389-396`), and integrated into every consumer that must see the type: `ir_imports.rs:123` (import collection), `ir_validate.rs:125` (type validation), `expr_render_helpers/field_and_stdlib_rewrites.rs:554-566` (forced-`SifrInt` declaration type, mirroring the `Let` arm). Remaining match sites correctly treat it as expression-free/non-mutating (`ir_optimize/mutability_and_clone_rewrites.rs:105,397,579`, `string_key_loop_rewrite.rs:145`, `iterator_lowering.rs:630`, `simple_dispatch_and_bindings.rs:749`).

**Mutability is correct.** `declaration_only_binding_needs_mutability` (`value_liveness.rs:17`) sets `mut` only when a single path reassigns twice or the value is mutated in place; loop bodies count as 2 (`:50-79`), alternative branches take the max, not the sum (`:32-48`). This matches Rust's rules for deferred init: one assignment needs no `mut`, repeated assignment does. Verified by the emitted `let text: String;` (no `mut`) versus `let mut value: i64;` in `declaration_only_try_binding_is_mutable_for_repeated_replacement`.

**Value-dependent mutation is rejected before Rust generation.** `ensure_mutable_parameter_binding` now checks `ctx.scope.is_moved` first (`crates/sifr_lowering/src/lower/binding_mutability.rs:9-12`). Every value-dependent mutation target routes through it — field, nested field, subscript, nested subscript, attribute subscript (`lower/statements/control_flow.rs:62,92,141,186,220`), the four aug-assign roots (`lower/aug_assign_lowering.rs:99,152,229,287`), and tuple-unpack field targets (`lower/tuple_unpack.rs:52`) — so `values[0] = 2` after a move yields `OWN_USE_AFTER_MOVE` rather than reaching codegen. Full whole-name reassignment still clears the flag (`scope.rs:556` via `reset_moved_with_flow`), which is why the fixture's `text = "new"` is accepted.

**No compatibility path or fallback added.** Grepping the added lines for `MaybeUninit|Default::default|unwrap(|expect(|clone()|fallback|compat` finds only compiler-internal `String` clones and `expect` in test bodies. `local_binding_types.get(..).unwrap_or(ty)` reuses the pre-existing pattern from `successful_bindings`, not a new fallback.

**Tests / gates run on the candidate:**
- `cargo test -p sifr_codegen` — 1116 passed, 0 failed (includes 13 `sequential_try_binding` and 6 `value_liveness` tests)
- `cargo test -p sifr_lowering` — 1030 passed, 0 failed (includes `moved_try_binding_rejects_value_dependent_subscript_assignment`)
- `cargo test -p sifr -- --skip test_e2e_pass` — all suites ok, 0 failed
- `cargo run -p sifr -- run crates/sifr/tests/e2e/pass/try_moved_binding_reinitialized.sifr` — native build and assertions pass
- `scripts/check_hir_maintainability_guardrails.py` — PASS; largest touched file 877 lines, new `value_liveness.rs` 603 lines, both under 900
- `cargo clippy -p sifr_codegen -p sifr_lowering --all-targets` — no warning anchored to an added line (the `statement_diagnostics_tests.rs` hits are at lines 15–30/154/186, all pre-existing and outside the added block at 287+)

The renamed test `try_binding_replaced_before_read_keeps_only_its_declaration` correctly inverts the Item 10B conservative expectation, which is the intended behavior change for this item.

## Follow-up findings (non-blocking, not in this item's scope)

- **suggestion / overlaps Item 10E** — `crates/sifr_codegen/src/hir_analysis/queries/queries_impl.rs:88` is shadow-blind for nested-function parameters, while `value_liveness.rs:400-406` is shadow-aware. A post-`try` nested function whose parameter shadows the binding therefore lands in *neither* set boundary cleanly: it is "referenced" but not value-required, so a `LetDecl` is emitted that is never assigned or read. Effect is a dead declaration (`unused_variables` warning in the generated crate), not invalid Rust. Item 10E owns the traversal-side reference rules.
- **pre-existing issue** — the `mut`-removal pass only rewrites `RustStmt::Let` (`ir_optimize/mutability_and_clone_rewrites.rs:66-79`), so a conservatively-`mut` `LetDecl` is never demoted. Warning-level only; the HIR-side computation above is already tight, so no case was observed.
- **pre-existing issue** — a binding moved in the try body and only *partially* reassigned before a later read (`if c: text = "new"` then `print(text)`) is still classified value-required and transported, exactly as at the base SHA. Unchanged by this diff; belongs with the remaining continuation-ownership work, not Item 10D's field/subscript criterion.

yaseralnajjar:
Exact-SHA gate evidence for candidate `f87b5f7e03c54d85b0a08b01ecc0130a16379bff`:

- `scripts/run_all_tests.sh --profile create-pr` ran once. It passed Cargo setup, HIR maintainability, file-size, demo freshness, dependency direction, ownership, sysroot certification, stdlib registry/manifest/bootstrap guardrails, driver maintainability, and verification self-tests. It stopped only at linked delivery A: missing Rust-interop evidence source `negative/package_generated_type_import_rejected.sifr`.
- `scripts/run_all_tests.sh` ran once on the same SHA. It passed the same preceding checks and stopped at the same linked delivery A row.
- Neither gate was repeated.

Focused evidence: 1,116 codegen tests and 1,030 lowering tests passed; the native fixture built and ran; affected Clippy, formatting, HIR maintainability, demo freshness, file-size, and diff checks passed.

PR #3458, opened 2026-08-22
Close pre-v1 nested-default traversal regressions

yaseralnajjar:
## Summary

- traverse nested function defaults in their defining scope while keeping body descent configurable
- preserve try-bound values referenced only by nested defaults and suppress parameter-shadow false positives
- route declared nested-function calls through canonical argument resolution, including keyword/default handling
- demote declared-function provenance on scalar and tuple rebinding
- add structured codegen, lowering, query, liveness, traversal, and native regression coverage

## Exact candidate

- Base: 218ba6029ff3de4ab75803925e15536c5bdeed0b
- Candidate: 76a19d209d875906b0d7c9751d7216b206a58976

## Review

- First exact-SHA Opus review on a7e0a8e4a0809d01445a37cbaebbd73c742e31f5: BLOCKED on keyword/default resolution and tuple-unpack provenance.
- One permitted remediation exact-SHA Opus review on 76a19d209d875906b0d7c9751d7216b206a58976: SATISFIED.
- A new non-blocking risk involving the unscoped name-keyed function registry will be recorded as a later phase item. No third review was run.

## Validation

Targeted validation passed before the final commit:
- sifr_codegen: 1121 passed
- sifr_lowering: 1032 passed, 1 ignored
- native nested-default fixture built and ran
- affected clippy passed with warnings denied
- generated demo freshness passed
- HIR maintainability guardrail passed
- file-size guardrail passed for 3217 files
- formatting and diff checks passed

Exact-SHA gates were each run once on 76a19d209d875906b0d7c9751d7216b206a58976:
- create-PR gate: all preceding checks passed; stopped only on linked delivery A stale Rust-interop evidence path negative/package_generated_type_import_rejected.sifr
- merge gate: same bounded external failure after all preceding checks passed

The stale Rust-interop fixture is outside Item 10E and remains owned by the linked delivery.

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_nested_function_default_binding.sifr
+++ b/crates/sifr/tests/e2e/pass/try_nested_function_default_binding.sifr
@@ -1,0 +1,27 @@
+class ProbeError(Error):
+    message: str
+
+
+def load_value() -> Result[int, ProbeError]:
+    return 42
+
+
+def read_after_try() -> Result[int, ProbeError]:
+    try:
+        value: int = load_value()
+    except ProbeError as error:
+        raise error
+
+    def read_value(candidate: int = 1) -> int:
+        return candidate
+
+    return value + read_value() + read_value(candidate=5)
+
+
+def main() -> None:
+    try:
+        result: int = read_after_try()
+    except ProbeError:
+        assert False
+        return
+    assert result == 48
--- a/crates/sifr_codegen/src/function_emitter/scope_and_function_types.rs
+++ b/crates/sifr_codegen/src/function_emitter/scope_and_function_types.rs
@@ -387,10 +387,7 @@ impl RustEmitter {
         if func.method_kind != sifr_ir::MethodKind::Regular
             || !func.decorators.is_empty()
             || !func.type_params.is_empty()
-            || func
-                .params
-                .iter()
-                .any(|param| param.default.is_some() || param.keyword_only)
+            || func.params.iter().any(|param| param.keyword_only)
         {
             return false;
         }
--- a/crates/sifr_codegen/src/hir_analysis/queries.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries.rs
@@ -3,6 +3,8 @@ pub(crate) use queries_impl::*;
 mod value_liveness;
 pub(crate) use value_liveness::*;
 #[cfg(test)]
+mod default_expression_tests;
+#[cfg(test)]
 mod tests;
 #[cfg(test)]
 mod type_and_operator_tests;
--- a/crates/sifr_codegen/src/hir_analysis/queries/default_expression_tests.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries/default_expression_tests.rs
@@ -1,0 +1,51 @@
+use super::{stmts_reference_var, stmts_reference_var_including_nested_functions};
+use sifr_ir::{HirExpr, HirFunction, HirParam, HirStmt, MethodKind};
+use sifr_type_system::{ParamConvention, Type};
+
+#[test]
+fn nested_function_defaults_use_the_defining_scope_and_parameters_shadow_the_body() {
+    let nested = HirFunction {
+        name: "inner".to_string(),
+        params: vec![HirParam {
+            name: "value".to_string(),
+            ty: Type::Int,
+            default: Some(HirExpr::Name {
+                name: "default_source".to_string(),
+                binding_id: None,
+                ty: Type::Int,
+            }),
+            keyword_only: false,
+            convention: ParamConvention::own(),
+        }],
+        return_type: Type::Int,
+        body: vec![HirStmt::Return {
+            value: Some(HirExpr::Name {
+                name: "value".to_string(),
+                binding_id: None,
+                ty: Type::Int,
+            }),
+        }],
+        is_async: false,
+        method_kind: MethodKind::Regular,
+        receiver: None,
+        decorators: vec![],
+        rust_interop: Vec::new(),
+        python_interop: Vec::new(),
+        compiler_intrinsic: None,
+        type_params: vec![],
+    };
+    let stmts = vec![HirStmt::NestedFunction {
+        func: nested,
+        move_captures: false,
+        capture_clones: Vec::new(),
+    }];
+
+    assert!(stmts_reference_var(&stmts, "default_source"));
+    assert!(stmts_reference_var_including_nested_functions(
+        &stmts,
+        "default_source"
+    ));
+    assert!(!stmts_reference_var_including_nested_functions(
+        &stmts, "value"
+    ));
+}
--- a/crates/sifr_codegen/src/hir_analysis/queries/queries_impl.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries/queries_impl.rs
@@ -98,6 +98,16 @@ fn stmts_reference_var_with_config(
     config: TraversalConfig,
 ) -> bool {
     let mut on_stmt = |stmt: &HirStmt| {
+        if config.descend_nested_functions {
+            if let HirStmt::NestedFunction { func, .. } = stmt {
+                let parameter_shadows_var = func.params.iter().any(|param| param.name == var_name);
+                if !parameter_shadows_var
+                    && stmts_reference_var_with_config(&func.body, var_name, config)
+                {
+                    return TraversalControl::Stop;
+                }
+            }
+        }
         let target_references_var = match stmt {
             HirStmt::Assign { name, .. } | HirStmt::AugAssign { name, .. } => name == var_name,
             HirStmt::FieldAssign { object, .. }
@@ -133,7 +143,12 @@ fn stmts_reference_var_with_config(
         TraversalControl::Continue
     };
     matches!(
-        traversal::walk_stmts_until(stmts, config, &mut on_stmt, &mut on_expr,),
+        traversal::walk_stmts_until(
+            stmts,
+            TraversalConfig::LOCAL_SCOPE_ONLY,
+            &mut on_stmt,
+            &mut on_expr,
+        ),
         TraversalControl::Stop
     )
 }
--- a/crates/sifr_codegen/src/hir_analysis/queries/value_liveness.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries/value_liveness.rs
@@ -408,9 +408,16 @@ fn stmt_requires_value(stmt: &HirStmt, var_name: &str, live_after: bool) -> bool
                 || live_after
         }
         HirStmt::NestedFunction { func, .. } => {
+            let defaults_reference_value = func.params.iter().any(|param| {
+                param
+                    .default
+                    .as_ref()
+                    .is_some_and(|default| expr_references_var(default, var_name))
+            });
             let parameter_shadows_value = func.params.iter().any(|param| param.name == var_name);
-            (!parameter_shadows_value
-                && super::stmts_reference_var_including_nested_functions(&func.body, var_name))
+            defaults_reference_value
+                || (!parameter_shadows_value
+                    && super::stmts_reference_var_including_nested_functions(&func.body, var_name))
                 || live_after
         }
         HirStmt::Match { subject, arms, .. } => {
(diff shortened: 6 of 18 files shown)


yaseralnajjar:
## BLOCKED

### What I verified (static review of `218ba60..a7e0a8e`, no files modified, no gates re-run)

**Canonical traversal**
- `crates/sifr_codegen/src/hir_analysis/traversal/traversal_impl.rs:631-641` — `HirStmt::NestedFunction` now walks every `param.default` through `walk_expr_until` *before* the config-gated body descent, so defaults are visited in the defining scope under both configs; body descent still keys off `descend_nested_functions` only. `walk_stmt`/`walk_stmts`/`walk_stmts_until` all delegate to `walk_stmt_until`, so there is one implementation, and the doc contract at lines 32-38 was updated to match. `traversal/tests.rs:174-205` pins both configs (`["default_only"]` vs `["default_only", "nested_only"]`).
- `HirExpr::Lambda` (`traversal_impl.rs:223`) still walks only the body, but both lambda lowering paths hard-code `default: None` (`lower/expressions/methods_lambdas_and_comprehensions.rs:38,86`), so there is no reachable gap — not an omission.

**Shadowing and liveness**
- `queries/queries_impl.rs:99-152`: body descent moved into the query with a `parameter_shadows_var` guard, walker pinned to `LOCAL_SCOPE_ONLY`; defaults are therefore still seen when a parameter shadows the name, bodies are not. Recursion passes `config`, so deeper nesting behaves the same.
- `queries/value_liveness.rs:408-420`: `defaults_reference_value` is OR-ed ahead of the shadow-suppressed body check — correct, since defaults evaluate in the defining scope. Covered by `value_liveness.rs:611-641`, `queries/default_expression_tests.rs`, and codegen tests `post_try_hir_default_reference_keeps_the_binding_live` / `shadowing_nested_parameter_does_not_keep_try_binding_live` (`lib_codegen_tests/sequential_try_binding_codegen_tests.rs:216-299`).

**Supported defaults / no new fallback**
- `collect_param_default` (`lower/default_args.rs:28-52`) still restricts defaults to `lower_expr_simple` (literals only; `simple_expr.rs:17-57`), so name/call defaults keep emitting `TYPE_UNSUPPORTED_DEFAULT_ARGUMENT`. The Name-default liveness case is exercised by synthesizing HIR, which is the right shape for a canonical-contract test.
- Structured codegen: the `param.default.is_some()` bail-outs were removed in `lower_stmt/simple_dispatch_and_bindings.rs:535` and `function_emitter/scope_and_function_types.rs:387-390`; since lowering inlines defaults at the call site, ignoring `param.default` during closure emission is consistent, and `nested_function_with_literal_default_uses_structured_codegen` pins `|candidate: i64|` + `read_value(42_i64)`. No compatibility or fallback branch added.
- `BindingKind` consumers: only `lower/method_receiver_places.rs:382-398` matches exhaustively and the new `Function` arm returns `InvalidPlace::Unsupported` (conservative). `is_inferred_local_binding` (`scope.rs:110`) now excludes function names, which only narrows optional-widening at `tuple_unpack.rs:167` / `control_flow.rs:332` — safe direction.
- File-size guardrail holds: largest touched file is 875 lines; `scope.rs` is 731 after the `#[path = "scope_tests.rs"]` split, a pattern already used in-repo (`sifr/src/self_update_metadata.rs:868`).

### Blocking finding

**regression — `crates/sifr_lowering/src/lower/expressions/regular_calls.rs:129-142`** (criterion: "local function default filling applies only to declared function bindings"; AGENTS: no user-triggerable wrong behavior, solve root causes / reuse the canonical mechanism)

The new fill loop inspects only `call.arguments.args` and never `call.arguments.keywords`, yet it satisfies the arity check at line 143 and returns a `HirExpr::Call` at line 228 — after which no stage still has the AST.

Failure scenario:
```python
def outer() -> int:
    def read(candidate: int = 1) -> int:
        return candidate
    return read(candidate=5)
```
- Base: `args.len()==0 != 1` → `CALL_NOT_CALLABLE_OR_ARITY` ("callable 'read' expects 1 argument(s), got 0"). No wrong binary.
- Candidate: the default `1` is filled, arity passes, `candidate=5` is silently discarded → `read(1)` is emitted and `outer()` returns `1`. A previously diagnosed shape now compiles to the wrong value.

The keyword-aware canonical filler already exists and is what the module-level path uses: `lower_function_call_args` (`lower/method_call_args.rs:170-260`) resolves positional + keyword + defaults and emits `CALL_UNEXPECTED_KEYWORD` / `CALL_MISSING_REQUIRED_ARGUMENT` / duplicate-argument diagnostics (see `expressions_tests/ownership_and_async.rs:803` for the module-level contract).

Correction: route the `Callable`/`AsyncCallable` declared-function branch through `lower_function_call_args` with `ctx.function_defaults`, or — minimally — reject a non-empty `call.arguments.keywords` before filling defaults, plus a lowering test for the keyword shape.

### Follow-up findings

- *plausible defect in the same new mechanism* — `crates/sifr_lowering/src/lower/tuple_unpack.rs:155-185` rebinds an existing name without calling `invalidate_rebound_binding_facts`, so `Scope::mark_rebound_local` (`scope.rs:458-470`) never runs and the binding keeps `BindingKind::Function`. A function name rebound via tuple unpack would keep filling the *old* declaration's default (`local, n = (helper, 2)` then `local()`), i.e. the same silent-wrong-call class the simple-assignment path now guards against. Worth fixing in the same batch as the blocking finding; I could not confirm reachability without building.
- *scope observation, not blocking* — Item 10E's written scope is traversal only, but the candidate also enables nested-function default filling at call sites (`define_function` + `regular_calls.rs`, new `BindingKind::Function`, rebind invalidation). The review criteria you supplied cover it and the e2e fixture `crates/sifr/tests/e2e/pass/try_nested_function_default_binding.sifr` depends on it, so I treat it as authorized; consider recording the widened scope in the Item 10E record.
- *pre-existing* — the `Callable`-binding branch is keyword-blind in general (e.g. `read(2, candidate=5)` already dropped the keyword before this candidate). Fixing the blocking finding via `lower_function_call_args` closes this too.

Validation note: per the phase-closure rules and this session's read-only constraint I did not re-run the focused suites or the native build; the SATISFIED-relevant behavior above is established by inspection of the diff and surrounding code, and the blocking finding is derived from the same reading rather than an executed repro.

yaseralnajjar:
# SATISFIED

## What I verified (static exact-SHA review, no edits, no builds)

**Canonical resolver now drives declared function calls** — `crates/sifr_lowering/src/lower/expressions/regular_calls.rs:122-174`. The scope-callable branch splits: a `BindingKind::Function` binding resolves its signature from `ctx.functions`, its defaults from `ctx.function_defaults`, its vararg index from `ctx.vararg_functions`, and calls `lower_function_call_args` — the same entry point the top-level declared-call path uses at `regular_calls.rs:326-328, 395-402`. That resolver supplies positional/keyword/default resolution, `CALL_DUPLICATE_ARGUMENT`, `CALL_MISSING_REQUIRED_ARGUMENT`, `CALL_WRONG_POSITIONAL_COUNT`, and `CALL_UNEXPECTED_KEYWORD` (`method_call_args.rs:170-263, 638-678`), including the vararg variant (`method_call_args.rs:479-560`). The a7e0a8 defect (positional-only default filling that dropped keywords) is gone: the ad-hoc `for index in args.len()..param_types.len()` fill no longer exists.

**Explicit keyword values are preserved** — `method_call_args.rs:242-245` prefers the keyword value over the default. Evidence: e2e fixture `crates/sifr/tests/e2e/pass/try_nested_function_default_binding.sifr:18,27` (`42 + 1 + 5 == 48`) and codegen assertions `read_value(42_i64)` **and** `read_value(5_i64)` in `sequential_try_binding_codegen_tests.rs:216-229`.

**Ordinary `Callable` bindings gain nothing** — `Function` provenance is set only by `Scope::define_function` (`scope.rs:210`), called only from `register_local_function_signature` (`signatures_and_effects.rs:614-615`). Parameters, imported functions, and assigned callables keep `Parameter`/`Local`, so they take the else-branch at `regular_calls.rs:145-172`, which is byte-for-byte the base behavior (strict positional arity, `CALL_NOT_CALLABLE_OR_ARITY`, no defaults).

**Rebinding removes provenance, both shapes** — `Scope::mark_rebound_local` demotes `Function → Local` in the frame *and* in the retained binding fact (`scope.rs:458-469`, no-op for non-`Function` kinds). It is reached from the scalar path via `invalidate_rebound_binding_facts` (`patterns_and_assignments.rs:312-313`, used by simple assign `control_flow.rs:372` and chained assign `:583,:615`) and from tuple unpack (`tuple_unpack.rs:180`). Both covered by `callable_and_builtin_diagnostics.rs:25-63`. Other rebinding shapes (ann-assign `patterns_and_assignments.rs:488-494`, for/with/except/comprehension) call `define_*`, which inserts a fresh binding with a non-`Function` kind, so provenance cannot survive them.

**Ranges, ownership, conventions, async** — args are now param-indexed, and `call_argument_ranges_by_param` (`call_argument_ranges.rs:6-30`) yields a same-length param-indexed range vector, so `conventions.get(i)`, `consume_owned_value`, `validate_borrowed_structural_coercion`, and `validate_regular_call_arguments` all stay aligned; the old `ranges.resize(...)` hack is removed. `is_async_callable` → `coroutine_result_type` is untouched (`regular_calls.rs:236-240`).

**Traversal/shadowing contract (original 10E) still intact** — defaults are walked in the defining scope unconditionally (`traversal_impl.rs:629-636`), body descent stays config-gated, `queries_impl.rs:98-151` handles nested bodies explicitly with a parameter-shadow guard while walking with `LOCAL_SCOPE_ONLY`, and `value_liveness.rs:408-421` keeps a try-bound value live when only a default references it. Tests: `queries/default_expression_tests.rs`, `traversal/tests.rs:135-206`, `value_liveness.rs:611-641`, `sequential_try_binding_codegen_tests.rs:231-300`. Codegen simple-nested-function gating no longer excludes defaults (`simple_dispatch_and_bindings.rs:532-535`, `scope_and_function_types.rs:387-390`), which is sound because no emitter materializes `param.default` — defaults are lowering-resolved at every call site.

**Guardrail** — all touched files ≤ 892 lines (`scope.rs` 731 after tests moved to `scope_tests.rs`); `scripts/check_hir_maintainability_guardrails.py` PASS.

## Non-blocking findings (later-item, per the phase rule — not a third round)

1. **Name-keyed `ctx.functions` now feeds nested-call resolution** — `regular_calls.rs:123-143`. The signature comes from `ctx.functions[name]`, while `param_types`/`conventions` come from the scope binding, and the declared branch no longer cross-checks `args.len()` against `param_types.len()` (the `zip` at `:176` truncates silently). Because `ctx.functions.insert` (`signatures_and_effects.rs:616`) is unscoped and never restored on frame pop, two same-named local defs in sibling block scopes can leave the scope Callable and the registry signature disagreeing — yielding a spurious missing/extra-argument diagnostic or an unchecked argument. The unscoped registry is pre-existing (base already resolved top-level calls through it at `regular_calls.rs:310`); this item newly routes nested calls through it. I did not build, so this is reasoned from the diff, not reproduced.
2. **Optional widening on function-name rebinding** — `Function` kind is excluded from `VarInfo::is_inferred_local_binding` (`scope.rs:110-113`), and `can_widen` is read (`control_flow.rs:332`, `tuple_unpack.rs:167`) before `mark_rebound_local` demotes the kind. So `local = None` after `def local(...)` now reports a type mismatch instead of widening to `Optional[...]`. Stricter, and arguably correct, but a behavior change worth recording.
3. **Vararg range mapping** — for vararg calls, positional arguments past the vararg index map onto the vararg param slot in `call_argument_ranges_by_param`, so a per-element type diagnostic anchors to the first vararg element rather than the offending one. Cosmetic; no worse than the base `resize` behavior.

yaseralnajjar:
Exact-SHA gate evidence for 76a19d209d875906b0d7c9751d7216b206a58976:

- scripts/run_all_tests.sh --profile create-pr: all setup and guardrail steps passed, including HIR maintainability, 3217-file size guardrail, demo freshness, dependency direction, ownership, stdlib certification, driver maintainability, and verification-runner self-tests. It stopped only in rust_interop matrix because shared_bridge_crate references missing evidence source negative/package_generated_type_import_rejected.sifr. Wall time 103.70s; no swaps.
- scripts/run_all_tests.sh: the same steps passed and it stopped only on the same linked-delivery-A Rust-interop path. Wall time 104.61s; no swaps.

Each gate was invoked exactly once. The failure is pre-existing, external to Item 10E, and owned by plans/issues/active/ad-hoc-rust-interop-fixture-matrix-repair.md.

PR #3460, opened 2026-08-22
Fix imported union nominal paths

yaseralnajjar:
## Summary

- classify compiler-owned errors by canonical identity instead of basename alone
- preserve distinct project registry paths for sifr.csv.Error and sifr.configparser.Error
- add focused registry tests and a native same-basename imported-union fixture
- keep strict missing-path behavior; no basename lookup, duplicate nominal, or fallback was added

## Exact candidate

- Base: 0b45970f5980f94225f2a9e95813deb433960952
- Candidate: a26b21d10e9da58efce486ce1e2e09297cb9e6ce

## Review

The one exact-SHA Opus review returned SATISFIED. It confirmed that the original project-codegen panic is closed and classified the later ChildFormatter-to-Formatter rustc failure as a separate stdlib-parent upcast mechanism. That failure will be recorded as Item 10K.

## Validation

- 1123 sifr_codegen tests passed
- 8 focused project nominal tests passed
- the imported_union_nominal_paths fixture built and ran natively
- nominal_identity_alias_paths completes project code generation without the imported-union panic
- affected Clippy, formatting, HIR maintainability, and the 3218-file size guardrail passed

Each exact-SHA gate was run once. Both passed all preceding checks and stopped only on linked delivery A stale Rust-interop evidence path negative/package_generated_type_import_rejected.sifr.

### DIFF
--- a/crates/sifr/tests/e2e/pass/imported_union_nominal_paths.sifr
+++ b/crates/sifr/tests/e2e/pass/imported_union_nominal_paths.sifr
@@ -1,0 +1,17 @@
+from sifr.configparser import Error as ConfigError
+from sifr.csv import Error as CsvError
+
+
+def error_message(value: CsvError | ConfigError) -> str:
+    match value:
+        case CsvError(message=message):
+            return "csv:" + message
+        case ConfigError(message=message):
+            return "config:" + message
+
+
+def main() -> None:
+    csv_error: CsvError | ConfigError = CsvError("one")
+    config_error: CsvError | ConfigError = ConfigError("two")
+    assert error_message(csv_error) == "csv:one"
+    assert error_message(config_error) == "config:two"
--- a/crates/sifr_codegen/src/project_stdlib_nominals.rs
+++ b/crates/sifr_codegen/src/project_stdlib_nominals.rs
@@ -398,7 +398,7 @@ fn collect_shared_nominals(
             methods,
             ..
         } => {
-            if crate::BUILTIN_ERROR_CLASSES.contains(&name.as_str())
+            if is_compiler_builtin_error(identity.as_deref(), name)
                 && sifr_type_system::io_error_kind(name).is_none()
             {
                 builtin_types
@@ -494,7 +494,7 @@ fn collect_nominal_identity(
     let Some((module, name)) = identity.rsplit_once('.') else {
         return;
     };
-    if crate::BUILTIN_ERROR_CLASSES.contains(&name) {
+    if is_compiler_builtin_error(Some(identity), name) {
         return;
     }
     declarations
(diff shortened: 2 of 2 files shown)


yaseralnajjar:
Exact-SHA gate evidence for a26b21d10e9da58efce486ce1e2e09297cb9e6ce:

- scripts/run_all_tests.sh --profile create-pr: all setup and guardrail steps passed, including HIR maintainability, 3218-file size guardrail, demo freshness, dependency direction, ownership, stdlib certification, driver maintainability, and verification-runner self-tests. It stopped only because shared_bridge_crate references missing evidence source negative/package_generated_type_import_rejected.sifr. Wall time 98.58s; no swaps.
- scripts/run_all_tests.sh: the same checks passed and it stopped only on the same linked-delivery-A Rust-interop path. Wall time 102.00s; no swaps.

Each gate was invoked exactly once. The failure is external to Item 10F and remains owned by plans/issues/active/ad-hoc-rust-interop-fixture-matrix-repair.md.

yaseralnajjar:
# SATISFIED

## Diff under review

`0b45970f5..a26b21d10` — 2 files, +103/−2:
- `crates/sifr_codegen/src/project_stdlib_nominals.rs` (2 call-site changes, 1 new predicate, 2 new unit tests)
- `crates/sifr/tests/e2e/pass/imported_union_nominal_paths.sifr` (new native fixture)

## Root cause: verified and exact

The panic owner is `RustEmitter::project_nominal_path` (`crates/sifr_codegen/src/union_type_helpers.rs:21`), which keys strictly on the identity (`key = identity.unwrap_or(name)`, line 14) and panics when the project registry has no entry.

Two identity-blind basename tests fed it:
- `collect_nominal_identity` (`project_stdlib_nominals.rs:497`) returned early for any `BUILTIN_ERROR_CLASSES` **basename**, so `sifr.csv.Error` / `sifr.configparser.Error` never produced a `declarations` entry.
- `collect_shared_nominals` (`:401`) instead funnelled them into `builtin_types`, a `BTreeMap` keyed by **name** with `or_insert_with` (`:404-406`) — so both identities collapsed onto one `"Error"` slot, the first-seen identity got `crate::__sifr_project_nominals::Error`, and the second identity got no path at all → panic.

`stdlib/sifr/csv.sifr:12` and `stdlib/sifr/configparser.sifr:7` are the only checked-stdlib declarations whose basename collides with `BUILTIN_ERROR_CLASSES` other than `sifr.parallel.WorkerError` / `WorkerRuntimeError`, so the blast radius of the predicate change is exactly the intended pair. The fix (`:506-512`) narrows the test to identity-less classes, the `sifr.builtin.` compiler namespace (no such stdlib module exists — it is synthetic, produced only by `sifr_lowering/src/lower/async_with.rs:49` and `.../signatures_and_effects.rs:104,121`), and `GLOBAL_RUST_NOMINAL_IDENTITIES` (`sifr_type_system/src/types/source_names.rs:53`), which is what keeps `sifr.parallel.WorkerError`/`WorkerRuntimeError` on the global prelude.

This is a semantic-owner correction, not a fallback: no basename lookup was added anywhere, and `project_nominal_path` still panics on a genuinely unregistered identity.

## Contract checks

| Check | Evidence |
|---|---|
| Distinct canonical paths | `cargo run -p sifr -- emit imported_union_nominal_paths.sifr`: exactly one `pub struct __SifrStdlib_sifr_x2econfigparser_x2eError` (line 2564) and one `__SifrStdlib_sifr_x2ecsv_x2eError` (line 2585), both inside `mod __sifr_project_nominals` (opens line 19), each re-exported once at crate root (lines 2693-2694). Union variants reference those paths. |
| No duplicate local nominal | Only one definition site per struct in the whole emitted crate; the csv/configparser module bodies are stripped by `relocate_project_stdlib_nominals` (`:65-97`). |
| Compiler builtins intact | `pub use __sifr_project_nominals::Error / IOError / ParseError` still emitted; `builtin_error_union_keeps_its_shared_module_definition` and `emitted_builtin_nominals_join_the_project_registry` pass unchanged. |
| `sifr.builtin` + global Rust identities | `is_compiler_builtin_error` clauses cover both; `parallel_map_worker_panic_typed` and `parallel_map_basic` run green (exit 0). |
| Direct `IOError`-kind handling | `direct_io_error_kind_type_does_not_create_a_project_nominal` and `io_error_kind_handlers_do_not_create_dangling_project_nominals` pass; the `sifr.builtin.` clause is what preserves them. |
| Registry/emission ownership | `register_transitive_stdlib_nominals` (`:305`) and `register_emitted_builtin_nominals` (`:325`) untouched; crate-root partition (`:217-240`) untouched; no snapshot baseline references these paths. |

## Validation run (targeted only; no create-PR or merge gate)

- `cargo test -p sifr_codegen --lib` — 1123 passed, 0 failed (both new tests green).
- `cargo test -p sifr_frontend --lib` — 122 passed.
- New fixture `imported_union_nominal_paths.sifr` builds and runs natively, exit 0.
- Regression sweep of every csv/configparser/parallel consumer fixture — 13 fixtures, all exit 0 (`cpython_csv_subset`, `cpython_configparser_subset`, `csv_reader_file`, `imported_error_not_catch_all`, `parsers_and_encoders`, `stdlib_csv_consolidated`, `stdlib_configparser`, `structured_data_formats`, `try_union_builtin_error_channel`, `try_union_error_alias`, `try_union_error_channel`, `parallel_map_basic`, `parallel_map_worker_panic_typed`).
- `cargo clippy -p sifr_codegen --all-targets` — no warning in `project_stdlib_nominals.rs`; all 16 reported sites are in files this diff does not touch. `cargo fmt --check` clean.
- `scripts/check_file_size_guardrails.py` — PASS (3218 files, limit 900); the touched file is 810 lines. `check_hir_maintainability_guardrails.py` — PASS.

Both new tests are genuine regression guards: under the base predicate, `same_basename_stdlib_errors_keep_distinct_nominal_declarations` sees empty `declarations`, and `same_basename_stdlib_errors_get_distinct_project_paths` loses the second identity to `or_insert_with` on the `"Error"` key.

## Findings

1. **Out of Item 10F scope — `nominal_identity_alias_paths` residual failure (classified, must be recorded).** The fixture now completes project code generation with no panic (acceptance criterion 1 met) and the emitted union members carry distinct `__SifrStdlib_sifr_x2ecsv_x2eError` / `..._configparser_x2eError` paths. It now fails in `rustc`: `error[E0277]: the trait bound '__SifrStdlib_sifr_x2elogging_x2eFormatter: From<ChildFormatter>' is not satisfied`, emitted at the `formatter_template(own formatter: LogFormatter)` call in `main` (`crates/sifr/tests/e2e/pass/nominal_identity_alias_paths.sifr:96`). That is a subclass-to-parent by-value conversion for a user class inheriting a checked-stdlib class — `Into::<Parent>::into(child)` is emitted without a corresponding `From` impl. It has no relationship to imported-union nominal path registration and is correctly not repaired here. Per AGENTS.md it needs an owning record: the Item 10F record commit should add an ownership-transfer row and a new item (next free id, `10K`) for "stdlib-parent upcast conversion", alongside the standard Item 10F record block.

2. **Test-coverage, non-blocking — the fabricated `StdlibCode` in `same_basename_stdlib_errors_get_distinct_project_paths` is inert.** `project_stdlib_nominals.rs:559-577` builds two `StdlibRustSource` entries, but both asserted `rust_paths` entries are produced unconditionally by the `declarations` loop at `:150-157`; nothing in the test depends on the generated source. The test would pass identically with `StdlibCode::default()`, so it pins registry keys only and cannot detect a registry-vs-emission divergence. Adding `assert!(plan.prelude.contains(&format!("pub struct {}", class_rust_name(Some("sifr.csv.Error"), "Error"))))` for both identities — the pattern already used at `:621` — would close that gap cheaply.

3. **Coverage note, low severity.** In the currently-green suite, same-basename union coverage is now: construction + `match` (new fixture) and `except` dispatch through an alias error channel (`try_union_error_alias.sifr`). Structural equality and `set`/hash over a `CsvError | ConfigError` value exist only in `nominal_identity_alias_paths.sifr:105-111`, which still fails for the unrelated reason in finding 1, so those two behaviors remain unguarded until that owner lands.

No blocking defect found. The scope-bounded predicate change, the registry outcome, and the native evidence all match Item 10F's acceptance criteria.

PR #3462, opened 2026-08-22
Propagate unmatched conditional try errors

yaseralnajjar:
## Summary

- preserve unmatched conditional-handler errors through an enclosing try or compatible Result channel
- retain SIFR-RESULT-0005 when no checked channel accepts the error
- emit explicit residual Err returns for conditional, branchless, single, and union carrier members
- isolate lowering try state at every function boundary
- add lowering, codegen, and native regression coverage

## Exact candidate

- Base: d25f897d9ccd1b90fd8c9cda6e2221fd15413df3
- Candidate: 00e67dc56ce27ec6ce2a64ac91216b19afbf87b3

## Review

- First exact-SHA Opus review on f2e7df701: BLOCKED on branchless carrier members and nested-function lowering state leakage.
- One permitted remediation exact-SHA review on 00e67dc56ce27ec6ce2a64ac91216b19afbf87b3: SATISFIED.
- Per the phase rule, two new mechanisms from the second review will be recorded as later Items 10L and 10M. No third review ran.

## Validation

- lowering: 1035 passed, 1 ignored
- codegen: 1126 passed
- existing IOError subclass fixture and expanded residual propagation fixture ran natively
- affected Clippy, formatting, HIR maintainability, and the 3220-file size guardrail passed

Each exact-SHA gate ran once. Both passed all preceding checks and stopped only on linked delivery A stale Rust-interop evidence path negative/package_generated_type_import_rejected.sifr.

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_unmatched_ioerror_propagation.sifr
+++ b/crates/sifr/tests/e2e/pass/try_unmatched_ioerror_propagation.sifr
@@ -1,0 +1,50 @@
+def bubble_other() -> Result[str, IOError]:
+    try:
+        raise IOError("result other")
+    except FileNotFoundError:
+        return "missing"
+
+
+def catch_outer() -> str:
+    try:
+        try:
+            raise IOError("outer other")
+        except FileNotFoundError:
+            return "missing"
+    except IOError as error:
+        return error.message
+
+
+def bubble_union(flag: bool) -> Result[str, IOError | ValueError]:
+    try:
+        if flag:
+            raise IOError("union other")
+        raise ValueError("union value")
+    except FileNotFoundError:
+        return "missing"
+
+
+def main() -> None:
+    try:
+        _: str = bubble_other()
+        assert False
+    except IOError as error:
+        assert error.message == "result other"
+
+    assert catch_outer() == "outer other"
+
+    try:
+        _: str = bubble_union(True)
+        assert False
+    except IOError as error:
+        assert error.message == "union other"
+    except ValueError:
+        assert False
+
+    try:
+        _: str = bubble_union(False)
+        assert False
+    except IOError:
+        assert False
+    except ValueError as error:
+        assert error.message == "union value"
--- a/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
@@ -400,3 +400,57 @@ def classify() -> str:
     assert!(!generated.contains("if let Err(__sifr_try_err) = __sifr_try_res"));
     syn::parse_file(&generated).expect("total raising try Rust should parse");
 }
+
+#[test]
+fn unmatched_ioerror_kind_returns_through_the_checked_error_channel() {
+    let generated = generate_rust_from_source(
+        r#"
+def classify() -> Result[str, IOError]:
+    try:
+        raise IOError("other")
+    except FileNotFoundError:
+        return "missing"
+"#,
+    );
+
+    assert!(generated.contains("__sifr_try_err.kind == \"FileNotFound\""));
+    assert!(generated.contains("return Err(__sifr_try_err);"));
+    syn::parse_file(&generated).expect("residual IOError propagation Rust should parse");
+}
+
+#[test]
+fn branchless_handler_member_returns_through_the_checked_error_channel() {
+    let generated = generate_rust_from_source(
+        r#"
+def classify() -> Result[str, IOError]:
+    try:
+        raise IOError("other")
+    except ValueError:
+        return "value"
+"#,
+    );
+
+    assert!(generated.contains("Err(__sifr_try_err) => {\n            return Err(__sifr_try_err);"));
+    assert!(!generated.contains("structured statement emission missing"));
+    syn::parse_file(&generated).expect("branchless residual Rust should parse");
+}
+
+#[test]
+fn union_member_without_a_handler_keeps_its_residual() {
+    let generated = generate_rust_from_source(
+        r#"
+def classify(flag: bool) -> Result[str, IOError | ValueError]:
+    try:
+        if flag:
+            raise IOError("other")
+        raise ValueError("value")
+    except FileNotFoundError:
+        return "missing"
+"#,
+    );
+
+    assert!(generated.contains("__sifr_try_variant_error.kind == \"FileNotFound\""));
+    assert!(generated.matches("return Err(").count() >= 2);
+    assert!(!generated.contains("structured statement emission missing"));
+    syn::parse_file(&generated).expect("union residual Rust should parse");
+}
--- a/crates/sifr_codegen/src/stmt_support_emitter/try_handlers.rs
+++ b/crates/sifr_codegen/src/stmt_support_emitter/try_handlers.rs
@@ -184,10 +184,23 @@ impl RustEmitter {
             branches.push((cond_expr, handler_body));
         }
 
+        let has_unconditional_branch = branches.iter().any(|(cond, _)| cond.is_none());
+        let mut current_else = if has_unconditional_branch {
+            None
+        } else {
+            let Some(source_error_type) = source_error_type else {
+                return Ok(None);
+            };
+            let Some(residual) =
+                self.unmatched_try_error_return_for_ir(err_ident, source_error_type)
+            else {
+                return Ok(None);
+            };
+            Some(vec![residual])
+        };
         if branches.is_empty() {
-            return Ok(None);
+            return Ok(current_else);
         }
-        let mut current_else: Option<Vec<RustStmt>> = None;
         for (cond, body) in branches.into_iter().rev() {
             if let Some(cond) = cond {
                 current_else = Some(vec![RustStmt::If {
@@ -200,5 +213,29 @@ impl RustEmitter {
             }
         }
         Ok(Some(current_else.unwrap_or_default()))
+    }
+
+    fn unmatched_try_error_return_for_ir(
+        &self,
+        err_ident: &str,
+        source_error_type: &Type,
+    ) -> Option<RustStmt> {
+        let target_error_type = self
+            .try_closure_error_type_info
+            .last()
+            .and_then(Clone::clone)
+            .or_else(
+                || match self.current_return_type.as_ref()?.resolve_alias() {
+                    Type::Result(_, error_type) => Some(error_type.as_ref().clone()),
+                    _ => None,
+                },
+            )?;
+        let error = RustExpr::Ident(err_ident.to_string());
+        let converted =
+            self.consuming_value_conversion_for_ir(&target_error_type, source_error_type, error);
+        Some(RustStmt::Return(Some(RustExpr::FnCall {
+            func: Box::new(RustExpr::Ident("Err".to_string())),
+            args: vec![converted],
+        })))
     }
 }
--- a/crates/sifr_lowering/src/lower/mod.rs
+++ b/crates/sifr_lowering/src/lower/mod.rs
@@ -190,6 +190,7 @@ mod task_join_set_calls;
 mod task_owner_scope_state;
 mod task_scope_calls;
 mod task_scope_offload_calls;
+mod try_error_propagation;
 mod tuple_unpack;
 #[cfg(test)]
 mod type_alias_tests;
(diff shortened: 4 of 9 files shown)


yaseralnajjar:
# BLOCKED

## What the diff does (verified)

**Lowering** — `crates/sifr_lowering/src/lower/try_error_propagation.rs:1-54`, called from `statement_dispatch.rs:598-606` after `ctx.try_block_error_types` has already been swapped back to the enclosing try's set (`statement_dispatch.rs:491-493`), so `extend` (`try_error_propagation.rs:33`) correctly deposits unmatched errors on the *enclosing* try and they reach that try's `body_error_types`/carrier. Acceptance via the function channel uses `func_type.return_type.resolve_alias()` → `Type::Result(_, e)` + `is_assignable_to` (`:37-47`); otherwise `SIFR-RESULT-0005` is re-emitted unchanged (`:52-53`).

**Codegen** — `try_handlers.rs:190-203` gives a chain with no unconditional branch an explicit residual, and `unmatched_try_error_return_for_ir` (`:218-240`) targets `try_closure_error_type_info.last()` else the function's `Result` error type. The stack is popped before the handler chain is lowered (`loops_try_finally.rs:340-341`), so the residual correctly returns through the *enclosing* closure's channel, and the union path re-wraps per member (`try_handlers.rs:58-88`).

**Empirically confirmed working** (candidate SHA, `target/debug/sifr`, temp dir `/tmp/g10g`): single carrier + Result channel; nested try closure (`catch_outer`, residual returns through the outer closure's `Result<String, IOError>`); double propagation inner→outer→function channel; binding-transport shape (residual inside the `let (value,) = match …` Err arm); async fn; converted channel (`return Err(::std::convert::Into::<Error>::into(__sifr_try_err))` with the `From<IOError>` impl emitted); union carrier when every member has a branch (variant re-wrap). All six built with rustc and ran `exit=0`. Catch-all/base handler suppresses the residual (`except IOError` after `except FileNotFoundError` → 0 residual returns, `has_unconditional_branch` at `try_handlers.rs:190`). Kind/nominal matching in `try_except_handler_condition_expr` is untouched; no panic, fallback, or synthetic catch-all is inserted; ownership is sound (handlers `clone`, residual moves in the disjoint `else`). Rejection is retained (`p_nonresult` → `SIFR-RESULT-0005`). `cargo test -p sifr_lowering` 1034 pass, `-p sifr_codegen` 1124 pass, `cargo test -p sifr -- --skip test_e2e_pass` all pass (all four `try_*uncovered*` fail fixtures still fire). File-size guardrail OK (`statement_dispatch.rs` 871, `try_handlers.rs` 241, new module 54).

## Blocking findings

**1. A carrier (or union member) with no representable handler branch gets no residual — the compiler now accepts the program and emits `compile_error!` instead of propagating or diagnosing.** `try_handlers.rs:187-189` returns `Ok(None)` on `branches.is_empty()` *before* the new residual logic, and in the union path a single branchless member aborts the whole chain (`try_handlers.rs:63-71`). The lowering relaxation removed the guard that previously kept these shapes out of codegen.

- `def pick(flag: bool) -> Result[str, IOError | ValueError]` raising `IOError`/`ValueError` with only `except FileNotFoundError` — the canonical 10G shape generalized to a union: `sifr check` → "no errors found"; `sifr emit` → `compile_error!("structured statement emission missing for production path: TryExcept { … }")`; `sifr build` → `SIFR-BUILD-0005`. The `IOError` member's residual is destroyed too, because the branchless `ValueError` member bails the whole union chain.
- Minimal single-carrier repro: `def f() -> Result[str, IOError]: try: raise IOError("x") / except ValueError: return "value"` — same `compile_error!`.
- Base SHA `d25f897d9` rejected both with `SIFR-RESULT-0005` (the old hunk diagnosed whenever `uncovered` was non-empty). So this is newly admitted by this commit. It violates Item 10G scope ("Give a conditional handler chain one explicit residual-error outcome. Route the unmatched value through the checked error channel"), its acceptance criterion ("propagates **or** produces a structured compile-time diagnostic"), and governing rule 8.

**2. `enclosing_try_accepts_errors` leaks across a nested `def` boundary, dropping `SIFR-RESULT-0005` for a function that cannot propagate.** `statement_dispatch.rs:602` passes `prev_in_try`, and `ctx.in_try_block` is never reset when lowering a nested `FunctionDef` body, so `try_error_propagation.rs:32-35` records the nested function's unmatched error on the *outer* function's try.

- `def outer() -> str:` containing `def inner() -> str:` with `try: raise IOError("leak") / except FileNotFoundError: return "missing"`, wrapped in `try: … except IOError as error:` — `sifr check` reports only a FLOW-0901 warning. `inner` returns plain `str` and has no channel for the `IOError`, yet it is accepted; base rejected it. This contradicts the item's scope line "Reject an unhandled error when the enclosing function cannot propagate it." (When the outer try does *not* cover the type the diagnostic reappears, but attributed to the outer `try:` range rather than the nested one.)

## Non-blocking notes

- Focused coverage is thin relative to the mechanism: one codegen test (`sequential_try_binding_codegen_tests.rs:404-419`, simplest single-carrier case) and two lowering accept-tests. There is no unit or fixture coverage for the union carrier, nested-closure channel selection, converted channel, async, or binding transport — precisely where finding 1 lives. The native fixture covers only single carrier + one nested-try shape.
- Diagnostic records still list the owner as `sifr_lowering::lower::statements` (`internal_docs/diagnostic_codes.md:237`, generated `docs/errors/SIFR-RESULT-0005.mdx`); emission moved to `lower::try_error_propagation`. The docs also still describe the rule unconditionally, without the new Result-channel/enclosing-try exemption.
- Two build failures I reproduced are **pre-existing, not 10G**: nested try/except inside `try/finally` in a non-`Result` function, and a nested `def` inside a `try` — both fail with `E0308` at the candidate SHA using base-legal, totally-handled equivalents (return capture is not `Ok`-wrapped for the enclosing closure). These belong to Item 10H / the try-closure return-capture owner.
- Cosmetic: kind comparisons render as `"FileNotFound".to_string().to_string()` (pre-existing double `to_string`).

yaseralnajjar:
Exact-SHA gate evidence for 00e67dc56ce27ec6ce2a64ac91216b19afbf87b3:

- scripts/run_all_tests.sh --profile create-pr: all setup and guardrail steps passed, including HIR maintainability, 3220-file size guardrail, demo freshness, dependency direction, ownership, stdlib certification, driver maintainability, and verification-runner self-tests. It stopped only because shared_bridge_crate references missing evidence source negative/package_generated_type_import_rejected.sifr. Wall time 106.59s; no swaps.
- scripts/run_all_tests.sh: the same checks passed and it stopped only on the same linked-delivery-A Rust-interop path. Wall time 105.23s; no swaps.

Each gate was invoked exactly once. The failure is external to Item 10G and remains owned by plans/issues/active/ad-hoc-rust-interop-fixture-matrix-repair.md.

yaseralnajjar:
# SATISFIED

## Blocker 1 — residual before `None` (closed)

`crates/sifr_codegen/src/stmt_support_emitter/try_handlers.rs:187-203`: the residual is now built **before** the empty-branch bail-out, and `branches.is_empty()` returns `current_else` (the residual) instead of `Ok(None)`. `unmatched_try_error_return_for_ir` (`try_handlers.rs:218-240`) selects the channel as enclosing-try-closure carrier → else function `Result` error type, and converts via the shared `consuming_value_conversion_for_ir` (`class_upcasts.rs:35-46`, union-member wrap at `class_upcasts.rs:172-190`), so no new conversion path was invented.

- Single carrier, conditional chain: `try_handlers.rs:187` + test `unmatched_ioerror_kind_returns_through_the_checked_error_channel` (`sequential_try_binding_codegen_tests.rs:405`).
- Branchless carrier member (handler dropped as `Unsupported`): test at `sequential_try_binding_codegen_tests.rs:421` asserts `Err(__sifr_try_err) => { return Err(...) }` and absence of `structured statement emission missing`.
- Union carrier: per-member call at `try_handlers.rs:63-71` no longer collapses the whole match to `None`; test `union_member_without_a_handler_keeps_its_residual` (`sequential_try_binding_codegen_tests.rs:438`) plus native `bubble_union` in `crates/sifr/tests/e2e/pass/try_unmatched_ioerror_propagation.sifr:19-24,36-50` distinguishing `IOError` vs `ValueError` at runtime.
- Handler chains are emitted in the enclosing scope (`loops_try_finally.rs:446-578`), while the closure error stack is pushed only around the body (`loops_try_finally.rs:331-341`), so the residual `return` lands in the same frame whose error type it converts to — this holds for the binding-transport arm (`:447`), the return-capturing arm (`:499`), the always-raises arm (`:532`), the plain arm (`:563`), and the async closure variant (`closure_is_async`, `:405-413`).
- Ownership: conditional branches clone the error (`try_handlers.rs:109-113`); only the exclusive innermost `else` (or the union match arm binding) moves it. No `unwrap`/`expect`/panic added; the residual is a plain `return Err(...)`.
- Catch-all preserved: `has_unconditional_branch` (`try_handlers.rs:187`) suppresses the residual whenever a source-level `except Error`/bare `except`/exact-nominal handler exists; kind matching (`try_handlers.rs:19-40`) is unchanged.

## Blocker 2 — `in_try_block` leak across function boundaries (closed)

`crates/sifr_lowering/src/lower/statements/function_body.rs:9-14` replaces `in_try_block` with `false` and takes `try_block_error_types` for the whole nested body, restoring both afterwards. Every nested-function body path goes through `lower_function_stmts` (`statement_dispatch.rs:783`, `annotations_and_function_lowering.rs:279`, `class_body_lowering.rs:181,326,612`); no other caller lowers a function body via `lower_stmts` directly. Diagnostic evidence: `statement_diagnostics_tests.rs:191-238` (Result channel accepts, outer try accepts, nested function does **not** leak into the outer try).

## SIFR-RESULT-0005 preserved

`try_error_propagation.rs:31-53` only suppresses the diagnostic when an enclosing try accepts the error (`prev_in_try`, passed at `statement_dispatch.rs:600-608`, after the outer set is restored at `:493` so unmatched members join the outer carrier) or when the function's resolved return type is `Result` with an assignable error. Otherwise `uncovered_try_errors` still fires (`result_diagnostics.rs:55`, `SIFR-RESULT-0005` at `codes/registry.rs:186`); pre-existing test `ioerror_subclass_handlers_do_not_cover_the_open_base_error` (`statement_diagnostics_tests.rs:158`) still passes.

## Validation run locally (no create-PR/merge gate)

- `cargo test -p sifr_lowering`: 1035 passed, 1 ignored, 0 failed.
- `cargo test -p sifr_codegen`: 1126 passed, 0 failed.
- `cargo run -p sifr -- run crates/sifr/tests/e2e/pass/try_unmatched_ioerror_propagation.sifr`: exit 0 (all asserts pass — Result channel, outer try, both union members).
- `cargo fmt --check` clean; `cargo clippy -p sifr_lowering -p sifr_codegen -- -D warnings` clean (the only `--all-targets` warnings are pre-existing test-helper lints, e.g. `statement_diagnostics_tests.rs:15,18` — untouched lines).
- File-size guardrail: all touched files well under 900 lines (max `statement_dispatch.rs` 871; new `try_error_propagation.rs` 54).

## Findings (later-item, per the phase rule — no third review requested)

1. **PLAUSIBLE — a user-defined parent handler over a child error now propagates instead of failing the build.** `try_except_handler_condition_expr` (`try_handlers.rs:41-48`) matches only exact nominal identity, while lowering treats a parent handler as covering a child (`type_rendering.rs:596-608` parent-chain assignability), so for `class Base(Error)` / `class Child(Base)`, `raise Child` + `except Base` in a `Result[str, Base]` function the handler is dropped as `Unsupported`, `branches` is empty, and the new residual emits `return Err(child→Base)` — the handler body never runs. Before this change the same shape declined structured emission and produced `compile_error!` (`structured_stmt_entrypoints.rs:24`), i.e. a hard build failure rather than wrong runtime behavior. Item 10G's scope fixes exact nominal + `IOError` kind matching, so the matching gap itself is outside it, but the residual now masks it. No e2e fixture exercises a two-level user error hierarchy with `except` (scanned all 693 `pass` fixtures — none).
2. **PLAUSIBLE — codegen does not reset try-closure error-channel state at a nested-function boundary.** `unmatched_try_error_return_for_ir` (`try_handlers.rs:223-232`) reads `try_closure_error_type_info.last()`, but the nested-function emitter saves and restores only `current_return_type` (`function_emitter/scope_and_function_types.rs:472,493,560`) — `try_closure_error_type_info` / `try_closure_error_type` / `try_closure_depth` stay from the enclosing try closure. Lowering now correctly routes a nested function's unmatched error to that function's own `Result` channel, so a `def inner() -> Result[str, E1]` declared inside an outer try body with an uncovered conditional handler would emit `return Err(convert_to_E2(err))` inside a function returning `Result<_, E1>`. This is the codegen mirror of Blocker 2 and shares an owner boundary with Item 10H/10J; not exercised by any current fixture.

PR #3464, opened 2026-08-22
Preserve nested try/finally error propagation

yaseralnajjar:
## Summary
- recognize active closure error carriers independently of return capture
- propagate nested try/finally failures through an enclosing try closure
- add code-generation and native regressions for one-time cleanup and outer handling

## Validation
- `cargo test -p sifr_lowering try_finally -- --nocapture`
- `cargo test -p sifr_codegen --lib`
- `cargo run -q -p sifr -- run crates/sifr/tests/e2e/pass/try_nested_finally_error_propagation.sifr`
- `cargo clippy -p sifr_codegen -- -D warnings`
- `cargo fmt --check`

Exact-SHA review and compiler gates are pending.

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_nested_finally_error_propagation.sifr
+++ b/crates/sifr/tests/e2e/pass/try_nested_finally_error_propagation.sifr
@@ -1,0 +1,26 @@
+class CleanupError(Error):
+    message: str
+
+
+def fail() -> Result[None, CleanupError]:
+    raise CleanupError("expected")
+
+
+def main() -> None:
+    cleanup_markers: list[int] = []
+    caught: bool = False
+
+    try:
+        try:
+            _value: None = fail()
+        finally:
+            cleanup_markers.append(1)
+    except CleanupError as error:
+        assert error.message == "expected"
+        caught = True
+
+    assert caught
+    assert cleanup_markers == [1]
+
+
+main()
--- a/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
@@ -77,6 +77,26 @@ def value_or_zero() -> int:
 }
 
 #[test]
+fn nested_try_finally_uses_enclosing_try_error_channel_without_return_capture() {
+    let generated = generate_rust_from_source(&format!(
+        r#"{PROBE_ERROR}
+def run() -> None:
+    try:
+        try:
+            _value: int = load_int(-1)
+        finally:
+            marker: int = 1
+    except ProbeError:
+        marker = 2
+"#
+    ));
+
+    assert!(generated.contains("return Err(__sifr_finally_err.into());"));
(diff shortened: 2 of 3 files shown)


yaseralnajjar:
Reviewed the exact diff and the surrounding try/finally emission mechanism in `crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs`.

Mechanism check on the one-line fix (`loops_try_finally.rs:672`): the closure push sites at lines 327–332 and 713–719 both push `try_closure_error_type` unconditionally but increment `try_closure_depth` only under `capture_returns`. So `depth > 0` implied `!error_type.is_empty()` but not the reverse — exactly the gap that made a nested try/finally inside a non-return-capturing `except` closure fall into the `unreachable!` arm. Swapping the predicate is a strict widening of the previously-correct set, so no prior shape changes behavior. Both stacks are popped before the handler chain (line 340) and before the finally body (lines 725–726, 778), so a try/finally in a handler or a finally body still correctly sees an empty stack. `err_ty` is taken from `try_closure_error_type.last()` (line 650), so the emitted `Err(__sifr_finally_err.into())` is an identity conversion into the enclosing closure's error channel — the original typed error survives, and the finally body is still emitted unconditionally at line 783 before the match/if-let, so it runs exactly once on both paths.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: the new codegen test covers only the `capture_returns == false` path (the `IfLet` at `loops_try_finally.rs:833`). The identical `can_return_error` branch in the `Match` arm at `loops_try_finally.rs:806-827` (nested try/finally that *does* contain a `return`, inside a non-Result enclosing function with a try closure) is unexercised by any new test. Same predicate, so low risk, but it is an untested twin.
- suggestion: `loops_try_finally.rs` is 859 lines against the 900-line guardrail. It passes today, but the next change in this file will likely force the split.
- infrastructure issue: `try_nested_finally_error_propagation.sifr` is not in `verification/areas/core_language/data/create_pr_e2e_manifest.json` (curated 140-fixture subset), so the new fixture only runs under the merge/nightly/release corpora, not the create-pr lane. Intentional for a curated subset, but worth a deliberate call given this is a regression guard.

yaseralnajjar:
Exact candidate: `37e70db47fb08fc8e60defd242eb3f437056cdd1`

Gate evidence:
- `scripts/run_all_tests.sh --profile create-pr` ran once. It passed cache setup, HIR maintainability, the 3,221-file size guardrail, demo freshness, source dependency direction, submodule ownership, sysroot certification, stdlib ownership/reachability/schema/bootstrap guards, driver maintainability, and verification-runner self-tests. It stopped only at linked delivery A: missing `negative/package_generated_type_import_rejected.sifr` in the Rust-interop fixture matrix.
- `scripts/run_all_tests.sh` ran once. It passed the same preceding checks and stopped only at the same linked delivery A path. Merge-lane wall time was 98.33s, within its warm budget.

Per the phase rules and user authorization, neither exact-SHA gate will be rerun. The stale evidence path remains owned by linked delivery A.

PR #3466, opened 2026-08-22
Close diagnostic harness strict Clippy

yaseralnajjar:
## Summary
- declare the data-free fixture-layout mode as a copyable value
- preserve existing by-value call sites and canonical fixture-path behavior
- close the strict `needless_pass_by_value` workspace Clippy failure

## Validation
- both diagnostic-rendering target seed commands
- `python3 verification/areas/developer_tooling/check_diagnostic_source_canonicalization_rules.py`
- `cargo fmt --check`

Exact-SHA review and compiler gates are pending.

### DIFF
--- a/crates/sifr_driver/src/bin/diagnostic_rendering_harness.rs
+++ b/crates/sifr_driver/src/bin/diagnostic_rendering_harness.rs
@@ -263,6 +263,7 @@ fn check_cycle_runtime_rules(root: &Path) -> Result<(), String> {
     Ok(())
 }
 
+#[derive(Clone, Copy)]
 enum FixtureLayout {
     Flat,
     SourceRoot,


yaseralnajjar:
## Review: Item 10I

**Diff reviewed:** single line — `#[derive(Clone, Copy)]` added to `FixtureLayout` at `crates/sifr_driver/src/bin/diagnostic_rendering_harness.rs:266`.

**Mechanism check:**
- `FixtureLayout` is a data-free two-variant enum. `Copy` is the correct value semantics here; it is the canonical resolution for `clippy::needless_pass_by_value` on a by-value param (the alternative — borrowing — would be gratuitous for a 0-sized discriminant).
- `fixture_name_for_seed` (`:272-313`) takes `layout: FixtureLayout` and only `match`es it once (`:286`). Deriving `Copy` changes no move/borrow behavior at either call site (`:197` `Flat`, `:225` `SourceRoot`) — both pass a freshly constructed literal. Fixture-name derivation (`strip_prefix`, single-component check) is untouched, so canonical fixture names are unchanged by construction.
- No `#[allow]` was added. No test signatures depend on the change, consistent with the scope note to touch tests only if the signature required it — it did not.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `Clone` is only present because `Copy` requires it; that is fine, but if the harness later grows more layout variants carrying data (e.g. a nested-source-root path segment), the `Copy` derive will need revisiting rather than being assumed. Non-blocking and speculative.

yaseralnajjar:
Exact candidate: `192eed970cf67d917ccf39c25c425d204edd2d9c`

Gate evidence:
- `scripts/run_all_tests.sh --profile create-pr` ran once. It passed cache setup, HIR maintainability, the 3,221-file size guardrail, demo freshness, source dependency direction, submodule ownership, sysroot certification, stdlib ownership/reachability/schema/bootstrap guards, driver maintainability, and verification-runner self-tests. It stopped only at linked delivery A: missing `negative/package_generated_type_import_rejected.sifr` in the Rust-interop fixture matrix.
- `scripts/run_all_tests.sh` ran once. It passed the same preceding checks and stopped only at the same linked delivery A path.

Per the phase rules and user authorization, neither exact-SHA gate will be rerun. The stale evidence path remains owned by linked delivery A.

PR #3468, opened 2026-08-22
Scope nested function signature metadata

yaseralnajjar:
## Summary
- key nested-function metadata by lexical binding identity
- preserve full parameter shape during nested-function inference
- restore codegen callable metadata at scoped block boundaries
- add lowering, codegen, and native regressions for same-named helpers

## Validation
- `cargo test -q -p sifr_lowering --lib`
- `cargo test -q -p sifr_codegen --lib`
- `cargo clippy -q -p sifr_lowering -p sifr_codegen -- -D warnings`
- `cargo fmt --check`
- `git diff --check`
- native fixture `nested_function_signature_scope.sifr`

### DIFF
--- a/crates/sifr/tests/e2e/pass/nested_function_signature_scope.sifr
+++ b/crates/sifr/tests/e2e/pass/nested_function_signature_scope.sifr
@@ -1,0 +1,19 @@
+def evaluate(flag: bool) -> int:
+    def helper(value: int = 40, *extra: int) -> int:
+        return value + len(extra)
+
+    if flag:
+        def helper(prefix: str, suffix: str = "!") -> str:
+            return prefix + suffix
+
+        assert helper("ok") == "ok!"
+
+    return helper() + helper(value=1) + helper(1, 2, 3)
+
+
+def main() -> None:
+    assert evaluate(False) == 44
+    assert evaluate(True) == 44
+
+
+main()
--- a/crates/sifr/tests/e2e/pass/nested_function_signature_scope.sifr
+++ b/crates/sifr/tests/e2e/pass/nested_function_signature_scope.sifr
@@ -11,9 +11,47 @@ def evaluate(flag: bool) -> int:
     return helper() + helper(value=1) + helper(1, 2, 3)
 
 
+def evaluate_loop(count: int) -> int:
+    def helper(value: int = 3, *rest: int) -> int:
+        return value + len(rest)
+
+    for i in range(count):
+        def helper(value: str) -> str:
+            return value
+
+        assert helper("x") == "x"
+
+    return helper() + helper(1, 2, 3)
+
+
+def evaluate_exiting_branch(flag: bool) -> int:
+    def helper(value: int = 3, *rest: int) -> int:
+        return value + len(rest)
+
+    if flag:
+        def helper(value: str) -> str:
+            return value
+
+        return len(helper("x"))
+
+    return helper() + helper(1, 2, 3)
+
+
+def inferred_vararg() -> int:
+    def helper(*extra):
+        return len(extra)
+
+    return helper(1, 2)
+
+
 def main() -> None:
     assert evaluate(False) == 44
     assert evaluate(True) == 44
+    assert evaluate_loop(0) == 6
+    assert evaluate_loop(1) == 6
+    assert evaluate_exiting_branch(False) == 6
+    assert evaluate_exiting_branch(True) == 1
+    assert inferred_vararg() == 2
 
 
 main()
--- a/crates/sifr_codegen/src/lib_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests.rs
@@ -71,6 +71,8 @@ mod multi_module_stdlib_feature_tests;
 #[cfg(test)]
 mod nested_container_capture_codegen_tests;
 #[cfg(test)]
+mod nested_function_signature_scope_codegen_tests;
+#[cfg(test)]
 mod performance_codegen_tests;
 #[cfg(test)]
 mod performance_nested_mutation_codegen_tests;
--- a/crates/sifr_codegen/src/lib_codegen_tests/nested_function_signature_scope_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/nested_function_signature_scope_codegen_tests.rs
@@ -1,0 +1,23 @@
+use super::generate_rust_from_source;
+
+#[test]
+fn shadowed_nested_helper_restores_outer_calling_conventions() {
+    let generated = generate_rust_from_source(
+        r#"def evaluate(flag: bool) -> int:
+    def helper(value: int = 40, *extra: int) -> int:
+        return value + len(extra)
+
+    if flag:
+        def helper(prefix: str, suffix: str = "!") -> str:
+            return prefix + suffix
+        assert helper("ok") == "ok!"
+
+    return helper() + helper(value=1) + helper(1, 2, 3)
+"#,
+    );
+
+    assert!(generated.contains("helper(40_i64, &"));
+    assert!(generated.contains("helper(1_i64, &vec![2_i64, 3_i64])"));
+    assert!(!generated.contains("helper(&(40_i64)"));
+    syn::parse_file(&generated).expect("scoped nested-function calls should parse as Rust");
+}
--- a/crates/sifr_codegen/src/lib_codegen_tests/nested_function_signature_scope_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/nested_function_signature_scope_codegen_tests.rs
@@ -21,3 +21,45 @@ fn shadowed_nested_helper_restores_outer_calling_conventions() {
     assert!(!generated.contains("helper(&(40_i64)"));
     syn::parse_file(&generated).expect("scoped nested-function calls should parse as Rust");
 }
+
+#[test]
+fn loop_scopes_restore_outer_calling_conventions() {
+    let generated = generate_rust_from_source(
+        r#"def outer(n: int) -> int:
+    def helper(value: int = 3, *rest: int) -> int:
+        return value + len(rest)
+
+    for i in range(n):
+        def helper(value: str) -> str:
+            return value
+        assert helper("x") == "x"
+
+    return helper() + helper(1, 2, 3)
+"#,
+    );
+
+    assert!(generated.contains("helper(3_i64, &"));
+    assert!(generated.contains("helper(1_i64, &vec![2_i64, 3_i64])"));
+    syn::parse_file(&generated).expect("loop-scoped nested-function calls should parse as Rust");
+}
+
+#[test]
+fn exiting_branch_scopes_restore_outer_calling_conventions() {
+    let generated = generate_rust_from_source(
+        r#"def outer(flag: bool) -> int:
+    def helper(value: int = 3, *rest: int) -> int:
+        return value + len(rest)
+
+    if flag:
+        def helper(value: str) -> str:
+            return value
+        return len(helper("x"))
+
+    return helper() + helper(1, 2, 3)
+"#,
+    );
+
+    assert!(generated.contains("helper(3_i64, &"));
+    assert!(generated.contains("helper(1_i64, &vec![2_i64, 3_i64])"));
+    syn::parse_file(&generated).expect("exiting-branch nested-function calls should parse as Rust");
+}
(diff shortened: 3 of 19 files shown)


yaseralnajjar:
VERDICT: BLOCKED

**Finding 1 (blocking, correctness + phase-contract): scope-end restoration of callable conventions is missing on three block-emitter paths, so accepted source still emits non-compiling Rust.**

The fix is placed only in the shared scoped helper:

- `crates/sifr_codegen/src/stmt_support_emitter/stmt_block_helpers.rs:505-517` — `try_lower_scoped_stmt_block_for_ir` now snapshots/restores `callable_var_conventions` and `nested_fn_captures`.

But several block emitters bypass that helper and call the non-scoped `try_lower_stmt_block_for_ir`, hand-rolling a save/restore of `string_char_cache_vars` only:

- `crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs:122` (structured `for` body)
- `crates/sifr_codegen/src/stmt_support_emitter/stmt_block.rs:795` (`HirStmt::For` body in the IR block path)
- `crates/sifr_codegen/src/stmt_support_emitter/if_condition_lowering.rs:425` (`then_body` that always exits, no elif/else)

Mechanism: the shadowing inner `def` overwrites `callable_var_conventions["helper"]` (1 param, owned) inside the block; because the entry is never restored, the post-block call sites of the outer helper look up index 1 and find nothing, so the packed vararg is emitted by value instead of borrowed — exactly the defect the item's `if`-based fixture asserts against (`assert!(generated.contains("helper(40_i64, &"))`, `nested_function_signature_scope_codegen_tests.rs:20`).

Repro A (`for` body), same shape as the shipped fixture:

```python
def outer(n: int) -> int:
    def helper(value: int = 3, *rest: int) -> int:
        return value + len(rest)
    for i in range(n):
        def helper(a: str) -> str:
            return a
        assert helper("x") == "x"
    return helper() + helper(1, 2, 3)
```

`sifr check` passes; lowering resolves the restored outer signature correctly. Emitted body declares `let helper = |value: i64, rest: &Vec<i64>|` but emits `helper(3_i64, { … Vec<i64> … })` and `helper(1_i64, vec![2_i64, 3_i64])` — no `&`. `sifr build` fails: `error[E0308]: mismatched types … expected &Vec<i64>, found Vec<i64>` (x2), `SIFR-BUILD-0005`. Replacing the `for` with `if flag:` on the candidate produces `&({…})` / `&vec![2_i64, 3_i64]` and builds — so the difference is precisely the un-restored path.

Repro B (`if` branch that always exits, hits `if_condition_lowering.rs:425`):

```python
def outer(flag: bool) -> int:
    def helper(value: int = 3, *rest: int) -> int:
        return value + len(rest)
    if flag:
        def helper(a: str) -> str:
            return a
        return len(helper("x"))
    return helper() + helper(1, 2, 3)
```

Same missing `&` in the tail calls.

Both cases are inside the item's contract ("restore/remove metadata when a scope ends"; lowering does push/pop a scope frame for `for` bodies at `crates/sifr_lowering/src/lower/statements/control_flow.rs:774`), and both fail closed only at `rustc`, i.e. the compiler accepts source it cannot build. The focused regressions added by this diff cover the `if` case only, so the loop and always-exiting-branch paths have no coverage.

**Finding 2 (blocking, contract: nested inference vararg shape): positional argument unification now writes into the newly added vararg/kwarg slots, producing a scalar-typed vararg parameter.**

`crates/sifr_lowering/src/lower/nested_function_inference/state_collection.rs:293-353` appends vararg/kwonly/kwarg entries to `state.params` (correct, and it is what fixes the truncation). But `crates/sifr_lowering/src/lower/nested_function_inference/expression_inference.rs:393-399` maps positional call arguments onto `state.params` **by index**, so argument *k* of a variadic call now unifies into the vararg slot. For an explicitly annotated vararg this is absorbed (`unify_types(List(Int), Int)` → `List(Int)` via the `_ => current` arm, `type_unification.rs:58`), but for an inferred (unannotated) vararg the slot is `List(Unknown)` and gets collapsed to the element type:

```python
def outer() -> int:
    def helper(*extra):
        return len(extra)
    return helper(1, 2)
```

Candidate output:
```
error[SIFR-TYPE-0002]: len() argument must be a string, bytes, list, dict, tuple, set, or sized class, got 'int'
error[SIFR-TYPE-0002]: argument 1 of callable 'helper': expected 'int', got 'list[int]'
```
The second diagnostic is the parameter-vector mismatch the contract forbids, now visible in the binding's `Callable` type (`extra: int`) versus the vararg packing lowering performs (`list[int]`). Also note `finalize_nested_function_types` (`state_collection.rs:423-436`) only flags `param.ty.is_unknown()`, so an untouched inferred vararg keeps `List(Unknown)` with no `TYPE_MISSING_ANNOTATION` diagnostic and reaches codegen as `Vec<Box<dyn Any>>` (verified via `sifr emit`). This is not a regression in acceptance (pre-diff the slot did not exist at all and the call vector was silently truncated), but it is the same "vararg parameter shape" guarantee the item claims, and it has no regression coverage in the diff.

Everything else in the diff checks out: `BindingId`s are monotonic and never recycled (`crates/sifr_lowering/src/scope.rs:349-356`), so `local_function_metadata` cannot alias across popped scopes; `BindingKind::Function` is created only by `Scope::define_function`, so the `INTERNAL_COMPILER_PANIC` guard in `regular_calls.rs:122-130` is unreachable by construction; dropping the by-name `ctx.functions`/`function_defaults`/`vararg_functions`/`function_workload_annotations` writes for nested defs does not strand any remaining reader (module-level/imported callables are registered on separate paths, and `LoweringResult.function_*` maps are consumed only for module-exported callables); the `declared_workload`/`offload_workload` helpers preserve the stdlib fallback for names that are not scope bindings; the new inference param order matches `extract_function_type` (`signatures_and_effects.rs:643-720`) exactly; and the two added lowering tests reproduce the asserted diagnostics verbatim on the candidate (`SIFR-CALL-0001`/`SIFR-CALL-0004`).

yaseralnajjar:
VERDICT: SATISFIED

Both original blockers are resolved.

**Blocker 1 (scope leak on paths bypassing the scoped helper) — resolved.**
- `crates/sifr_codegen/src/stmt_support_emitter/stmt_block_helpers.rs:505-517`: the scoped wrapper now snapshots/restores `callable_var_conventions` and `nested_fn_captures` in addition to `string_char_cache_vars`. These are exactly the two maps a nested `def` mutates by name (`function_emitter/scope_and_function_types.rs:454-469,639-641` for conventions; `function_emitter/nested_function_block.rs:12-17` and `lib_emitter_state.rs:406-413` for captures), and they are the maps consumed at call emission (`intrinsic_method_emitters/plain_call_args.rs:338,359`).
- Grep over `crates/sifr_codegen/src` shows the raw `try_lower_stmt_block_for_ir` now has exactly one caller — the scoped wrapper itself (`stmt_block_helpers.rs:512`). The three previously-bypassing sites are converted: structured-for (`stmt_block.rs:794`), IR-for (`loops_try_finally.rs:121`), always-exiting-if (`if_condition_lowering.rs:424`).
- The `if_condition_lowering.rs` rewrite also moves the restore earlier (before the let-else detection helpers). That is behavior-preserving: `detect_or_is_none_vars_with_bindings_for_ir` (`if_condition_lowering.rs:587-630`), `option_binding_value_expr_for_ir` and `option_binding_pattern_for_ir` (`condition_lowering.rs:129-156`) read `borrowed_params`/`mut_borrowed_params`/`local_binding_types`/`mutated_vars` only, never `string_char_cache_vars`.
- Regressions cover the three paths: `nested_function_signature_scope_codegen_tests.rs:86,108,129` assert the restored outer form (`helper(3_i64, &…)`, `helper(1_i64, &vec![2_i64, 3_i64])`) plus `syn::parse_file`, and the native fixture exercises loop and always-exiting-branch shapes (`crates/sifr/tests/e2e/pass/nested_function_signature_scope.sifr:20-43`).

**Blocker 2 (vararg/kwarg slots unified as ordinary positionals) — resolved.**
- `state_collection.rs:40-46,299-380` now records `ParamKind` and appends vararg/kwonly/kwarg in the *same order* `extract_function_type` uses (`signatures_and_effects.rs:644-722`), so the inferred `FunctionType` is shape-identical to the annotated one. That also fixes the downstream body binding at `statements/statement_dispatch.rs:707-715`, which reads the vararg type from `ft.params[args.len()]` — previously out of range for inferred varargs, silently `Type::Any`.
- `call_argument_inference.rs:12-25` maps fixed positionals by `ParamKind::Positional` index and merges surplus positionals into the vararg *element* type via `unify_variadic_param_value` (`:51-78`), re-wrapping with `variadic_container_type` (`:80-86`), so an inferred `*extra` stays `List[T]` instead of collapsing to a scalar. Keyword handling is symmetric: named keywords bind to `Positional | KeywordOnly` (`:40-44`), unmatched names merge into the kwarg value type, and `**splat` (`name == None`) unifies whole-dict against the kwarg param (`:33-38`).
- `state_collection.rs:437` tightening to `contains_unknown_or_any()` closes the `List[Unknown]`/`Dict[_, Unknown]` escape.
- Coverage: `nested_function_tests.rs:759,771,786` (restoration, non-inheritance of default/vararg with both exact arity diagnostics, inferred vararg list shape) plus `inferred_vararg` in the native fixture.

Contract items check out: metadata is keyed by `BindingId` from a monotonic counter (`scope.rs:350-356`), the only `BindingKind::Function` producer is `define_function` (`scope.rs:210-212`), which always inserts metadata (`signatures_and_effects.rs:609-617`), so the `INTERNAL_COMPILER_PANIC` guard in `regular_calls.rs:120` is unreachable by construction. Resolution is lexical via `visible_local_function_metadata` (`mod_context.rs:379-388`) with block scoping supplied by `ctx.scope.push()/pop()` around `lower_stmts` (`statements/control_flow.rs:495-497`); no basename fallback was added, and non-function `Callable` bindings still take the ordinary path (`regular_calls.rs:108-112`). All touched files are under the 900-line guardrail (largest: `expression_inference.rs` 836, `state_collection.rs` 823).

Findings:

1. NEW MECHANISM — `crates/sifr_lowering/src/lower/method_receiver_analysis.rs:191` (PLAUSIBLE, not verified by execution). Dropping the `ctx.functions.insert` for local functions (`signatures_and_effects.rs:606-617`) removes the only signal this pass has for plain calls: `ctx.functions.get(func)` is what marks a class method receiver-mutating when a `mut`-borrow parameter receives a receiver-rooted argument. Nested helpers previously landed in that map by basename (the pollution this item removes), so the signal existed before and does not now. Mechanism: a method whose only mutation is `helper(self.field)` — where `helper` is a nested def whose parameter is inferred `mut_borrow` because it mutates the argument (`state_collection.rs:80-95`) — is no longer added to `mutable`, since `body_contains_receiver_mutation` (`classes/class_semantics.rs:10-20`) only sees direct `self.…` assignment forms. The method would be emitted `&self` while passing `&mut self.field`. No fixture covers this shape (no `crates/sifr/tests/e2e/pass/*.sifr` combines a class method with a nested def taking a mut-borrowed receiver field), which is consistent with the suites passing. This is a distinct mechanism from either original blocker and should be recorded as a later item rather than reopening this one.

2. Note, not a finding: the `state_collection.rs:437` predicate change also newly rejects a non-explicit parameter that resolves to top-level `Type::Any` (`contains_unknown_or_any` returns true for `Any`; the old `is_unknown()` did not). That is a tightening consistent with the item's "no dynamic escape" intent and no test regressed; flagging only so the widening is on record.

yaseralnajjar:
Exact-SHA create-PR gate evidence for `796af10658f825f1df1387af91a63dcf11d4c455`:

- `scripts/run_all_tests.sh --profile create-pr` ran once.
- All executed checks passed except the already tracked linked-delivery Rust interop manifest failure: `shared_bridge_crate: missing evidence source negative/package_generated_type_import_rejected.sifr`.
- Lane wall time: 108.16s; no resource advisory.
- This stale manifest reference is outside Item 10J and belongs to the phase linked-delivery cleanup item. Per the authorized phase rule, the create-PR gate will not be rerun.

yaseralnajjar:
Exact-SHA merge gate evidence for `796af10658f825f1df1387af91a63dcf11d4c455`:

- `scripts/run_all_tests.sh` ran once.
- All executed checks passed except the same already tracked linked-delivery Rust interop manifest failure: `shared_bridge_crate: missing evidence source negative/package_generated_type_import_rejected.sifr`.
- Lane wall time: 98.54s; no resource advisory.
- This stale manifest reference is outside Item 10J and belongs to the phase linked-delivery cleanup item. Per the authorized phase rule, the merge gate will not be rerun.

PR #3470, opened 2026-08-22
Preserve child upcasts through nominal relocation

yaseralnajjar:
## Summary
- retain a user child's `From<Child>` conversion when its checked-stdlib parent moves to the project nominal prelude
- remove parent-owned conversions and relocated child conversions as before
- derive eligible conversion sources from exact HIR user classes
- add focused relocation and project-plan regressions

## Validation
- 1,132 codegen unit tests passed
- `nominal_identity_alias_paths.sifr` built and ran release-native
- strict codegen Clippy
- formatting, HIR maintainability, file-size, and diff checks

### DIFF
--- a/crates/sifr_codegen/src/lib_project_codegen.rs
+++ b/crates/sifr_codegen/src/lib_project_codegen.rs
@@ -392,6 +392,11 @@ pub fn generate_rust_multi_with_metadata(
             module_name,
             &stdlib_nominal_plan,
             &crate_root_modules,
+            &module
+                .classes
+                .iter()
+                .map(|class| source_class_rust_name(&class.name))
+                .collect(),
         );
         let imports = [local_imports, union_imports]
             .into_iter()
--- a/crates/sifr_codegen/src/lib_test_project_codegen.rs
+++ b/crates/sifr_codegen/src/lib_test_project_codegen.rs
@@ -125,6 +125,11 @@ pub fn generate_rust_test_project_with_metadata(
             module_name,
             &stdlib_nominal_plan,
             &crate_root_modules,
+            &module
+                .classes
+                .iter()
+                .map(|class| sifr_type_system::source_class_rust_name(&class.name))
+                .collect(),
         );
         support_rust_files.insert(
             (*module_name).to_string(),
--- a/crates/sifr_codegen/src/project_stdlib_nominals.rs
+++ b/crates/sifr_codegen/src/project_stdlib_nominals.rs
@@ -1,6 +1,6 @@
 use crate::stdlib_filter::{
     partition_rust_items_by_name, rust_source_defined_item_names, rust_source_references_item_name,
-    strip_rust_items_by_name,
+    strip_relocated_rust_items_by_name, strip_rust_items_by_name,
 };
 use crate::{
     build_error_into_error_impl, generate_rust_with_stdlib_for_module_with_structural_policy,
@@ -61,6 +61,7 @@ pub(crate) fn relocate_project_stdlib_nominals(
     module_name: &str,
     plan: &ProjectStdlibNominalPlan,
     crate_root_modules: &HashSet<&str>,
+    local_class_rust_names: &HashSet<String>,
 ) -> String {
     if plan.registry.shared_rust_names.is_empty() && plan.registry.crate_root_rust_names.is_empty()
     {
@@ -73,7 +74,7 @@ pub(crate) fn relocate_project_stdlib_nominals(
         .chain(&plan.registry.crate_root_rust_names)
         .map(String::as_str)
         .collect();
-    let stripped = strip_rust_items_by_name(source, &names);
+    let stripped = strip_relocated_rust_items_by_name(source, &names, local_class_rust_names);
     if crate_root_modules.contains(module_name) {
         return stripped;
     }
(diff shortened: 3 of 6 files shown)


yaseralnajjar:
VERDICT: SATISFIED

## Mechanism verification (base..candidate only)

**Retention predicate is exact and identity-safe.** `crates/sifr_codegen/src/stdlib_filter/relocation.rs:29-62` retains an item only when it is an `Item::Impl` whose trait path ends in `From`, whose first generic argument is a `Type::Path`, and whose last path segment ident is an exact member of the module's own class Rust-name set — with `!relocated_names.contains(source_name)` (line 61) killing conversions whose source is itself relocated. No `.clone()`, no basename/substring matching, no fallback branch, no duplicate parent emission. Non-`From` traits, inherent impls, structs, and `From<String>`-style parent-owned conversions still hit the unchanged strip path (`relocation.rs:44-46`), so exactly one project-wide parent definition survives in `__sifr_project_nominals` (`project_stdlib_nominals.rs:255-286`).

**Name set matches the emitter.** `local_class_rust_names` is built with `source_class_rust_name(&class.name)` (`lib_project_codegen.rs:395-399`, `lib_test_project_codegen.rs:128-132`), which is the same function `RustEmitter::class_impl_target` (`class_emitter.rs:85-91`) uses to spell the `From<Child>` argument, including `__Sifr*`-escaped user names such as the fixture's `__SifrIdentityBox`. Generic children render as `Child<T>`; the predicate reads the last segment ident, so `From<Child<T>>` is retained. Generic classes are not re-emitted in importing modules — `register_imported_generic_classes` (`lib_project_codegen.rs:278-306`) only registers templates for inference — and `class_parent_deref_impls` is reached solely through `emit_module_classes` over `module.classes` (`module_body.rs:26-29`), so the conversion is emitted exactly once, in the defining module, which is precisely the module whose class set the filter consults. No multi-module omission or duplication.

**Parent identity and coherence.** Deref/DerefMut keep the child as self type and are never candidates for stripping, so field + `Deref` + consuming upcast stay on the one relocated parent identity. Target resolution: non-crate-root modules get `use crate::<Name>;` from the pre-existing loop (`project_stdlib_nominals.rs:78-93`), now reachable because the retained impl is the reference; `main` returns early and resolves through the prelude's `pub use __sifr_project_nominals::<Name>;`. Both the child and the parent are crate-local, so the retained impl is orphan-rule legal.

**Parent-owned conversions are not falsely retained.** The per-module `From<X> for Error` widenings at `lib_modules_and_codegen.rs:561-573` are sourced only from `BUILTIN_ERROR_CLASSES`, which are registered as relocated shared names, so the line-61 guard strips them — asserted directly by `stdlib_filter/tests.rs:34-40` (`From<String>` and `From<RelocatedChild>` both gone, `fn inherited` gone).

## Non-blocking observations

- `crates/sifr/tests/e2e_support/fixture_compilation.rs:4` compiles pass fixtures via `sifr_driver::compile_with_metadata`, i.e. the single-file codegen path (`build/project_codegen.rs:45-70`), which never calls `relocate_project_stdlib_nominals`. `nominal_identity_alias_paths.sifr` therefore exercises the fixed path only when built through the CLI (workspace root found → `CompilationMode::Project`, `cli_model_and_entrypoint.rs:717-725`), as your release-native run did. In the automated suite the fix is guarded by the two new unit tests (`stdlib_filter/tests.rs:4-41`, `project_stdlib_nominals.rs:812-838`) only. Matches the Item 10F precedent; noting it so the guard's scope is on record rather than assumed.
- The predicate does not check that the retained source class is the *declared child* of that parent, only that it is a locally declared class. The one other producer of `From<X> for Error` with a possibly-local `X` is the async-Python contract path (`lib_modules_and_codegen.rs:574-580`); a duplicate with the prelude copy (`project_stdlib_nominals.rs:243-251`) would require a user-declared Python error contract class in a project module, which the prelude already emits as a bare unresolvable source ident inside `mod __sifr_project_nominals` — i.e. that combination is broken independently of this diff, not regressed by it.

yaseralnajjar:
Exact-SHA create-PR gate evidence for `b3146d9fca6647132fc5ffd1a7e3fbc03fb9b233`:

- `scripts/run_all_tests.sh --profile create-pr` ran once after the mandatory private-target cleanup.
- All executed checks passed except the already tracked linked-delivery Rust interop manifest failure: `shared_bridge_crate: missing evidence source negative/package_generated_type_import_rejected.sifr`.
- The cold-cache lane completed in 166.47s with no resource advisory; timing is not used as performance evidence.
- The stale path is outside Item 10K and belongs to linked delivery A. Per the phase rule, this gate will not be rerun.

yaseralnajjar:
Exact-SHA merge gate evidence for `b3146d9fca6647132fc5ffd1a7e3fbc03fb9b233`:

- `scripts/run_all_tests.sh` ran once.
- All executed checks passed except the same already tracked linked-delivery Rust interop manifest failure: `shared_bridge_crate: missing evidence source negative/package_generated_type_import_rejected.sifr`.
- Lane wall time: 97.27s; no resource advisory.
- The stale path is outside Item 10K and belongs to linked delivery A. Per the phase rule, this gate will not be rerun.

PR #3472, opened 2026-08-22
Match user error parent handlers by ancestry

yaseralnajjar:
## Summary
- recognize transitive user-defined error subclasses through the canonical error registry
- match handlers with exact nominal ancestry when resolved source and target types are available
- convert child bindings to the declared parent and preserve unrelated residuals
- add lowering, codegen, exact-identity, and native regressions

## Validation
- 1,039 lowering tests passed; one ignored
- 1,134 codegen tests passed
- new `try_user_error_parent_handler.sifr` release-native pass
- existing `imported_error_not_catch_all.sifr` release-native pass
- strict targeted Clippy, fmt, HIR maintainability, file-size, and diff checks

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_user_error_parent_handler.sifr
+++ b/crates/sifr/tests/e2e/pass/try_user_error_parent_handler.sifr
@@ -1,0 +1,58 @@
+class BaseError(Error):
+    message: str
+
+
+class ChildError(BaseError):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+
+
+class OtherError(Error):
+    message: str
+
+
+def classify(child: bool) -> Result[str, ChildError | OtherError]:
+    if child:
+        raise ChildError("child")
+    raise OtherError("other")
+
+
+def handle_child() -> str:
+    try:
+        value: str = classify(True)
+        return value
+    except BaseError as error:
+        return "base:" + error.message
+    except OtherError as error:
+        return "other:" + error.message
+
+
+def handle_other() -> str:
+    try:
+        value: str = classify(False)
+        return value
+    except BaseError as error:
+        return "wrong:" + error.message
+    except OtherError as error:
+        return "other:" + error.message
+
+
+def propagate_other() -> Result[str, OtherError]:
+    try:
+        value: str = classify(False)
+        return value
+    except BaseError as error:
+        return "wrong:" + error.message
+
+
+def main() -> None:
+    assert handle_child() == "base:child"
+    assert handle_other() == "other:other"
+    try:
+        _: str = propagate_other()
+        assert False
+    except OtherError as error:
+        assert error.message == "other"
+
+
+main()
--- a/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
@@ -474,3 +474,36 @@ def classify(flag: bool) -> Result[str, IOError | ValueError]:
     assert!(!generated.contains("structured statement emission missing"));
     syn::parse_file(&generated).expect("union residual Rust should parse");
 }
+
+#[test]
+fn user_error_parent_handler_converts_child_and_preserves_unrelated_residual() {
+    let generated = generate_rust_from_source(
+        r#"
+class BaseError(Error):
+    message: str
+
+class ChildError(BaseError):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+
+class OtherError(Error):
+    message: str
+
+def classify(flag: bool) -> Result[str, ChildError | OtherError]:
+    if flag:
+        raise ChildError("child")
+    raise OtherError("other")
+
+def handle(flag: bool) -> Result[str, OtherError]:
+    try:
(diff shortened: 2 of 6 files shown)


yaseralnajjar:
VERDICT: SATISFIED

**Contract-by-contract confirmation (read-only inspection of base..candidate):**

1. **Same predicate as lowering coverage.** `crates/sifr_codegen/src/stmt_support_emitter/try_handlers.rs:46-52` uses `source.is_assignable_to(target)`, which is exactly the predicate lowering coverage uses at `crates/sifr_lowering/src/lower/try_error_propagation.rs:22-24` (`error_ty.is_assignable_to(covered)`). Codegen and lowering can no longer disagree by construction.

2. **Identity-based ancestry, no basename matching.** `Type::is_assignable_to`'s class arm (`crates/sifr_type_system/src/types/type_rendering.rs:571-597`) compares `identity` (falling back to `name` only when identity is absent) and walks the `parent_class` chain by identity. Chains are built transitively and identity-qualified in `class_type_collection.rs:341-359` (`{parent_identity}|{chain}`), so multi-level ancestry matches and `other.BaseError` vs `pkg.BaseError` does not — asserted by the new unit test at `try_handlers.rs:271-302`.

3. **No global catch-all.** The catch-all shortcut is still confined to `handler_is_catch_all` (`try_error_carrier.rs:38-43`, bare `except`/builtin `Error` base only), checked before the new branch; `same_basename_import_is_not_a_catch_all` still guards that. The new branch returns `Unsupported`, not `Always`, on non-ancestry, so a parent handler does not absorb unrelated errors.

4. **Handler ordering intact.** The condition is evaluated per handler in declaration order in `lower_single_try_except_handler_chain_for_ir`; the reverse fold at `try_handlers.rs:200-210` lets the first `Always` branch shadow later ones, matching Python first-match semantics. Child-before-parent and parent-before-child orders both resolve correctly (a child-typed handler against a parent source is correctly `Unsupported`, not a false match).

5. **Binding conversion through checked ancestry.** No new conversion path was added; the existing `consuming_value_conversion_for_ir` → `consuming_class_upcast_for_ir` (`class_upcasts.rs:197-248`) walks the identity chain up to the target index, which is why the codegen test asserts `Into::<BaseError>::into(__sifr_try_variant_error.clone())`. The binding uses `.clone()`, so the residual `return Err(__sifr_try_variant_error)` still owns the original value — no double move; both are asserted in `sequential_try_binding_codegen_tests.rs:504-506`.

6. **Residual preserved.** The unmatched-member path is untouched (`try_handlers.rs:189-199` / `unmatched_try_error_return_for_ir`), and `propagate_other()` in the new e2e fixture exercises parent match on one union member plus residual propagation of the other.

7. **Classifier unification.** `is_error_class_with_ctx` was already present and dead; the diff activates it and removes the duplicate direct-only `is_error_class` (no remaining references anywhere in `crates/`). Transitive registration flows outward correctly: `ctx.error_types` → `HirClass::is_error_type` → `error_exports` (`sifr_frontend/src/query_diagnostics.rs:252`), so imported child errors stay classified as errors.

8. **Regressions present and focused.** Lowering (`statement_diagnostics_tests.rs:242-260`), codegen string-level (`sequential_try_binding_codegen_tests.rs:478-508`), a direct identity/ancestry unit test on the condition function, and a native e2e fixture covering parent match, unrelated-handler selection, and residual. The e2e fixture follows pass-suite convention (asserts + trailing `main()`, no companion expectation file, auto-discovered).

**Non-blocking observations (no action required for this item):**

- `try_handlers.rs:42-45`: `resolved_error_type` is still rendered eagerly but is now reachable only when `source_error_type` is `None` (the new branch returns whenever both are `Some`). Harmless dead work on the metadata-bearing path; the string fallback itself is correctly retained for handlers/carriers without resolved types.
- Transitive error classification resolves the parent by local name against the `ctx.error_types` name set, so it is declaration-order sensitive within the two source-order class passes (`mod_impl.rs:78-90`). Safe in practice because a base class must be defined before its subclass, and the second pass reruns after the first registers parents — but it is a name-set lookup, not an identity lookup, unlike the dispatch predicate.
- Cross-module and grandparent-level handler ancestry are not covered by a fixture in this diff. They fail closed rather than miscompiling: any identity/chain gap makes lowering report uncovered errors with the same predicate codegen uses, so no divergence is possible.

yaseralnajjar:
Create-PR gate on exact candidate `176e67d2752a12000b0727017ca875da40bfc974`: STOPPED only by the recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed, including HIR maintainability, 3,226-file size guardrail, demo freshness, dependency direction, ownership, sysroot/stdlib guards, driver maintainability, and verification self-tests. This gate will not be rerun, per the phase rule and standing authorization.

yaseralnajjar:
Merge gate on exact candidate `176e67d2752a12000b0727017ca875da40bfc974`: STOPPED only by the same recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed. This is the single merge gate and will not be rerun, per the phase rule and standing authorization.

PR #3474, opened 2026-08-22
Isolate nested function try channels

yaseralnajjar:
Closes phase Item 10M. Nested structured functions now save and clear all active try-closure state, build their own error channel, and restore the enclosing state after emission. Adds sync and async codegen regressions plus a release-native fixture. Exact-SHA Opus review and the single permitted gates will be attached.

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_nested_function_channel_isolation.sifr
+++ b/crates/sifr/tests/e2e/pass/try_nested_function_channel_isolation.sifr
@@ -1,27 +1,16 @@
-def fail_inner() -> Result[int, ValueError]:
-    raise ValueError("inner")
-
-
-async def fail_inner_async() -> Result[int, ValueError]:
-    await task.sleep(0.0)
-    raise ValueError("inner async")
-
-
 def fail_outer() -> Result[int, IOError]:
     raise IOError("outer")
 
 
 def run_sync() -> Result[int, IOError]:
     try:
-        def inner() -> Result[int, ValueError]:
-            try:
-                value: int = fail_inner()
-            except FileNotFoundError:
-                return 0
-            return value
+        def inner(succeed: bool) -> Result[int, ValueError]:
+            if succeed:
+                return 1
+            raise ValueError("inner")
 
-        nested: Result[int, ValueError] = inner()
-        assert str(nested) == 'Err(ValueError { message: "inner" })'
+        nested_text: str = str(inner(False))
+        assert nested_text == 'Err(ValueError { message: "inner" })'
         outer_value: int = fail_outer()
         return outer_value
     except IOError:
@@ -30,15 +19,14 @@ def run_sync() -> Result[int, IOError]:
 
 async def run_async() -> Result[int, IOError]:
     try:
-        async def inner() -> Result[int, ValueError]:
-            try:
-                value: int = await fail_inner_async()
-            except FileNotFoundError:
-                return 0
-            return value
+        async def inner(succeed: bool) -> Result[int, ValueError]:
+            await task.sleep(0.0)
+            if succeed:
+                return 1
+            raise ValueError("inner async")
 
-        nested: Result[int, ValueError] = await inner()
-        assert str(nested) == 'Err(ValueError { message: "inner async" })'
+        nested_text: str = str(await inner(False))
+        assert nested_text == 'Err(ValueError { message: "inner async" })'
         outer_value: int = fail_outer()
         return outer_value
     except IOError:
--- a/crates/sifr/tests/e2e/pass/try_nested_function_channel_isolation.sifr
+++ b/crates/sifr/tests/e2e/pass/try_nested_function_channel_isolation.sifr
@@ -1,0 +1,59 @@
+def fail_inner() -> Result[int, ValueError]:
+    raise ValueError("inner")
+
+
+async def fail_inner_async() -> Result[int, ValueError]:
+    await task.sleep(0.0)
+    raise ValueError("inner async")
+
+
+def fail_outer() -> Result[int, IOError]:
+    raise IOError("outer")
+
+
+def run_sync() -> Result[int, IOError]:
+    try:
+        def inner() -> Result[int, ValueError]:
+            try:
+                value: int = fail_inner()
+            except FileNotFoundError:
+                return 0
+            return value
+
+        nested: Result[int, ValueError] = inner()
+        assert str(nested) == 'Err(ValueError { message: "inner" })'
+        outer_value: int = fail_outer()
+        return outer_value
+    except IOError:
+        return 42
+
+
+async def run_async() -> Result[int, IOError]:
+    try:
+        async def inner() -> Result[int, ValueError]:
+            try:
+                value: int = await fail_inner_async()
+            except FileNotFoundError:
+                return 0
+            return value
+
+        nested: Result[int, ValueError] = await inner()
+        assert str(nested) == 'Err(ValueError { message: "inner async" })'
+        outer_value: int = fail_outer()
+        return outer_value
+    except IOError:
+        return 42
+
+
+async def main() -> None:
+    try:
+        sync_value: int = run_sync()
+        assert sync_value == 42
+    except IOError:
+        assert False
+
+    try:
+        async_value: int = await run_async()
+        assert async_value == 42
+    except IOError:
+        assert False
--- a/crates/sifr_codegen/src/function_emitter/scope_and_function_types.rs
+++ b/crates/sifr_codegen/src/function_emitter/scope_and_function_types.rs
@@ -487,6 +487,11 @@ impl RustEmitter {
         let saved_current_sifr_int_result_return = self.current_sifr_int_result_return.get();
         let saved_python_context_counter = self.python_context_counter;
         let saved_python_context_envelope_depth = self.python_context_envelope_depth;
+        let saved_try_closure_depth = std::mem::replace(&mut self.try_closure_depth, 0);
+        let saved_try_closure_option_wrap = std::mem::take(&mut self.try_closure_option_wrap);
+        let saved_try_closure_error_type = std::mem::take(&mut self.try_closure_error_type);
+        let saved_try_closure_error_type_info =
+            std::mem::take(&mut self.try_closure_error_type_info);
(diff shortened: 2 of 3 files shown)


yaseralnajjar:
VERDICT: BLOCKED

## Blocker 1 — All three added regressions are vacuous (they pass with the fix reverted)

**Files:** `crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs:254-326`, `crates/sifr/tests/e2e/pass/try_nested_function_channel_isolation.sifr:1-59`

**Evidence (produced read-only; repo and git state untouched — scratch copy at `/tmp/sifr_base_check`, scratch target `/tmp/sifr_base_target`):**

1. Copied the tree, reverse-applied *only* the source hunk (`scope_and_function_types.rs:490-494` + `575-578`), left both new unit tests in place, rebuilt:
   `cargo test -p sifr_codegen -- nested_function_try_channel_is_isolated nested_async_function_try_channel` → **2 passed, 0 failed** at base behavior.
2. `sifr emit crates/sifr/tests/e2e/pass/try_nested_function_channel_isolation.sifr` at base-behavior vs. candidate → **byte-identical generated Rust** (193 lines, `diff` clean). The native fixture therefore also passes at base and cannot distinguish E1 from E2.

**Mechanism of the vacuity:**
- In both test programs the only carrier-sensitive construct in `inner` is a residual that already sits inside `inner`'s own `try`, which pushes `ValueError` onto `try_closure_error_type` — so base already resolves E1 there.
- `inner`'s `return value` is a *tail* return, emitted as a tail expression (`Ok(value)`), not through `try_lower_structured_return_stmt`; the enclosing `direct_return_capture` is also true, so `try_closure_option_wrap` is `false` and even a non-tail return would coincide with the correct form. I confirmed this with a probe using a non-tail `return` inside a structured nested fn: base and candidate emit the same `return Ok(7_i64)`.
- Assertion-level weaknesses: `:286`/`:324` `!generated.contains("Into::<IOError>::into(__sifr_try_err)")` can never fire — the enclosing-carrier conversion is rendered as a `.into()` **method call**, never as `Into::<T>::into`. `:285`/`:323` `contains("return Err(__sifr_try_err);")` and `:284`/`:322` `contains("Result<Result<i64, IOError>, IOError>")` hold verbatim at base. Assertions are matched against the whole file, not scoped to `inner`.

**A shape that does exercise the change** (use this, or equivalent, for the required regression): a nested function containing a `raise` of E1 inside an enclosing E2 try, which goes through `coerce_error_type_for_ir` (`structured_return_if_while.rs:168-176`, reading `try_closure_error_type_info.last()`):

```python
def run() -> Result[int, IOError]:
    try:
        def inner() -> Result[int, ValueError]:
            raise ValueError("inner")
        text: str = str(inner())
        ...
    except IOError:
        return 42
```
- base emits `Err(ValueError::new("inner".to_string()).into())` inside the nested closure; candidate emits it without `.into()`.
- Compiling the base-emitted file with `rustc` fails hard: `error[E0283]: type annotations needed … cannot satisfy _: From<ValueError>`. So the defect is real and the fix is real — but nothing in this diff guards it. Add sync + async unit regressions on this shape (and a fixture that raises E1 from the nested function), and confirm each fails when the four-field save/clear is reverted.

## Non-blocking observations

- **Isolation mechanism itself is correct.** `scope_and_function_types.rs:490-494` moves all four fields out before `register_local_body_binding_types` / body lowering, and `:575-578` restores them after; there is no early `return`, `?`, or panic between save and restore (verified by scanning 490-580), so restoration cannot be skipped. Restoration happens before the param/return-type lowering at `:589-646`, which does not read try-closure state. Both the recursive `RustStmt::LocalFn` branch and the closure branch share the isolated region, and `func.is_async` only affects the emitted form — no sync/async divergence. Nested `try` statements push/pop onto the now-empty stacks (`loops_try_finally.rs:326-339`, `:713-724`), and `try_finally_error_type_name_for_ir` (`:643-661`) correctly falls back to the nested function's own `current_return_type`. The stateless simple path (`lower_stmt/simple_dispatch_and_bindings.rs:522`) needs no change; `try_lower_nested_function_stmt_for_block` delegates to the same function, so both structured entry points are covered. No other emission path for `HirStmt::NestedFunction` exists.
- The stated root cause "nested residual dispatch read the enclosing resolved error stack" is not directly reachable: the frontend rejects a residual outside a `try` (`SIFR-TYPE-0002`), so a residual always sits under a `try` that pushes its own carrier. The reachable form of the defect is the `raise`/re-raise coercion path above — worth correcting in the item record so the regression targets the right construct.
- Pre-existing on both SHAs (not a regression from this diff, no action requested): an unannotated nested closure whose result is only consumed by `str(...)` emits `Err(...)` with an uninferable `Ok` type (`error[E0282]`).
- File sizes fine: `scope_and_function_types.rs` 836 lines, test file 583 lines. E2E fixture conventions (no expected-output companion, `async def main`, `str(...) == 'Err(... )'` assertions) match existing fixtures; no fixture manifest needs updating.

yaseralnajjar:
VERDICT: SATISFIED

**Original blocker: closed.** The corrected regressions now exercise the reachable mechanism — direct raise coercion inside a nested function while an enclosing try channel is active.

Mechanism confirmation (read-only, on the exact remediation candidate):

- `crates/sifr_codegen/src/stmt_support_emitter/stmt_block.rs:682` routes `HirStmt::Raise` through `coerce_raised_error_for_ir` → `coerce_error_type_for_ir` (`stmt_support_emitter/structured_return_if_while.rs:165-199`), whose target is `try_closure_error_type_info.last()` with fallback to `current_return_type`. Base leaks the enclosing `Some(IOError)` pushed at `loops_try_finally.rs:330`/`717`, so `ValueError` ≠ `IOError` → `.into()`. With the save/clear at `function_emitter/scope_and_function_types.rs:490-494` the stack is empty during nested-body lowering and `current_return_type` is already set to the nested signature (line 498, before body lowering at 556-563), so source == target → no conversion. This is exactly the direct-raise mechanism, and it is the same for sync and async.
- The positive assertions are the discriminating ones, not the negative ones: `Err(ValueError::new("inner".to_string()))` does **not** occur as a substring of `Err(ValueError::new("inner".to_string()).into())` (mismatch at the `)` vs `.` position), so `sequential_try_binding_codegen_tests.rs:274` and `:307` fail under base behavior on their own. The `!contains(...into())` lines are redundant reinforcement, not the load-bearing check. Neither test is vacuous.
- The `str(inner(False))` consumers avoid an explicitly typed `Result` assignment, so base cannot resolve the stale `.into()` as an identity conversion. Verified in emitted Rust: the nested closure has no return annotation, so `Ok(1_i64)` + `Err(ValueError::new(...))` are the only type anchors — the stale form leaves the error type unresolved (matches the reported E0282).

State isolation / restoration:

- All four try-closure fields (`depth`, `option_wrap`, `error_type`, `error_type_info`) are the complete channel set (`lib_emitter_state.rs:128-134`), cleared at 490-494 and restored at 575-578.
- The save/restore window is straight-line: no `return`, `?`, or panic between 490 and 578, so restoration cannot be skipped.
- `try_lower_structured_nested_function_stmt` is the only structured nested-function lowering entry (`function_emitter/nested_function_block.rs` → `scope_and_function_types.rs:377`). The other `saved_return_type` sites (`generator_bodies.rs:252` `emit_function`, `class_emitter.rs`, `class_method_emitter.rs`, `function_like_lowering.rs`) are top-level/method emission and are not reachable with an active enclosing try channel. The simple path `try_lower_simple_nested_function_stmt` (`lower_stmt/simple_dispatch_and_bindings.rs:325`) coerces from its passed `return_type` (`coerce_simple_raised_error`, same file) and never reads the try-closure stack, so it is not a parallel leak.
- Restoration is genuinely exercised: in the emitted fixture code, `fail_outer()?` follows the nested def inside the try closure and still resolves against the `Result<Result<i64, IOError>, IOError>` carrier in both `run_sync` and `run_async`; a missing restore would leave the popped stacks empty for that `?`.

Independent validation I ran:

- `nested_function_try_channel_is_isolated_and_enclosing_channel_is_restored` — pass.
- `nested_async_function_try_channel_is_isolated` — pass.
- `sifr emit` on the fixture: nested raises emit `Err(ValueError::new("inner".to_string()))` and `Err(ValueError::new("inner async".to_string()))` with no `.into()` and no widening to the enclosing `IOError` carrier; both enclosing bodies keep `fail_outer()?` plus the `Err(__sifr_try_err) => return Ok(42_i64)` handler; sync and async shapes are identical apart from `async move` / `.await`.
- `sifr run` on the fixture: exits clean, all asserts hold.
- Touched files under the 900-line guardrail (836 and 571). Working tree clean at `5f7c85d5`; I made no edits and no git-state changes.

Non-blocking observations (no action requested for this item):

- `sequential_try_binding_codegen_tests.rs:305-311`: the async test omits the `Err(__sifr_try_err) =>` restoration assertion that the sync test carries at line 275. Restoration on the async path is still covered by the native fixture, so this is an assertion-strength asymmetry, not an async/sync behavioral divergence.
- Later-item finding (not a blocker on this diff): `stmt_support_emitter/try_error_helpers.rs:134` lets `first_try_error_type_in_stmts` descend into a nested function's body, so a `try/finally` with no other error-type source can adopt a nested function's try error type. That is the inbound direction of channel bleed and is untouched by this diff.

yaseralnajjar:
Create-PR gate on exact remediation candidate `5f7c85d559c34203b0392496d0f49323f7428a41`: STOPPED only by the recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed, including HIR maintainability, 3,227-file size guardrail, demo freshness, dependency direction, ownership, sysroot/stdlib guards, driver maintainability, and verification self-tests. This gate will not be rerun, per the phase rule and standing authorization.

yaseralnajjar:
Merge gate on exact remediation candidate `5f7c85d559c34203b0392496d0f49323f7428a41`: STOPPED only by the same recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed. This is the single merge gate and will not be rerun, per the phase rule and standing authorization.

PR #3476, opened 2026-08-22
Preserve nested helper receiver mutation

yaseralnajjar:
Closes phase Item 10N. Method receiver analysis now consumes checked per-call mutable-place metadata resolved from the lexical callable binding, instead of re-looking up plain calls in the module function map. Adds focused lowering/codegen regressions and a native sibling-helper fixture. Exact-SHA Opus review and the single permitted gates will be attached.

### DIFF
--- a/crates/sifr/tests/e2e/pass/nested_helper_receiver_mutation.sifr
+++ b/crates/sifr/tests/e2e/pass/nested_helper_receiver_mutation.sifr
@@ -1,0 +1,26 @@
+class Bucket:
+    values: list[int]
+
+    def __init__(self) -> None:
+        self.values = []
+
+    def update(mut self) -> None:
+        def helper(values: list[int]) -> None:
+            values.append(1)
+
+        helper(self.values)
+
+    def size(self) -> int:
+        def helper(values: list[int]) -> int:
+            return len(values)
+
+        return helper(self.values)
+
+
+def main() -> None:
+    bucket = Bucket()
+    bucket.update()
+    assert bucket.size() == 1
+
+
+main()
--- a/crates/sifr_codegen/src/lib_codegen_tests/receiver_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/receiver_codegen_tests.rs
@@ -80,6 +80,39 @@ fn test_mut_on_local_nested_function_mutborrow_call_argument() {
 }
 
 #[test]
+fn nested_helper_receiver_mutation_uses_lexical_call_metadata() {
+    let rust_code = generate_rust_from_source(
+        r#"
+class Bucket:
+    values: list[int]
+
+    def __init__(self) -> None:
+        self.values = []
+
+    def update(mut self) -> None:
+        def helper(values: list[int]) -> None:
+            values.append(1)
+
+        helper(self.values)
+
+    def size(self) -> int:
+        def helper(values: list[int]) -> int:
+            return len(values)
+
+        return helper(self.values)
+"#,
+    );
+
+    assert!(rust_code.contains("fn update(&mut self)"), "{rust_code}");
+    assert!(
+        rust_code.contains("helper(&mut self.values)"),
+        "{rust_code}"
+    );
+    assert!(rust_code.contains("fn size(&self) -> i64"), "{rust_code}");
+    assert!(rust_code.contains("helper(&self.values)"), "{rust_code}");
+}
+
+#[test]
 fn explicit_class_receiver_signatures_are_emitted_from_hir_metadata() {
     let rust_code = generate_rust_from_source(
         r#"
--- a/crates/sifr_lowering/src/lower/method_receiver_analysis.rs
+++ b/crates/sifr_lowering/src/lower/method_receiver_analysis.rs
@@ -184,17 +184,20 @@ fn collect_method_facts(
                 call_mutates_receiver = true;
             }
         }
-        HirExpr::Call { func, args, .. } => {
+        HirExpr::Call {
+            func,
+            args,
+            mutable_arg_places,
+            ..
+        } => {
             if func == "anext" && args.first().is_some_and(|arg| receiver_rooted(arg, ctx)) {
                 call_mutates_receiver = true;
             }
-            if ctx.functions.get(func).is_some_and(|signature| {
-                args.iter()
-                    .zip(&signature.params)
-                    .any(|(arg, (_, _, convention))| {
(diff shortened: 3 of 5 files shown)


yaseralnajjar:
VERDICT: SATISFIED

## Contract verification

**1. Lexical callable identity (contract items 1 & 3)** — `lower_regular_call` resolves the callee through `ctx.scope.lookup(&func_name)` and, for function bindings, through `ctx.local_function_metadata[binding_id]` (`crates/sifr_lowering/src/lower/expressions/regular_calls.rs:99-142`), i.e. per-binding, not per-name. The conventions used to prove places come from that same lexical signature (`regular_calls.rs:207-234`), so sibling `helper`s in `update` and `size` get independent facts. No basename fallback and no re-registration of nested functions into `ctx.functions` appears anywhere in the diff.

**2. Receiver-rooted marking (contract item 2)** — `method_receiver_analysis.rs:187-203` now consumes `mutable_arg_places` index-aligned with `args`. Alignment is structural: `prove_mutable_arguments` maps over `args` (`method_receiver_places.rs:181-210`), so `places.len() == args.len()` for every call built by `validate_regular_call_arguments`, and `args` is not reordered or re-bound between that call (`regular_calls.rs:523`) and node construction (`regular_calls.rs:756`).

**3. Checked-place soundness** — `target.is_some() && receiver_rooted(arg)` is sound for `Place` targets because `verify_mutable_argument_targets` / `expression_matches_place` (`method_call_verifier.rs:206-280`) rejects any `Some` target that does not structurally match its argument, so a `Some(Place)` at index *i* provably has the same root as `args[i]`. Cross-index contamination is impossible (index-wise `zip`).

**4. No lost module-function signal** — the removed `ctx.functions.get(func)` path is fully subsumed: `bump(self.values)` for a module-level `def bump(mut v: ...)` still goes through `validate_regular_call_arguments`, and `prove_mutable_place` is called with `allow_pending_receiver = true`, so a `BindingKind::Receiver` root yields `Ok(place)` (`method_receiver_places.rs:398-404`). I enumerated every non-test `HirExpr::Call` construction site in `sifr_lowering`: the ones with `mutable_arg_places: Vec::new()` are compiler-synthesized builtins/runtime helpers (`__sifr_task_*`, `__sifr_join_set_new`, `list`, `dict`, `abs`, `sum`, `sorted`, `isinstance`, `Decimal`, `__sifr_python_{present,omitted}_argument`) — none carry `mut`-borrow params, and `register_builtins` only inserts `print` and error/marker classes into `ctx.functions`, so the old lookup never fired for them either. Static/class calls (`Class::method`, `method_argument_ownership.rs:170`) were never matched by the old lookup because `ctx.functions` keys methods with `.`, not `::`.

**5. Nested-body traversal / recursion** — `transform_hir_function` descends into `HirStmt::NestedFunction` (`crates/sifr_ir/src/type_visit.rs:656`), so calls inside nested helper bodies are still visited; a recursive nested call passes the helper's own parameter (not receiver-rooted), so it neither loses nor fabricates a receiver fact.

**6. `anext` and iterator `next`** — the explicit `anext` rule is preserved (`method_receiver_analysis.rs:193-195`) and remains backed by the verifier's `requires_first_mutable_place` invariant (`method_call_verifier.rs:81-91`); the `IteratorCall { Next }` arm is untouched.

**7. Codegen coherence** — plain-call `&mut` emission for arguments flows from the same metadata (`place_emitter.rs:146-159` → `emit_checked_place`), so lowering's receiver decision and codegen's borrow now derive from one authority. The new codegen test asserts both directions (`&mut self.values` in `fn update(&mut self)`, `&self.values` in `fn size(&self)`), which is exactly the divergence the item was filed against.

**8. Regression is not vacuous** — with the reverted hunk, `helper` is absent from `ctx.functions` (Item 10J), `body_contains_receiver_mutation` is false for `update`, and `prove_mutable_place` succeeds under `allow_pending_receiver`, so nothing marks the receiver: `nested_mutating_helper_requires_a_mutable_method_receiver` (`method_receiver_nested_helper_tests.rs:29-51`) would lower successfully and panic. The assertion checks both the diagnostic code and the `self` subject, so it cannot be satisfied by an unrelated error.

**9. Guardrails** — all touched first-party files are well under 900 lines (largest neighbor `method_receiver_analysis_tests.rs` at 875, untouched — putting the new tests in a separate module was the correct split). The e2e fixture needs no sidecar (the pass corpus is `.sifr`-only, discovered by directory scan) and no fixture-count assertion exists.

I read the diff and surrounding code only; I did not re-run the suites, so the numeric validation results are as reported, not independently reproduced here.

## Non-blocking observations

- `method_receiver_analysis.rs:196-200` accepts any `target.is_some()`, which includes `MutableArgumentTarget::OwnedTemporary`. `is_owned_temporary` returns `true` for `HirExpr::Slice` (`method_receiver_places.rs:568`) while `receiver_rooted` descends through `Slice` (`method_receiver_analysis.rs:541-543`), so `helper(self.values[1:3])` against a `mut` param marks the receiver mutable even though the argument is a fresh owned temporary. This is an over-approximation (spurious `mut self` requirement), never an unsound under-approximation, and it is the identical behavior the removed `ctx.functions` path had for module-level helpers — so not a regression. Matching on `Some(MutableArgumentTarget::Place(_))` would be strictly more precise if a later item wants it.
- `collect_method_facts` still has no `HirExpr::GenericCall` arm, so attached-API generic calls (`expressions/attached_api_calls.rs:172`) contribute no receiver-mutation fact. The old `HirExpr::Call`-only arm had the same gap, so this diff neither introduces nor widens it; noted only because it is adjacent to the "lost module signal" question.

yaseralnajjar:
Create-PR gate on exact candidate `60a65c6b909329697fa8e148efe8b96b7774f5e9`: STOPPED only by the recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed, including HIR maintainability, 3,229-file size guardrail, demo freshness, dependency direction, ownership, sysroot/stdlib guards, driver maintainability, and verification self-tests. This gate will not be rerun, per the phase rule and standing authorization.

yaseralnajjar:
Merge gate on exact candidate `60a65c6b909329697fa8e148efe8b96b7774f5e9`: STOPPED only by the same recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed. This is the single merge gate and will not be rerun, per the phase rule and standing authorization.

PR #3478, opened 2026-08-22
Isolate nested try-finally discovery

yaseralnajjar:
Closes phase Item 10O. Enclosing try/finally carrier discovery now treats nested functions as scope boundaries, while nested bodies still discover their own errors when emitted. Infallible try/finally uses unit instead of an undefined Error fallback. Adds sync/async codegen regressions and a native fixture. Exact-SHA Opus review and the single permitted gates will be attached.

### DIFF
--- a/crates/sifr/tests/e2e/pass/try_finally_nested_function_error_discovery.sifr
+++ b/crates/sifr/tests/e2e/pass/try_finally_nested_function_error_discovery.sifr
@@ -1,0 +1,46 @@
+def fail_value() -> Result[int, ValueError]:
+    raise ValueError("local")
+
+
+def run_sync() -> None:
+    try:
+        def inner() -> None:
+            try:
+                try:
+                    _: int = fail_value()
+                    assert False
+                except ValueError as error:
+                    assert error.message == "local"
+            finally:
+                marker: int = 1
+                assert marker == 1
+
+        inner()
+    finally:
+        marker: int = 2
+        assert marker == 2
+
+
+async def run_async() -> None:
+    try:
+        async def inner() -> None:
+            await task.sleep(0.0)
+            try:
+                try:
+                    _: int = fail_value()
+                    assert False
+                except ValueError as error:
+                    assert error.message == "local"
+            finally:
+                marker: int = 1
+                assert marker == 1
+
+        await inner()
+    finally:
+        marker: int = 2
+        assert marker == 2
+
+
+async def main() -> None:
+    run_sync()
+    await run_async()
--- a/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests/sequential_try_binding_codegen_tests.rs
@@ -97,6 +97,58 @@ def run() -> None:
 }
 
 #[test]
+fn nested_function_errors_do_not_select_enclosing_try_finally_carriers() {
+    let generated = generate_rust_from_source(
+        r#"
+def fail_value() -> Result[int, ValueError]:
+    raise ValueError("local")
+
+def run_sync() -> None:
+    try:
+        def inner() -> None:
+            try:
+                try:
+                    _: int = fail_value()
+                except ValueError:
+                    pass
+            finally:
+                marker: int = 1
+
+        inner()
+    finally:
+        marker: int = 2
+
+async def run_async() -> None:
+    try:
+        async def inner() -> None:
+            try:
+                try:
+                    _: int = fail_value()
+                except ValueError:
+                    pass
+            finally:
+                marker: int = 1
+
+        await inner()
+    finally:
+        marker: int = 2
+"#,
+    );
+
+    assert_eq!(
+        generated.matches("Result<(), ()>").count(),
+        2,
+        "{generated}"
+    );
+    assert_eq!(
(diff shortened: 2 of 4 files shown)


yaseralnajjar:
VERDICT: SATISFIED

**Diff verified** (`bf980bd7` → `594286bb`): only 4 files, exactly the two one-line behavior changes plus one focused unit test and one native fixture.

Contract checks:

1. **Scope boundary** — `try_error_helpers.rs:134` now returns `None` for `HirStmt::NestedFunction { .. }`. This is the only recursion-descent change; the sibling arms (`TryExcept` body+handlers, `TryFinally` body+finalbody, `If`/elif/else, `While`/`For`+else, `With`/`AsyncWith`/`AsyncFor`, `Match` arms) are untouched, so enclosing body/finalbody/handler discovery remains intact.
2. **Sync/async parity** — `sifr_ir/src/hir_nodes.rs:524` has a single `NestedFunction` variant carrying `HirFunction` (async is `func.is_async`), so ordinary and async nested defs are governed by the same arm. No name-based carrier was added.
3. **Nested body still discovers its own errors** — `function_emitter/scope_and_function_types.rs:490-494,556-563` `mem::take`s `try_closure_error_type`/`_info`/`depth` and swaps `current_return_type` before lowering `func.body` statement-by-statement. The nested body is therefore the scan root and the emitter state matches the discovery boundary exactly — no special path needed. Confirmed in the fixture: `inner`'s own `TryFinally`→`TryExcept(ValueError)` still yields `Result<(), ValueError>`.
4. **Blast radius** — `first_try_error_type_in_stmts` has exactly one non-recursive caller pair (`loops_try_finally.rs:658-659`). The try/except carrier path uses `select_try_error_type(handlers)` (`loops_try_finally.rs:329` region) and is unaffected.
5. **Priority order and `()` reachability** — `try_finally_error_type_name_for_ir` (`loops_try_finally.rs:648-661`) keeps active try channel → current `Result` error → body → finalbody → `()`. `()` is only reachable when the try channel is empty, the return type is not `Result`, and the body/finalbody contain no `Try*` at all; in that state `can_return_error` (line 670) is false and both `Err` arms (lines 817-824, 843-850) emit `unreachable!` rather than a conversion, so no error-carrying path can select `()`. Note also that a `()` carrier can never enclose a `TryExcept`/`TryFinally` — the presence of one would have supplied a type — so the `?`-conversion target (`stmt_expr_method_and_question_mark.rs:346`) and the Python-context target (`python_context/sync.rs:92`) cannot see `()` on a reachable error path. `assert` lowers to `RustStmt::Assert` (`assert_and_augassign.rs:23`), not an `Err`, so the fixture's asserts inside a `()`-carrier region are sound.
6. **Return capture** unaffected: `capture_returns`/`direct_return_capture`/`ok_ty` (lines 677-696) are independent of the error-name change.
7. **Regression is non-vacuous**: with nested descent restored, the outer carriers become `Result<(), ValueError>`, so `matches("Result<(), ()>").count() == 2` fails (0) and the ValueError count becomes 6. With the `Error` fallback restored, the outer count is also 0. The native fixture emits no `Error` nominal anywhere (no bare `except`, no `Error` usage), so the fallback defect surfaces there as E0425 as claimed.
8. **File-size guardrail**: `loops_try_finally.rs` 856 lines, `try_error_helpers.rs` 176, `sequential_try_binding_codegen_tests.rs` 623 — all under 900.

Non-blocking observations (no action requested):

- `crates/sifr/tests/e2e/pass/try_finally_nested_function_error_discovery.sifr` distinguishes only the fallback defect. Restoring nested-function descent yields `Result<(), ValueError>` outer carriers, which compile and run identically, so the fixture is neutral on the boundary defect; that half of the contract rests entirely on the new carrier-count unit test. Combined coverage is complete.
- `sequential_try_binding_codegen_tests.rs:139-149` asserts whole-file totals (2 and 4), not per-function counts, so a hypothetical sync/async split (e.g. 2/0 vs 1/1) would still pass. Given both functions are structurally identical inputs this is a theoretical gap only.
- Pre-existing and unchanged by this diff: line 652 pattern-matches `Type::Result` on `current_return_type` without `resolve_alias()`, while `can_return_error` at line 671 does resolve aliases. A function returning an *aliased* whole `Result` type would skip the second priority source. That asymmetry predates the diff (it produced the bogus `Error` carrier before); it is not exercised by any fixture (`try_union_error_alias.sifr` aliases only the error type, which matches fine).

yaseralnajjar:
Create-PR gate on exact candidate `594286bb20245629a1d2b3aada98cc16b0d9c5a5`: STOPPED only by the recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed, including HIR maintainability, 3,230-file size guardrail, demo freshness, dependency direction, ownership, sysroot/stdlib guards, driver maintainability, and verification self-tests. This gate will not be rerun, per the phase rule and standing authorization.

yaseralnajjar:
Merge gate on exact candidate `594286bb20245629a1d2b3aada98cc16b0d9c5a5`: STOPPED only by the same recorded linked-delivery A Rust-interop manifest defect: `shared_bridge_crate` references missing evidence source `negative/package_generated_type_import_rejected.sifr`. All reached checks passed. This is the single merge gate and will not be rerun, per the phase rule and standing authorization.

PR #3480, opened 2026-08-22
fix: reconcile shared bridge evidence path

yaseralnajjar:
## Summary
- move the shared-bridge negative evidence source to the canonical flat evidence path
- point the Cargo probe and fixture README at that same source
- keep one source layout with no compatibility reader

## Validation
- `uv run --project verification --locked python -m sifr_verify areas run --area rust_interop --suite matrix`
- `python3 verification/areas/rust_interop/checks/check_fixture_matrix.py --self-test`
- `cargo test -p sifr_driver --lib test_check_shared_bridge_crate_negative_cargo_probe -- --ignored --test-threads=1`
- `git diff --check`

Sifr create-PR and merge gates are not required because compiler implementation files did not change.


yaseralnajjar:
Exact-SHA Opus review for `d9339de46a9aca8c0e6ddd95af807648bd5de398`:

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- Pre-existing record text still names the old `negative/src/...` location; update it in the post-merge issue record.
- Cosmetic README wrapping only.

Review confirmed that `fixture.json`, the only checked-in negative source, and the Cargo-probe `include_str!` all resolve to `negative/package_generated_type_import_rejected.sifr`, with no code or manifest reader accepting the old path.

PR #3483, opened 2026-08-22
fix: emit canonical generated Rust traits

yaseralnajjar:
## Summary
- implement `Default` by delegating to every generated zero-argument `new` constructor
- remove duplicate inherent sync-channel `clone` methods while preserving canonical `Clone` implementations and Sifr call behavior
- refresh all affected checked-in demo companions
- add focused mutation-sensitive codegen coverage

## Validation
- 1,140/1,140 `sifr_codegen` tests pass
- strict `sifr_codegen` Clippy passes
- all seven concurrency generated-code Clippy cases pass under Rust 1.94
- `sync_channel_demo` and `channel_sender_close_clone_closes_all` build and run release-native
- demo freshness passes after updating 32 affected companions
- formatting, HIR maintainability, diff, and file-size checks pass

## Gate provenance
- the single create-PR gate ran on initial candidate `9453b3780e971e21aca14c2316457edb34c9bb50` and stopped only on the 32 stale demo companions
- those in-scope generated outputs are refreshed in final candidate `3d9f26a9b3018957b5f09b4c275e444d815af076`
- per the phase rule, the create-PR gate will not be repeated; the one merge gate will run on the final reviewed candidate

### DIFF
--- a/crates/sifr_codegen/src/class_emitter.rs
+++ b/crates/sifr_codegen/src/class_emitter.rs
@@ -691,12 +691,42 @@ impl RustEmitter {
         }
         self.current_class_name = saved_class_name;
         if !constructor_items.is_empty() {
+            let constructor_type_params =
+                Self::class_constructor_impl_type_params(class, &constructor_items);
+            let has_zero_argument_constructor = constructor_items.iter().any(|item| {
+                matches!(
+                    item,
+                    RustItem::Fn { name, params, .. } if name == "new" && params.is_empty()
+                )
+            });
             self.body_items.push(RustItem::Impl {
                 target: Self::class_impl_target(class),
-                type_params: Self::class_constructor_impl_type_params(class, &constructor_items),
+                type_params: constructor_type_params.clone(),
                 trait_: None,
                 items: constructor_items,
             });
+            if has_zero_argument_constructor {
+                self.body_items.push(RustItem::Impl {
+                    target: Self::class_impl_target(class),
+                    type_params: constructor_type_params,
+                    trait_: Some("std::default::Default".to_string()),
+                    items: vec![RustItem::Fn {
+                        name: "default".to_string(),
+                        visibility: Visibility::Private,
+                        type_params: Vec::new(),
+                        params: Vec::new(),
+                        ret: Some(RustType::Named("Self".to_string())),
+                        body: vec![RustStmt::Return(Some(RustExpr::FnCall {
+                            func: Box::new(RustExpr::Path(vec![
+                                "Self".to_string(),
+                                "new".to_string(),
+                            ])),
+                            args: Vec::new(),
+                        }))],
+                        is_async: false,
+                    }],
+                });
+            }
         }
         if method_items.is_empty() && class.type_params.is_empty() {
             self.body_items.push(RustItem::Impl {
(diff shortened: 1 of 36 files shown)


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

**Evidence checked (read-only, against the exact `c2dffcc81..3d9f26a9b` diff):**

- `class_emitter.rs:694-729` emits exactly one `Default` impl per class that already has one zero-argument `new`, delegating to `Self::new()`. It reuses the same `constructor_type_params` as the inherent constructor impl, so generic classes get matching bounds, and it does not add a second constructor. Rendered output confirms the canonical path: `impl ::std::default::Default for X { fn default() -> Self { Self::new() } }`.
- No competing `Default` source exists — `Default` is not in the derive list built at `class_emitter.rs:614-633`, and no other emitter emits a `Default` impl for user/stdlib classes, so no conflicting-impl risk.
- `publicize_generated_module_source` (`lib_runtime_needs.rs:867-869`) skips trait impls, so `default` is correctly not publicized.
- `lib_runtime_needs.rs`: the two removed inherent methods were byte-for-byte equivalent to the retained trait impls (`Channel::clone` was `Clone::clone(self)`; `ChannelSender::clone` was `ChannelSender::new(self._channel.clone())`, identical to `impl Clone for ChannelSender` at line 215-219). Sender registration via `ChannelSender::new` → `register_sender` is preserved, so close/clone refcount semantics are unchanged. One `Clone` impl each remains for `Channel` and `ChannelSender`; `ChannelReceiver` is intentionally non-cloneable. No emitter site references `Channel::clone`/`ChannelSender::clone` by path, so call sites resolve through the prelude `Clone`.
- No `allow` attribute was added anywhere in the diff; no Sifr API was renamed.
- Demo companions: the aggregate demo diff is purely additive and contains only `impl ::std::default::Default … { fn default() -> Self { Self::new() } }` blocks (92 impls, 184 braces, zero deletions). `demos/sync_channel_demo/` has no checked-in `emitted.rs`, and no companion contained the removed inherent clones, so the runtime deletion leaves nothing stale.
- File-size guardrail: `class_emitter.rs` 878 lines, `lib_runtime_needs.rs` 891 lines — both under 900.
- Gate provenance treated as stated: the single create-PR gate on `9453b3780` **stopped** (did not pass) at demo freshness; it is not counted as a pass here. The source files are identical between `9453b3780` and the final candidate — only demo companions differ — so the setup/HIR/file-size portions it did clear still apply, and the remaining coverage rests on the focused evidence listed.

Follow-up findings:
- suggestion: `class_emitter.rs:696-701` gates the `Default` emission on `name == "new" && params.is_empty()` only; it ignores the constructor's `ret` and `is_async`. Every zero-argument constructor in the current corpus is `fn new() -> Self` (verified across all `demos/*/emitted.rs`), so this is not reachable today, but a future fallible or async zero-argument constructor would emit a `default() -> Self { Self::new() }` that does not type-check. Tightening the match to also require `ret == Some(Named("Self"))` and `!is_async` would make the guard self-enforcing.
- suggestion: `class_trait_contract_codegen_tests.rs:33-34` asserts the absence of `fn clone(&self) -> Channel<T>` / `-> ChannelSender<T>`. A reintroduced inherent `fn clone(&self) -> Self` would slip past both negatives; asserting a single total `fn clone` occurrence per type would be a tighter regression lock.
- infrastructure issue: `cargo clippy -p sifr_codegen` lints the compiler, not its output. Generated-code lint coverage for the new `Default` impls comes only indirectly through the concurrency group and e2e compilation; a generated-code Clippy entry containing a zero-argument class would pin the `new_without_default` fix that motivates this item.

PR #3487, opened 2026-08-22
test: lock pre-v1 regression closure inputs

yaseralnajjar:
## Summary

- add a static guard that locks the 15 original failed area variants to their exact owners
- reject stale Rust-interop and fuzz paths plus stale performance and ecosystem hashes
- keep nominal-identity, optional-conversion, and nested try/finally fixtures in the create-PR E2E subset
- retain the existing no-pre-v1 compatibility guard unchanged

## Validation

- pre-v1 regression closure guard and mutation self-test
- no-pre-v1 compatibility guard and mutation self-test
- developer tooling static suite: 16/16 variants
- create-PR E2E fixture lane: 143/143 fixtures
- cargo fmt --check
- Ruff on changed Python files

No compiler file changed, so the user-authorized Item 12 rule excludes Sifr create-PR and merge gates. The phase reuses Item 10P’s exact final implementation gate.


yaseralnajjar:
# SATISFIED

## Scope and gate applicability
`git diff f97fa7d5f..8be98bb04` touches exactly 4 files, all verification code/data (`developer_tooling/check_pre_v1_regression_closure.py`, `developer_tooling/pre_v1_regression_closure.json`, `developer_tooling/runner.py`, `core_language/data/create_pr_e2e_manifest.json`). No compiler file, no `crates/**`, no `scripts/**`. I found no uncovered mechanism that would require a compiler gate, so the user's ruling holds: Item 10P's gate (698 E2E pass fixtures, per the Item 11 record) remains the valid final implementation evidence. Working tree clean after review (the runner only wrote `target/verification/...`).

## Mechanical reconciliation of the 15 variants
- `EXPECTED_AREA_COUNTS` = eco 5 + fuzz 1 + generated_code_quality 2 + performance 6 + rust_interop 1 = **15** across **5** areas, matching the record's rebaseline row (`421 passed, 15 failed`, 5 failed areas) at base `b52fbc7e…` (record line 37).
- Every ID resolves to a real suite/variant emitter: `matrix/area-check` (`area_adapter.py:44`), `representative/*` (`performance/runner.py:127,164,178,223`), `corpus`/`panic-scan` (`generated_code_quality/runner.py:26`), `fuzz-smoke/*` (`property_and_fuzz.py:306`), `OSS-CURATED-000N/source-checksum` (`oss_and_determinism.py:183`).
- Owners map to actual item headings: Item 2 = canonical project root (line 416, 4 variants), Item 3 = evidence freshness (line 469, 8 variants), Item 6 = project-wide nominal identity (line 710, 2), linked delivery A + Item 10 (line 996, 1). Placeholder owners rejected via hyphen-segment match.
- Double-entry lock: JSON must equal the Python constants, so silent drift in either direction fails.

## Guard mechanisms verified
- **Stale Rust-interop path**: requires flat `negative/package_generated_type_import_rejected.sifr`, rejects the restored `negative/src/…` file *and* any `negative/src/…` reference in `fixture.json`.
- **Stale fuzz path**: exact `seed_files == [canonical]`, containment in `reproduction_command`, whole-manifest scan for the pre-`src` variant, plus on-disk seed existence.
- **Performance hash**: recomputes `sha256(benchmark_manifest.json)` bytes against `trend/current.json:manifest_sha256` — identical semantics to the owning check (`check_trend_policy.py:258`), so no competing definition was introduced.
- **Ecosystem freshness**: recomputes all 5 checksums through the owner's own `project_source_checksum`, and treats the `""` failure return as stale rather than passing.
- **No-pre-v1 compatibility guard**: untouched by the diff and still passing in the suite run below.
- **Enforcement path**: `developer_tooling` `static` suite is selected by both `create-pr.json` and `merge.json`, so the guard is gated, not orphaned. `area_developer_tooling` budget is 180 s; the two new variants cost 312 ms.

## Create-PR subset
143 unique names, all resolving to existing `crates/sifr/tests/e2e/pass/*.sifr`. The harness loads them into a `BTreeSet` (`fixture_compilation.rs:110`), so insertion position is immaterial, and the base list was also unsorted — no ordering contract broken. No hardcoded fixture count anywhere (`profiles.py:86` computes it). All three locked fixtures already passed in Item 10P's full 698-fixture lane. `nominal_identity_alias_paths.sifr` genuinely exercises aliased/imported nominal identity and inherited stdlib classes; `nested_optional_safe_operations.sifr` exercises optional/union conversion across `dict.get/pop`, `list.pop`, `next()`, and indexing.

## try_nested_finally / twin decision
Including `try_nested_finally_error_propagation` is correct: it is Item 10H's own native fixture (added in `37e70db4`, merged `15334528`) and the direct runtime detector of the fixed defect. The no-twin decision is also justified: Item 10H changed `try_closure_depth > 0` → `!self.try_closure_error_type.is_empty()` (`loops_try_finally.rs:672`). `try_closure_depth` increments only when `capture_returns` is true, while `try_closure_error_type` is pushed unconditionally — so the new predicate is a strict superset and the return-capturing path is bit-identical before and after the fix. A twin would have zero mutation-detecting power for this defect.

## Ancillary checks run
`ruff check` clean; `verification_taxonomy.py` passes (new files use no banned delivery terms); `check_hir_maintainability_guardrails.py` passes; `check_file_size_guardrails.py` passes (3232 files, new file 311 lines); developer tooling static suite `variants=16, failures=0`; guard and self-test both exit 0. Record integrity: 31 `State: complete` items, zero `blocked`, no unowned/TBD failure rows.

## Non-blocking observations
1. **Mislabeled twin anchor.** `item_10h_return_capture_twin_decision` asserts the presence of `test_try_finally_runs_cleanup_before_timeout_propagates` (`async_runtime_codegen_tests.rs:65`), which predates Item 10H by ~3 months and whose inner `try` body contains no `return` — so `capture_returns` is false and it is not return-capture coverage. The actual Item 10H regression test, `nested_try_finally_uses_enclosing_try_error_channel_without_return_capture` (`sequential_try_binding_codegen_tests.rs:80`), is not locked. Consider anchoring there, or renaming the key. The decision itself stands.
2. **Self-test breadth.** 6 of ~14 rejection branches are mutated. Unmutated: `base_sha` drift, area-count drift, placeholder owner, the `fixture.json` stale-path branch, ecosystem entry count, duplicate create-PR names, and twin-decision drift. Related: the `fixture.json` write inside the Rust-interop mutation is dead code — the duplicate-path check raises first.
3. **Self-test robustness.** `next(raw for raw in stale_fuzz["targets"] …)` raises an uncaught `StopIteration` rather than `GuardError` if that target ever disappears, producing a traceback instead of a clean FAIL line.
4. **Item 2 has no independent lock.** The guard duplicates the trend-hash and ecosystem-checksum checks but adds nothing for the canonical LSP project root (4 of the 15 variants). Acceptable as-is: `benchmark_manifest.py:167-181` still enforces `project_root/sifr.toml` and `project_root/src/main.sifr`.
5. **Record closure pending.** The Item 12 record, Phase Exit State, and any roadmap status update are absent from this candidate — consistent with every prior item's implementation/record split (e.g. `37e70db4` impl vs `b71d3a35` record), so this is expected follow-up work rather than a defect in the candidate.

PR #3491, opened 2026-08-23
build: pin Rust 1.98 toolchain

yaseralnajjar:
## Summary
- pin local and CI Rust selection to 1.98.0 while retaining edition 2021
- raise the latest-only workspace compiler requirement from 1.93 to 1.98
- update current architecture, package audit, runtime-platform, and dependency decision records
- resolve Rust 1.98 Clippy diagnostics with semantic-preserving rewrites
- add UTF-16 unit/remainder regression coverage

## Focused validation
- `rustc --version --verbose` / `cargo --version` (1.98.0)
- `cargo metadata --locked --offline --no-deps --format-version 1`
- `cargo check --workspace --locked --offline`
- `cargo test -p sifr_runtime -p sifr_type_system -p sifr_lowering -p sifr_codegen -p sifr_lint -p sifr_package --locked --offline`
- `cargo fmt --check`
- runtime-platform `platform-support-matrix`: 4/4
- distribution-release `full`: 67/67
- HIR maintainability guardrails
- first-party file-size guardrails: 3,232 files
- `git diff --check`

The exact final-SHA create-PR and merge gates will run once each after Opus approval, per the phase contract.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- infrastructure issue: `.github/workflows/preview-release.yml:159` previously pinned `dtolnay/rust-toolchain` to an immutable SHA (`4cda84d…`) and now uses the mutable `@1.98.0` tag. This is the correct trade for exact-version convergence and is explicitly deferred to Item 6, but that one workflow temporarily loses SHA immutability — Item 6 should restore it with a SHA that resolves to 1.98.0.
- suggestion: `crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs:448` — `sort_by_key(|(left, _)| (*left).clone())` allocates a `String` per key extraction where the previous `sort_by` compared borrowed keys. Ordering is identical and the map is small, so this is cosmetic cost only; `sort_unstable_by_key` or keeping a comparator is an option if this ever becomes hot.
- pre-existing issue: `verification/areas/runtime_platform/sanitizer_manifest.json` still declares `required_rustup_toolchains: ["nightly"]` for sanitizer cases. That is intentional (sanitizers need nightly) and is not a maintained-host toolchain row, so it does not conflict with the latest-only policy — but the policy docs do not say so explicitly.

Verification notes behind the verdict:
- Exactly six first-party `dtolnay/rust-toolchain` installs exist (4 in `local-first-validation.yml`, 1 each in `preview-release.yml` and `release-qualification.yml`); all read `@1.98.0`. Remaining hits are vendored crates' own CI. The other four workflows install no Rust toolchain.
- `rust-toolchain.toml` selects `1.98.0` with `clippy` + `rustfmt`; `Cargo.toml` sets `rust-version = "1.98"`, `edition = "2021"` unchanged, and every crate inherits `rust-version.workspace`. No `Cargo.lock` or dependency-version change is in the diff.
- All six maintained rows in `supported_platforms.json` moved `stable` → `1.98.0`; no `1.93`/floor policy statement survives in `README.md`, `internal_docs/`, `docs/`, or the audit reports, while the performance trend capture is untouched.
- Each source edit is semantics-preserving: the guard rewrites (`queries_impl.rs:381`, `imported_class_identity.rs:512`, `match_lowering.rs:196`, `method_receiver_analysis.rs:205`) all fall through to a no-op `_ => {}` with no later arm able to match, the `?` rewrites sit in `Option`-returning function scope, `is_ok_and`/`.cloned()?`/`.zip(x)`/pattern-field-removal are direct equivalences, and `as_chunks::<2>()` returns `(chunks, remainder)` matching the prior `chunks_exact(2)` + `remainder()` behavior — covered by the new complete-unit, strict trailing-byte, and replace trailing-byte assertions in `crates/sifr_runtime/src/encoding.rs:503`.

yaseralnajjar:
Final exact-SHA validation evidence for cf9702fabbf9230a13daaeff3e88609aacb0f73d:

- The sole create-PR gate attempt exposed an environment prerequisite: generated projects outside the repository resolved the host rustup default (1.94.0) instead of the repository-local 1.98.0 override. No candidate file changed.
- Updated the host stable toolchain to 1.98.0, then reran all 10 affected Python-interop variants directly: 10/10 passed with zero failures.
- The sole merge gate completed on the unchanged reviewed SHA with exit code 0. All lanes passed, including 698/698 E2E fixtures. Report: target/validation_lane_reports/merge.latest.json.
- The 6508.96s cold run exceeded the advisory warm-time budget; this was non-blocking and attributable to cold generated-artifact caches (0/173 E2E cache hits).

PR #3499, opened 2026-08-24
chore: converge on Node 24 LTS

yaseralnajjar:
## Scope

Item 5 of the latest-stable convergence phase.

- select exact Node.js 24.19.0 Krypton LTS and bundled npm 11.17.0
- make extension CI, stable qualification, and stable publication consume one checked-in `.node-version`
- adopt npm 11 `devEngines` to reject runtime and package-manager drift
- install exact dev dependencies without lifecycle scripts in all maintained workflows
- advance the sifr-vscode -> editor-integrations -> root pointer chain sequentially

Leaf PRs:
- sifr-lang/sifr-vscode#13 (`7bf12ce026dcbe13679211a06caa76a61e84760a`)
- sifr-lang/editor-integrations#11 (`b42360d0bbc99c45f625db67ff9a1cb4afdfeaa1`)

## Validation

- official Node archive checksum and exact Node/npm probes
- extension npm lock regeneration, clean install, lint, typecheck, unit, extension smoke, and VSIX packaging
- negative npm `devEngines` rejection on Node 24.16.0/npm 11.13.0
- Node toolchain and release workflow contracts
- distribution-release representative 57/57
- developer-tooling rules/package checks
- Bash syntax, diff check, and 900-line file-size guardrail

No compiler input changed, so the phase rules prohibit Sifr create-PR and merge gates.


yaseralnajjar:
## Exact-SHA Opus review — SATISFIED

Candidate: `e9225ac561fc27a65c8cd8d5f4619b8e8ab3df06`

Reviewed the exact root diff `f4da1f0dc..e9225ac56`, the editor-integrations pointer `d7577d49..b42360d0`, and the leaf `273fd5d3..7bf12ce0` (including the lock diff), plus the surrounding call sites I could not confirm from the evidence alone: every `setup-node`/`npm` site in root workflows, step-ordering of the `stable-source` checkout relative to `setup-node`, case discovery for the new contract, and the leaf's own toolchain files.

Independent checks beyond the supplied evidence:
- `.github/workflows/*.yml`: `setup-node` occurs exactly twice repo-wide (publication:196, qualification:161), both now version-file driven; `node-version-file:` count is 1 per file; no other workflow invokes `npm`/`vsce`, so no third consumer was left on Node 22.
- Publication: the `stable-source` recursive checkout (`release-publication.yml:166-172`) carries the identical `env.STABLE_CANDIDATE_OPERATION == 'true'` guard as the `setup-node` step, so `stable-source/editor_integrations/vscode/.node-version` always exists when the file is read.
- New case is picked up without registration: `verification/areas/distribution_release/manifest.json` uses the `distribution-case-directory` adapter over the `cases` dir for both `representative` and `full`.
- Leaf has no competing selector (`.nvmrc` absent), `.node-version` is exactly `24.19.0\n`, and `engines`/`devEngines`/`packageManager`/lock root engines agree pairwise as the new contract asserts.
- Chain is sequential and pointer-only: leaf `09ff33c`→`7bf12ce`, pointer `a174b94`→`b42360d` bumps only `vscode`, root bumps only the `editor_integrations` gitlink.
- Docs: no `Node 22` or bare `npm ci` claim survives in `internal_docs/`; remaining hits are archived review records and plan history, which are not contracts.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `verification/areas/distribution_release/cases/node_toolchain_contract.sh:22` pins Node to the `24.x` major via regex but constrains npm only to an exact `X.Y.Z`, so an npm-major selector drift (e.g. `npm@12.0.0`) would pass this contract. It is fail-closed downstream — `devEngines.onFail=error` would reject the Node-24-bundled npm on every install — but the contract could assert `11\.[0-9]+\.[0-9]+` and make the Node/npm pairing intent explicit at the check rather than at first CI failure.
- suggestion: `demos/editor_candidate_qualification_demo.sh:69` now runs an install governed by `devEngines.onFail=error` with no toolchain selection step, so the demo hard-fails on any host whose ambient Node/npm is not exactly 24.19.0/11.17.0. That is the intended enforcement, not a defect, but the demo is the one in-scope `npm ci` site with no `.node-version` consumer; a one-line note or version assertion there would turn `EBADDEVENGINES` into an actionable message.
- infrastructure issue: `node_toolchain_contract.sh:57` is the first `distribution_release` case to read files inside the nested submodule (the other two contracts only match those paths as literal strings in root YAML). Without a recursive checkout it aborts with an unhandled `FileNotFoundError` instead of a named failure. Not reachable today — every lane in `.github/workflows/local-first-validation.yml` checks out `submodules: recursive` — so this is robustness only.
- pre-existing issue: leaf `package.json:220` keeps `@types/node: ^22.10.5` against a Node 24 runtime selection, and npm still reports three high-severity advisories. Both are explicitly Item 33's scope and this diff changed no dependency versions.
- pre-existing issue: `release-qualification.yml:161` still uses the mutable `actions/setup-node@v4` tag while publication uses the pinned SHA. Item 6 owns action-version/immutable-SHA convergence; correctly untouched here.
- pre-existing issue: no Sifr compiler input changed (the diff touches only workflows, verification scripts, docs, the demo, and gitlinks), so the create-PR and merge gates were correctly not run for this item.

PR #3501, opened 2026-08-24
ci: pin latest stable GitHub Actions

yaseralnajjar:
## Scope

Item 6 of the latest-stable convergence phase.

- pin every maintained external action to the exact commit for its latest stable release
- advance checkout 7.0.1, setup-node 7.0.0, upload-artifact 7.0.1, and download-artifact 8.0.1
- retain current setup-uv 10.0.1 and bind the Rust 1.98 toolchain action to its exact commit
- use download-artifact v8 fail-closed digest mismatch handling
- add one policy and checker for all root and extension workflow references
- repair commented-SHA recognition in the submodule checkout guardrail
- advance the sifr-vscode -> editor-integrations -> root pointer chain sequentially

Leaf PRs:
- sifr-lang/sifr-vscode#14 (`732bcdc3ae2a494753025710dd138aa23a39b6e4`)
- sifr-lang/editor-integrations#12 (`d202b8c60240b6d2897c9deeda59be899bf47e24`)

## Validation

- official GitHub release/tag refs independently resolved for all six actions
- immutable-pin policy and negative self-test: 6 actions / 56 references
- submodule ownership guardrail and negative self-test
- all workflow YAML parses
- download-artifact v8 exact metadata confirms Node 24 and digest-mismatch:error
- distribution-release representative 58/58 on final state
- verification runner self-test, profile check, and area check
- Ruff 0.16.4 check/format, Bash syntax, diff check, and 900-line guardrail

No compiler input changed, so the phase rules prohibit Sifr create-PR and merge gates.

### DIFF
--- a/.github/workflows/local-first-validation.yml
+++ b/.github/workflows/local-first-validation.yml
@@ -12,11 +12,11 @@ jobs:
     name: lint-and-format
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
           submodules: recursive
-      - uses: dtolnay/rust-toolchain@1.98.0
+      - uses: dtolnay/rust-toolchain@f8be11a05b1d4f3fcebe6410cc16743212b999b0 # Rust 1.98.0
       # Advisory until legacy workspace lint debt is burned down by a dedicated cleanup.
       - name: Rustfmt check (advisory)
         continue-on-error: true
@@ -35,11 +35,11 @@ jobs:
       matrix:
         profile: [create-pr, merge, release]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
           submodules: recursive
-      - uses: dtolnay/rust-toolchain@1.98.0
+      - uses: dtolnay/rust-toolchain@f8be11a05b1d4f3fcebe6410cc16743212b999b0 # Rust 1.98.0
       - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
           version-file: verification/pyproject.toml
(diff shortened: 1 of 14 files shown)


yaseralnajjar:
## Exact-SHA Opus review — SATISFIED

Candidate: `c08ac5567d6acb64bbc76c731a6471806bc46b3a`

I inspected the exact root diff (`e73fd79f..c08ac556`), both nested submodule diffs, and re-ran the two guardrails locally.

**Independent checks I ran (not duplicating supplied evidence):**
- Root workflows contain exactly 54 external `uses:` refs, all six actions at the stated SHAs with exact labels; leaf `ci.yml` adds 2 → matches the reported 56. No old SHA (`11d5960a`, `ea165f8d`, `d3f86a10`, `49933ea5`) or mutable tag survives in any maintained workflow.
- Policy JSON entries match all six rechecked resolutions character-for-character.
- YAML-parsed every root workflow: all 10 `download-artifact` steps carry `digest-mismatch: error` in their own `with:` block.
- Submodule chain converges: leaf `732bcdc3`, pointer `d202b8c6` (bumps only `vscode`), root gitlink `d202b8c6`.
- No `.github/actions/**`, no `editor_integrations/.github`, no corpora workflows — root + leaf is the complete maintained set.
- New case `github_action_pins.sh` is executable, sources the existing `common.sh`, and is auto-discovered by the `distribution-case-directory` adapter (no manifest edit needed).
- Both updated contract-test literals match reality (4 upload steps in `release-qualification.yml`; the setup-node SHA fragment is present in `release-publication.yml`).

---

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `scripts/check_submodule_ownership.py:83` — `WORKFLOW_STEP_RE`'s `^(\s*)-\s+...(?=^\1-\s+|\Z)` backreference makes a lone list item swallow the remainder of a file. In `.github/workflows/release-publication.yml` the enum item at line 34 (indent 10) has no sibling at that indent, so lines 34–901 collapse into one pseudo-step: the guardrail detects **0** checkout blocks in that workflow and passes only because the single `submodules: recursive` at line 170 falls inside the swallowed body. Consequently the four checkout steps at lines 153, 159, 172, 179 have no recursive submodule init and are invisible to the guardrail. This bypass pre-dates the candidate (previously those steps were skipped by the comment-blind `CHECKOUT_USES_RE` instead), so the candidate causes no coverage loss — but the summary's "seven formerly skipped checkout steps now initialize recursive submodules" overstates closure: `release-publication.yml` remains wholly unguarded, and the new negative self-test only covers the comment bypass, not the step-splitting one. Worth an anchored-step-boundary fix (or YAML-parsed steps) in its own item.
- suggestion: adding `submodules: recursive` to the evidence-only checkouts (`release-publication-prepare.yml` `stable-evidence`/`incident-evidence`, `release-publication-drill.yml`, `schema-bootstrap-recovery.yml`) was forced by the restored guardrail coverage, not by the pinning criteria. It now pulls the `third_party/ruff` fork and all corpora into governance jobs that read only JSON evidence — a real runtime cost and a new failure surface on historical evidence commits whose gitlinks may no longer be fetchable. Consider whether the guardrail should permit an explicitly justified no-submodule checkout class.
- suggestion: `third_party/ruff` is a sifr-lang-owned fork with its own workflows but is outside both the stated audit scope and the repo-wide `third_party/**` exclusion. If its CI actually runs on the maintenance branch, its action pins are unowned by `verification/policy/github_actions.json`; record separately rather than widening this item.

PR #3505, opened 2026-08-24
chore(deps): update utility foundation train

yaseralnajjar:
## Summary
- update the Item 8 Rust utility/foundation dependency train to the audited latest stable releases
- reconcile root, generated, fixture, and package-owned lock identities and refresh the scoped sysroot vendor set
- bind source-tree generated-project cache identity to the live Cargo.lock SHA-256 so dependency updates cannot reuse stale locks
- adopt current dependency behavior without compatibility paths or fallbacks

## Validation
- `cargo metadata --locked --format-version 1`
- offline sysroot and direct-vendor Cargo probes
- `cargo test -p sifr_lint -p sifr_stdlib -p sifr_stdlib_manifest -p sifr_diagnostics -p sifr_codegen -p sifr_lsp -p sifr_driver -p sifr -- --skip test_e2e_pass`
- Rust interop area: 10/10 variants
- stdlib parity: 411/411 LeetCode and all module/dependency checks; known out-of-scope `m16_raw_api` Python requirement failure remains owned by pre-v1 compatibility Item 13
- `cargo fmt --check`

Candidate: `53320be497f298800c8f1d9001c22faa7a6acbe4`


yaseralnajjar:
Exact-SHA Opus review for candidate `53320be497f298800c8f1d9001c22faa7a6acbe4` (base `e747a20f06584d184d95bac10ee842136eea391c`):

I reviewed the exact diff `e747a20f0..53320be49` (925 changed paths) rather than re-running the broad gates. What I independently verified:

- **Version propagation**: every one of the 34 train identities resolves to the stated target in the root `Cargo.lock`; no active manifest, catalog, generated literal, checker, fixture manifest, evidence marker, README, doc, or report retains a superseded literal (remaining old-version hits are unrelated transitive packages such as `autocfg 1.5.0`, `atomic-waker 1.1.2`, `toml_writer 1.1.2+spec-1.1.0`, plus archived plan records). `toml` resolves to `1.1.4+spec-1.1.0`.
- **Lock/root identity**: `cargo metadata --locked --offline` passes; all 22 rust-interop fixture locks plus the core-language fixture lock were re-diffed against the root lock — every sourced entry in every interop fixture lock is present in the root graph (0 drifted). The `_scenario_checks.py` self-test mutation was correctly re-pointed (`2.8.3` → `99.99.99`, unique in that lock, absent from root).
- **Vendor boundary**: exactly one new vendor package (`syn-3.0.3`, required by `serde_derive 1.0.229`); no interop-only graph (blake3, cxx, crc32fast graphs) was absorbed. I recomputed every file checksum in all 43 touched vendor packages against `.cargo-checksum.json` — 0 mismatches, 0 unlisted extras (including the `libc` `nto/arch/*` renames), and each `package` checksum matches the root lock. All direct dependencies of train packages, and all exact `=` pins emitted by `retained_dependency_specs`, exist in vendor at satisfying versions. `blake3 1.8.7`'s build script still emits only names inside the existing 7-entry native-link envelope.
- **Snapshots**: I regenerated all five `cargo tree` feature snapshots with the documented commands — byte-identical to the checked-in files (the large structural deltas are genuine refreshes of previously stale evidence, including `hashbrown 0.17.1` and the `static_assertions` dev-dep edge).
- **Live-lock cache identity**: `ResolvedSysroot::from_root` computes the digest after `paths.validate` guarantees `Cargo.lock` exists, fails closed as `MissingAsset` otherwise, and feeds `cache_fingerprint`, which flows into `binary_project_cache_key` and thus into the cache-key-addressed artifact directory — so a lock change forces a fresh workspace and a fresh generated lock. This is a root-cause fix: the source-tree `sysroot.toml` carries all-zero `cargo-lock-sha256`/`sysroot-content-sha256` placeholders, which is exactly why the pre-existing fields never invalidated. All three test constructors were updated; no non-test literal constructor exists; `sha256_file` duplicates no existing helper; touched files stay far under 900 lines.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `verification/areas/rust_interop/fixtures/native_build_script/examples/native_trust_package/rust/cxx/src/lib.rs:9` still returns `1_000_198`, so the emitted evidence string reads `cxx=1.0.199;bridge=sifr_cxx_probe;value=1000198` (asserted verbatim at `crates/sifr_driver/src/tests/package_rust_interop_native_build_support.rs:80`). The constant transparently encodes 1.0.198 but is not documented or checker-bound as a version claim, so it is not an identity literal — still, aligning it to `1_000_199` would remove a misleading stale-version reading inside a version evidence marker. This is the same drift-hardening gap already recorded in `plans/reviews/active/rust-interop-certification-9-review-round-1.md:37`.
- pre-existing issue: `vendor/futures-core|futures-channel|futures-sink` are 0.3.32 and `vendor/hashbrown` is 0.17.0 while the root lock has 0.3.33 and 0.17.1 (unchanged base→candidate). Harmless because generated projects resolve fresh against the vendor directory source rather than the root lock, but it means vendor is not a mirror of the root graph — worth stating explicitly wherever the vendor boundary is documented.
- pre-existing issue: `verification/areas/core_language/fixtures/static_class_adapter/negative/non_string_leaf_rejected/Cargo.lock` carries `autocfg 1.5.1`, `num-bigint 0.4.8`, `num-integer 0.1.47`, none present in the root lock. The candidate actually reduced this drift (its `libc` now matches at 0.2.189), but unlike the rust_interop area no subset rule guards core_language fixture locks.
- infrastructure issue: `vendor/syn-3.0.0/` exists as an empty, untracked directory in this worktree. It is invisible to git and inert to Cargo (no manifest), but it is leftover cruft from the vendor swap and would break vendor resolution if it ever gained a manifest.
- suggestion: `annotate-snippets 0.12.16` and `cookie 0.18.2` are declared only in `[workspace.dependencies]` (`Cargo.toml:79,122`) and consumed by no crate, so they never enter `Cargo.lock` and no gate validates those two bumps. Either wire them to an owner or drop them from the workspace table.
- suggestion: the tracking row at `plans/issues/active/ad-hoc-latest-stable-release-convergence.md:98` still lists the pre-Item-8 versions as current; per the repository's record rule that update lands after merge, not in this diff.

yaseralnajjar:
Gate evidence for exact candidate `53320be497f298800c8f1d9001c22faa7a6acbe4`

- Exact-SHA Opus review: **SATISFIED**, no blocking findings ([review comment](https://github.com/sifr-lang/sifr/pull/3505#issuecomment-5393608119)).
- One create-PR gate invocation: all executed functional checks passed; the invocation exited `124` only because `area_runtime_platform` took `203.977s` against its `120s` timing budget. The cold generated-project case `binary_file_io_capability.sifr` accounted for `159.454s`. Report: `target/validation_lane_reports/create-pr.latest.json`.
- One merge gate invocation: **PASS** (`exit 0`). The same runtime-platform case completed in `3.656s` once materialized, confirming the create-PR timing result was cold-cache infrastructure rather than a mechanism defect. Full profile passed all areas, all crate suites, the 76 generated-build driver integrations, and `698/698` E2E pass fixtures across 173 groups. No swaps; peak RSS `2.1GiB`. Report: `target/validation_lane_reports/merge.latest.json`.
- The merge run's warm wall-time target remained advisory and was exceeded because the required pre-gate cleanup left `0/173` E2E cache hits; the canonical merge gate nevertheless completed successfully.

Per the phase rule, neither gate was rerun.

PR #3511, opened 2026-08-24
Upgrade Rust TLS stack to latest stable releases

yaseralnajjar:
## Summary
- upgrade Rustls to 0.23.43 and rcgen to 0.14.9
- adopt bounded RFC 9149 TLS 1.3 ticket requests and canonical rcgen BasicConstraints DER
- refresh package-owned vendors, independent Rust-interop locks, feature snapshots, and traceability evidence

## Validation
- targeted Rustls and rcgen runtime tests
- sifr_runtime and sifr_stdlib all-feature suites
- canonical Rust interop area plus one native scenario per changed independent lock
- public Sifr TLS loopback and config-error fixtures
- workspace Clippy, rustfmt, maintainability, stdlib audit fixtures, file-size, vendor checksum, and diff checks

Exact candidate: f432ab8d88d44b21bd193bca66d5deaa38ac1102


yaseralnajjar:
# Exact-SHA Implementation Review — Latest-Stable Convergence Item 11 (Rust TLS Stack)

## 1. SHA confirmation

| Role | SHA | Subject |
| --- | --- | --- |
| Base | `ebace873a977ff8e82bde5d55d06e3865e64b288` | Merge pull request #3510 from sifr-lang/codex/latest-stable-upgrades-item-10-record |
| Candidate | `f432ab8d88d44b21bd193bca66d5deaa38ac1102` | Upgrade Rust TLS stack |

Diff scope reviewed: `base..candidate` only, 57 files, +1121 / −670. Working tree clean at review time; no files, branches, index, or external state were modified.

## 2. Inspected scope

Production manifests and locks
- `Cargo.toml:118` `rcgen = { version = "0.14.9" }`; `Cargo.toml:125` `rustls = { version = "=0.23.43" }`.
- `Cargo.lock` changed for exactly two packages (rustls, rcgen); no transitive drift in the root lock.
- Confirmed against the local crates.io index that `0.23.43` and `0.14.9` are the highest non-yanked stable releases (`0.24.0-dev.0/.1` are pre-release and correctly excluded), and that the item's own target table (`plans/issues/active/ad-hoc-latest-stable-release-convergence.md:101`) matches exactly. No scope creep to `tokio-rustls`, `rustls-pemfile`, or `rustls-platform-verifier`.

Vendor coherence (independently verified, not taken on trust)
- `.cargo-checksum.json` `package` hashes equal the `Cargo.lock` checksums for both crates.
- All 119 rustls files and 22 rcgen files hash-match their manifest entries; no missing and no extra files.
- Extracted `rustls-0.23.43.crate` and `rcgen-0.14.9.crate` from the registry cache (tarball SHA-256 matches the lock checksums) and `diff -r` against `vendor/` is byte-identical. `vendor/tokio-rustls`, `vendor/rustls-platform-verifier` etc. retain their crate-owned pristine locks (still naming rustls 0.23.31 / 0.23.37) and were correctly left untouched.

API usage
- `rustls::client::TicketRequest` is a genuine public re-export in 0.23.43 (`vendor/rustls/src/lib.rs:598`); `ClientConfig::send_ticket_request` defaults to `None` (`vendor/rustls/src/client/builder.rs:188`) and is consumed only in the TLS 1.3 ClientHello path (`vendor/rustls/src/client/hs.rs:317-321`).
- Server semantics verified in upstream source, not from the changelog: `max_tls13_tickets` defaults to `0` = ignore client requests (`vendor/rustls/src/server/builder.rs:124`, doc at `server_conn.rs:401-411`), and `send_tickets = min(requested, max_tls13_tickets)` only when `max > 0` (`vendor/rustls/src/server/tls13.rs:334-346`), with the EE hint bounded by the same value (`tls13.rs:701-705`). Setting `max = 2` alongside `send_tls13_tickets = 2` is therefore a genuinely *bounded* policy that cannot increase server ticket work above the pre-existing default.
- rcgen 0.14.9 does change `IsCa::ExplicitNoCa` to emit an empty `SEQUENCE` (cA omitted per DER DEFAULT FALSE) — `vendor/rcgen/src/certificate.rs:252-282`, with upstream's own `test_explicit_no_ca` asserting `ext.value == [0x30, 0x00]`. The Sifr test's byte patterns are correct: `06 03 55 1d 13 01 01 ff 04 02 30 00` is critical basicConstraints with an empty SEQUENCE; the rejected `…04 05 30 03 01 01 00` is the 0.14.8 explicit-FALSE encoding.

Constructor coverage
- `crates/sifr_runtime/src/tls.rs` is the only `ClientConfig`/`ServerConfig` construction site in the workspace (repo-wide grep for `ClientConfig::builder`, `ServerConfig::builder`, `with_platform_verifier`, `rustls::{Client,Server}Config` outside this file returns nothing). All five public constructors route through `apply_client_policy` (lines 208, 220, 237) / `apply_server_policy` (lines 252, 272). Policy coverage is complete; no bypass path.

Independent locks
- All four Rust-interop fixture locks containing rustls were updated; the other 17 fixture locks have no rustls edge and are untouched. Verified no orphaned package entries: `windows-sys 0.52.0` remains referenced (by `ring`) in every lock, and `getrandom 0.3.4` has no residual references where its package block was dropped.

Independent validation I ran
- `cargo test -p sifr_stdlib_manifest --test network_http_dependency_snapshots` — 9 passed.
- `cargo test -p sifr_runtime --all-features tls::` — 6 passed (base had 4; the two new tests are the delta, consistent with the claimed 295/295).
- `cargo clippy --workspace -- -D warnings` — clean. `cargo fmt --check` — clean. `python3 scripts/check_file_size_guardrails.py` — PASS (3243 files, limit 900).
- Not re-run (relied on the stated pre-commit evidence): the E2E pass suite and the full `--all-features` runtime/stdlib suites. The two public TLS fixtures assert only loose substrings (`assert "certificate" in str(...)`), so they carry no brittle rustls message text that this upgrade could have shifted.

Invariants
- No new `unwrap`/`expect` in production code; `usize::from(u8)` is infallible; all new `unwrap`s are inside `mod tests`. No `unsafe`, no printing. Public Sifr API surface unchanged — `sifr.tls` exposes no ticket or key-material handle. Renaming `apply_alpn_to_*` → `apply_*_policy` is forward-only with no compatibility shim, and no fallback trust store, legacy provider, or parallel dependency path was introduced.

## 3. Findings

All findings below are **non-blocking**. Ordered by severity.

**N1 — Ticket-policy test asserts only 2 of the 5 config constructors.**
`crates/sifr_runtime/src/tls.rs:650-675` exercises `client_config_with_roots` and `server_config`. `client_config_platform`, `client_config_with_roots_and_client_auth`, and `server_config_require_client_auth` have no ticket-policy assertion. The policy is centralized in the two `apply_*_policy` helpers, so behavior is currently correct, but the guard against a future constructor being added (or one being re-wired) without the policy is weaker than the "explicit bounded policy" acceptance language implies. A table-driven assertion over all five handles would close this. Related: `network_http_tls_traceability.md:16` states the test "asserts both policies", which is true per-policy but not per-constructor.

**N2 — Coverage is config-field-only; nothing verifies the policy on the wire.**
Neither new test observes a handshake. There is no loopback asserting that the RFC 9149 `ticket_request` extension is actually offered and echoed, nor that resumption succeeds under the bounded counts. The existing TLS loopbacks (`tls_loopback_split_close_notify_and_alpn`, and the `network_http_tls_loopback_split.sifr` e2e fixture) prove only that the new settings do not break the handshake. Given rustls unit-tests the extension itself, this is a reasonable line to draw, but the item's claim of "adopting" the feature currently rests on struct-field equality.

**N3 — The adopted feature is behaviorally neutral for Sifr↔Sifr, with a small outbound-fingerprint cost that the docs do not state.**
Client requests 2/2; rustls's server default `send_tls13_tickets` is already 2; `max_tls13_tickets = 2` caps at 2. So for any peer already defaulting to 2 tickets the observable delta is an added ClientHello extension plus an EncryptedExtensions hint, not a resumption improvement. rustls deliberately ships this opt-out-by-default (`send_ticket_request: None`) — every Sifr production TLS client now sends a rarely-seen extension on every outbound handshake, which is a ClientHello-fingerprint change and a (low, RFC-permitted-to-be-ignored) interop surface against strict middleboxes. The real benefit is against peers whose default is `1`, or on resumption where many servers send fewer. This trade-off is defensible and is what the item asked for; the gap is that `network_http_tls_traceability.md:16` and `network_http_dependency_audit.md:15` describe the policy without recording that it is count-neutral against rustls-default peers, which invites a future reader to over-read it as a resumption win.

**N4 — Independent lock re-resolution carried non-TLS unification drift that CI hosts do not exercise.**
Beyond rustls, `advanced_data_runtime/Cargo.lock` moved 5 `windows-sys` edges `0.52.0 → 0.59.0` and `tempfile`'s `getrandom 0.3.4 → 0.4.2`; `reqwest_loopback_runtime` and `resource_lifecycle_runtime` moved `socket2`'s `windows-sys 0.52.0 → 0.61.2`; `backend_feature_package` changed rustls only. I verified this is benign re-resolution rather than hand-editing: every new target version already existed as a package block in the base lock (pure unification, no added nodes), `tempfile 3.27.0` genuinely permits `getrandom >=0.3.0, <0.5` and `windows-sys >=0.52, <0.62`, and `windows-sys 0.52.0` stays alive via `ring`. Still, three of four locks changed edges unrelated to TLS, the four were evidently not regenerated by a uniform procedure, and the changed edges are all `cfg(windows)`/`getrandom`-backend paths that the macOS/Linux gate never compiles. The traceability's "The independent Rust-interop locks contain the new versions" understates this. Recommend an explicit one-line statement that the non-TLS edges are intentional re-resolution byproducts.

**N5 — rcgen DER assertion scans raw bytes instead of anchoring to the extension.**
`crates/sifr_runtime/src/tls.rs:678-703` searches the whole certificate DER for a 12-byte window and asserts absence of a 15-byte window. The positive assertion is sound; the negative assertion is technically probabilistic against random signature bytes rather than structural (probability is astronomically small, so this is a note, not a defect). Parsing basicConstraints would be exact, but `x509-parser` is deliberately not enabled for rcgen, so byte scanning is a defensible trade-off. Worth a comment in the test recording *why* it scans rather than parses.

**N6 — Documentation voice churn beyond the version change.**
`network_http_dependency_audit.md:15` and `:27` were rewritten from the table's terse-fragment style into full sentences, while the other 21 rows retain the original style. Column structure is preserved (verified: 9 pipes before and after, row counts unchanged), so nothing is broken — but it mixes two voices in one table and enlarges the diff beyond what the upgrade required.

**N7 — `crates/sifr_runtime/src/tls.rs` is now 860 lines.**
Under the 900-line guardrail (which passes), but with 40 lines of headroom. The next TLS-touching item is likely to force a split; flagging now so it is planned rather than reactive.

## 4. Blocking vs non-blocking

- **Blocking: none.** No correctness, security, panic-invariant, vendor-integrity, dependency-coherence, backward-compatibility, or file-size violation was found. Every claim I could check independently (vendor byte-identity, checksum chain, latest-stable selection, upstream API semantics, constructor coverage, lock reference integrity, clippy/fmt/guardrail, the two new tests) holds.
- **Non-blocking: N1–N7 above.** N1, N2, and N4 are the ones worth acting on — the first two as coverage strengthening, the third as one sentence of traceability precision.

## 5. Verdict

**SATISFIED**

yaseralnajjar:
## Create-PR gate evidence

Exact candidate: `f432ab8d88d44b21bd193bca66d5deaa38ac1102`

The one permitted `scripts/run_all_tests.sh --profile create-pr` execution completed. All reached correctness checks passed, including 10 Rust-interop variants, coverage, diagnostics, 19 Python-interop variants, and 28 runtime-platform variants with zero failures. The gate result is not passing because the required cold-cache runtime-platform lane took 204.735 seconds against its 120-second blocking time budget. The cold target cleanup was mandatory because this worktree's private target exceeded 20 GiB. The gate is not rerun, per phase policy.

Report: `target/validation_lane_reports/create-pr.latest.json`
Wall time: 1200.87 seconds. Max RSS: 1.4 GiB. Swaps: 0.

yaseralnajjar:
## Merge gate evidence

Exact candidate: `f432ab8d88d44b21bd193bca66d5deaa38ac1102`

The one permitted `scripts/run_all_tests.sh` execution passed every blocking lane. Highlights: 698/698 native E2E fixtures; 1,140 codegen tests; 1,043 lowering tests plus one expected ignore; 76/76 explicit driver generated-build integrations; 30 runtime-platform variants; 25 Python-interop variants; 183 diagnostics variants; sysroot boundary equivalence; generated-code TLS/HTTP demos; and all Rust-interop, distribution, performance, package, stdlib, fuzz, ecosystem, and regression areas.

The warm runtime-platform lane passed in 24.656 seconds, confirming that the create-PR gate's earlier 204.735-second result was a mandatory-clean cold-cache timing anomaly. The final merge report contains advisory-only wall-time and fixture-group skew notes.

Report: `target/validation_lane_reports/merge.latest.json`
Wall time: 6565.55 seconds. Max RSS: 2.2 GiB. Swaps: 0.

PR #3513, opened 2026-08-24
chore(deps): update ICU4X family to 2.3

yaseralnajjar:
## Summary
- update the five direct ICU4X dependencies to the latest stable 2.3/2.3.1 releases
- adopt the split locale-fallback graph and patched provider/zerovec releases
- add exact graph certification plus locale and Burmese collation regression coverage
- regenerate all affected package-owned vendor copies from official crates

## Validation
- focused sequential checks for all five direct dependencies
- sifr_runtime i18n and all-feature tests
- sifr_stdlib and sifr_stdlib_manifest tests
- 5/5 text/i18n E2E fixtures
- workspace Clippy, formatting, HIR guardrails, file-size guardrails
- official vendor byte comparison and 1,176 file checksums

## Known unrelated area result
The broad stdlib parity runner passed 283/284 demo checks. Its only failure is the already-recorded m16_raw_api canonical Python requirement issue; demos/text_i18n passed.


yaseralnajjar:
## Verdict: SATISFIED

No blocking findings. Item 12 converges all five ICU4X direct dependencies as one coherent unit, with the graph, vendor tree, tests, and records consistent at the exact candidate SHA.

### What I independently verified (beyond the supplied evidence)

**Version certification against the official source.** I fetched the crates.io sparse index for each direct dependency and each newly-introduced transitive. Every lock checksum matches the official registry byte-for-byte, and each selected version is the newest non-prerelease:

| crate | locked | index cksum match |
|---|---|---|
| `icu_collator` | 2.3.1 | ✓ `08984ed5…` (newest) |
| `icu_locale` | 2.3.1 | ✓ `785f595c…` (newest) |
| `icu_datetime` | 2.3.0 | ✓ `c44df99c…` (newest) |
| `icu_decimal` | 2.3.0 | ✓ `9eb8f655…` (newest) |
| `icu_plurals` | 2.3.0 | ✓ `e475e676…` (newest) |
| `icu_locale_fallback` | 2.3.0 | ✓ `251af8e5…` (new crate) |
| `icu_provider` | 2.3.1 | ✓ `d27bbb9d…` |

`Cargo.toml:96-100` matches the lock exactly; no crate appears at two versions.

**Vendor integrity.** I recomputed SHA-256 for every file in every ICU-family vendor directory (`icu_*`, `zerovec`, `zerovec-derive`, `tinystr`, `writeable`, `yoke`, `zerotrie`, `potential_utf`) against its `.cargo-checksum.json`: **0 mismatches, 0 missing, 0 extra files**. All 418 vendored `.cargo-checksum.json` `package` hashes match `Cargo.lock`, which in turn matches the crates.io index above. This chains the vendor bytes to the official releases independently of the owned 1,176-check comparison.

**Locale-fallback split handled correctly.** `icu_locale`'s `fallback` module was extracted to the new `icu_locale_fallback` crate (rename-detected at `vendor/icu_locale/src/fallback/mod.rs → vendor/icu_locale_fallback/src/lib.rs`), and `icu_calendar`/`icu_collator`/`icu_datetime`/`icu_decimal`/`icu_plurals` now depend on the leaf crate instead of full `icu_locale`. No first-party code touches the moved API — `crates/sifr_runtime/src/i18n.rs:18` imports only `Locale`, `LocaleCanonicalizer`, `LocaleExpander`, all still on `icu_locale`. Correctly *not* added as a direct workspace dependency, and `sifr_stdlib_manifest` needs no change because generated projects never name ICU crates directly (`features_tests.rs:196` still holds).

**Forward-only / scope discipline.** No compatibility lane, no version fallback, no re-exported legacy path, no unused unstable API. Every non-vendor change is ICU-owned: the five pins, one regression test file, three runtime test cases, one doc sentence, three record files, one whitespace attribute. No unrelated dependency moved; `syn 3.0.3` (now used by `zerovec-derive`) was already vendored and locked from a prior item.

**Panic safety.** `i18n.rs` retains typed-error mapping on every user-controlled path (`parse_locale`, `Decimal::from_str`, `try_from` range checks, all `try_new` constructors). No `.unwrap()`/`.expect()` was introduced. The `icu_collator` 2.3.1 Burmese non-starter-contraction panic fix is pinned by `burmese_non_starter_contractions_do_not_panic` (`i18n.rs:294-300`) — a genuine user-data panic regression guard, not a version tautology.

**Behavior pins.** `maximize_locale("und") == "und"` and `canonicalize_locale("en-1-foobar")` (`i18n.rs:244-249`) pin the two ICU4X 2.3 likely-subtags/extension behavior changes that would otherwise regress silently.

**Hygiene.** `git diff --check` over the exact range is clean. The `.gitattributes` addition is necessary and load-bearing: `vendor/icu_properties/src/unicodeset_parse/parse.rs` and `tests/data/PropertyDiscriminants.txt` are newly added upstream files carrying trailing whitespace, and `whitespace=cr-at-eol` does not disable the default `blank-at-eol` rule. No stale `2.2` ICU reference survives outside the baseline audit table (`plans/issues/active/…:102`, which is the correct from→to record) and archived reviews. Records at `verification/areas/stdlib_parity/{data,reports}/` and `internal_docs/text_i18n_architecture.md:38` all read 2.3 and name the correct per-crate versions. File-size guardrail: largest touched first-party file is `i18n.rs` at 443 lines; the new test is 84.

---

### Non-blocking suggestions

1. **Vacuous negative assertion** — `crates/sifr_stdlib_manifest/tests/icu_dependency_versions.rs:78`: `assert!(!locale_dependencies.contains(&"icu_pattern"))`. `icu_locale` never depended on `icu_pattern`; at base 2.2.0 its deps were `icu_collections, icu_locale_core, icu_locale_data, icu_provider, potential_utf, tinystr, zerovec`. The assertion cannot fail and therefore certifies nothing. The dependency the 2.3 split actually removed from `icu_locale` is `potential_utf` (it moved to `icu_locale_fallback`), so `!contains("potential_utf")` is the assertion that would genuinely pin the split.

2. **Over-tight transitive pins** — same file, lines 51-56 pin `zerovec 0.11.8` and `zerovec-derive 0.11.6` exactly. These are hand-off-limits transitives under the phase policy; any unrelated regeneration bumping them fails with the message "workspace lock must contain the patched zerovec release." The durable invariants are (a) exactly one locked version per package and (b) `icu_locale_fallback` present in `icu_locale`'s dep list — both already expressed by `locked_versions(...).len() == 1` without the version literal.

3. **Naming consistency** — siblings in that directory are `text_i18n_dependency_snapshots.rs`, `network_http_dependency_snapshots.rs`, `concurrency_runtime_dependency_snapshots.rs`. `icu_dependency_versions.rs` breaks the suffix convention.

4. **`.gitattributes` breadth** — the two pre-existing exceptions are per-file; the new rule is crate-wide (`vendor/icu_properties/**`). Harmless given byte-identical vendoring, but a repo-wide `vendor/** whitespace=-blank-at-eol,cr-at-eol` would be more honest about the actual policy than accumulating per-crate exceptions each ICU bump.

5. **Pre-existing, out of scope (noted, not a defect)** — several `verification/areas/rust_interop/fixtures/*/examples/*/Cargo.lock` pin `icu_normalizer`/`icu_properties` at 2.2.0 via `idna_adapter`. I confirmed these are package-owned locks resolving from the registry, not the sysroot vendor (`reqwest`, `rustls`, and their graph are absent from `vendor/`), so the vendor bump cannot affect them. They belong to the rust_interop area's own upgrade cadence.

I did not treat byte-identical upstream vendor formatting or the recorded `demos/m16_raw_api` Python canonical-requirement defect as Item 12 findings.

yaseralnajjar:
**Verdict: SATISFIED** — no blocking remediation defect in `a9ac93764c66666291fe95da2cde9f867ef2d687`.

Scope confirmed: `cf22e08a..a9ac93764` touches exactly two files (`.gitattributes`, and the renamed test), 23 insertions / 20 deletions. Worktree is clean at the exact candidate SHA; implementation and vendor bytes are untouched relative to the SATISFIED initial review.

Findings per accepted suggestion:

1. **Vacuous negative → real fallback split** (`crates/sifr_stdlib_manifest/tests/text_i18n_icu_dependency_versions.rs:59-64`). Verified against the lock: `icu_locale 2.3.1` deps = `[icu_collections, icu_locale_core, icu_locale_data, icu_locale_fallback, icu_provider, tinystr, zerovec]` — no `potential_utf`; `icu_locale_fallback 2.3.0` deps include `potential_utf`. The negative is non-vacuous: at base `483519cee`, `icu_locale 2.2.0` *did* list `potential_utf` and no `icu_locale_fallback` package existed, so both assertions would fail on the pre-upgrade lock. The new positive assertion covers the fallback edge. This is a genuine improvement over the old `!contains("icu_pattern")` check (`icu_pattern` was never an `icu_locale` dep at either version).

2. **Transitive pins removed, ICU certification retained.** `zerovec 0.11.8` / `zerovec-derive 0.11.6` exact pins are gone; `icu_locale_fallback = 2.3.0` and `icu_provider = 2.3.1` remain exact single-version assertions via `locked_versions` (vec equality still enforces exactly one locked occurrence). Consistent with the phase rule against hand-managing transitives while keeping the ICU fallback/provider surface certified. The five direct crates remain certified in test 1 against both manifest and lock.

3. **Rename.** `tests/icu_dependency_versions.rs` → `tests/text_i18n_icu_dependency_versions.rs` matches the sibling family (`text_i18n_dependency_snapshots.rs`, `network_http_dependency_snapshots.rs`, `concurrency_runtime_dependency_snapshots.rs`). Repo-wide grep finds no reference to either the old or new target name in scripts, workflows, or docs, so nothing is left stale by the rename (auto-discovered `tests/` target).

4. **Whitespace exception narrowed.** `vendor/icu_properties/**` → the two exact files. Verified they are precisely the only files under `vendor/icu_properties/` containing trailing whitespace (`grep -rlP '[ \t]+$'` returns exactly `src/unicodeset_parse/parse.rs` and `tests/data/PropertyDiscriminants.txt`). Both replacement lines re-specify `cr-at-eol`, so they do not drop the inherited `vendor/**` attribute; the trailing `vendor/*/Cargo.lock` line does not shadow them. `git diff --check 483519cee a9ac93764` exits 0 under the narrowed attributes.

No new mechanism defect. The new `package_dependencies` helper is lifetime- and type-correct (`&'a [Value]` → `Vec<&'a str>`), panics only on programmer-invariant absence in a test, and removes the duplicated inline extraction. Nothing to record as a later item.

yaseralnajjar:
Create-PR gate evidence

The single authorized create-PR gate ran on SHA a9ac93764c66666291fe95da2cde9f867ef2d687. It stopped after 159.01 seconds at the rust-interop matrix. Five maintained fixture locks still had the ICU 2.2 shared graph. Cache setup and every earlier blocking guardrail passed. The lane reported no advisory failure.

I did not rerun the create-PR gate. The final candidate SHA 4e4f489586994d2ef3da1286a15be652a4d12dd5 updates only those five fixture locks with constrained package updates. The targeted rust-interop matrix now passes both cases: 40 fixtures, 61 package examples, 21 scenario examples, and 237 self-tests. Locked metadata, diff hygiene, formatting, and the 3,244-file size guardrail also pass.

The one remaining merge gate will run once on the final candidate SHA.

yaseralnajjar:
Final candidate: `d95f5b665fb33945985b1931132f109091215d4b`

One-shot gate evidence (not rerun, per phase rule):

- Create-PR gate ran once on `a9ac93764c66666291fe95da2cde9f867ef2d687`. It reached the Rust interop matrix and exposed five maintained fixture locks with the old ICU shared package identity. Those locks were updated, and the full targeted Rust interop matrix then passed: 40 fixtures, 11 diagnostics, 44 crates, 61 package examples, 21 scenario examples, and 237 self-tests.
- Merge gate ran once on `4e4f489586994d2ef3da1286a15be652a4d12dd5`. It passed all preceding guardrails and the full Rust interop area, then stopped because the new `text_i18n_icu_dependency_versions` test target lacked coverage classification. Advisories: none. Report: `target/validation_lane_reports/merge.latest.json`.
- The final mechanical correction classifies that test as a merge-profile `test_fixture`. The complete targeted coverage-matrix readiness suite passes all 4 variants with 0 failures. `git diff --check` and the 3,244-file size guardrail also pass.

The one-shot gates were preserved exactly; neither gate was rerun.

PR #3515, opened 2026-08-24
chore(deps): consolidate SHA-2 on 0.11

yaseralnajjar:
## Summary
- replace the SHA-2 0.10 workspace dependency and version-named 0.11 alias with one canonical `sha2 = 0.11.0` declaration
- migrate all first-party digest callers to SHA-2 0.11 output newtypes and explicit lowercase byte encoding
- update the exact Rust-interop catalog pin, maintained fixture locks, and `hybrid-array` 0.4.14 vendor content
- add regression certification for direct declarations and all first-party lock edges

## Stable-source audit
- RustCrypto SHA-2 changelog: 0.11.0 is released; 0.11.1 is unreleased
- SHA-2 0.11 adopts Rust 2024, digest 0.11, output newtypes, serializable state, and backend selection
- the remaining SHA-2 0.10 lock node is required only by external SQLx 0.8 and Polars 0.54.4 crates, owned by Items 20 and 23; no maintained direct 0.10 declaration or alias remains

## Validation
- focused SHA-2 manifest, stdlib, sysroot, package, structural-identity, CLI, fixture-lock, and Rust-interop tests passed
- exact SHA-224, SHA-256, SHA-384, and SHA-512 known-answer vectors passed
- Rust-interop matrix: 40 fixtures, 11 diagnostics, 44 crates, 61 package examples, 21 scenario examples, 237 self-tests
- coverage readiness: 4/4
- workspace Clippy, formatting, HIR maintainability, diff hygiene, and file-size guardrails passed
- official `hybrid-array` archive digest and all 24 vendored file checksums verified
- one-shot create-PR gate completed all functional steps; its cold runtime-platform step passed functionally but exceeded the 120 s blocking budget at 227.896 s
- one-shot merge gate exited 0; all 35 lane steps passed, including runtime platform at 24.251 s and 698/698 E2E fixtures
- merge-gate wall time was 6459.31 s and emitted advisory-only wall-time and group-skew observations

No compatibility adapter, legacy path, fallback, or parallel first-party SHA-2 dependency was added.


yaseralnajjar:
Initial exact-SHA Opus review for `7e5cb3609dac1efd74f7b2d82e2a028fba298ead` against base `524835c29cad0a22b69922bcef2fdfd038118463`:

## Verdict: SATISFIED

No blocking findings.

### What I verified independently (read-only; no builds, no writes)

**Canonical dependency.** `Cargo.toml:135` now carries a single `sha2 = { version = "0.11.0" }`; the `sha2_0_11` alias is gone, and `grep` over the whole tree (excluding `plans/`, `vendor/`, `target/`, `third_party/`) finds no surviving `sha2_` alias reference and no first-party `0.10.9` pin. The exact catalog declaration is updated at `crates/sifr_rust_interop_catalog/Cargo.toml:43` (`=0.11.0`), and `crates/sifr_stdlib/Cargo.toml:37,53` moves the optional dep and the `hash` feature to `dep:sha2` while keeping the feature name stable, so no downstream generated-manifest surface shifts.

**Stable API adoption, no compat layer.** All `format!("{:x}", …)` digest formatting is gone from first-party code — a repo-wide `{:x}` grep hits only `vendor/`, `third_party/`. Replacements are direct 0.11 newtype consumers: `crates/sifr_driver/src/digest.rs`, `crates/sifr_package/src/digest.rs`, `crates/sifr/src/digest.rs`, `crates/sifr_sysroot/src/digest.rs:74`. No shim, no version-conditional path, no parallel legacy route.

**Digest byte/hex equivalence.** The new encoders emit `DIGITS[b >> 4]`, `DIGITS[b & 0x0f]` per byte — byte-for-byte identical to `{:02x}` zero-padded lowercase, so every SHA-256 hex contract (sysroot tree digest, stdlib source provenance, package/Python bridge + certification, SQLx metadata, CLI self-update installer validation) is preserved. Indices are provably ≤ 15, so no panic path; nothing `unsafe` added; all new files are 13–76 lines.

**API-level compile soundness against the vendored sources.** `&Sha256::digest(..)` → `&[u8]` holds via `impl Deref for Array { type Target = [T] }` (`vendor/hybrid-array/src/lib.rs:718-722`); `for byte in digest` holds via `impl IntoIterator for Array` (`src/iter.rs:65`); `hash.finalize().into()` in `crates/sifr_structural_identity/src/lib.rs:322` holds via `From<Array<T,U>> for [T; N]` (`src/lib.rs:757`).

**Lock/vendor coherence.** Workspace lock moves every first-party edge (`sifr`, `sifr_driver`, `sifr_package`, `sifr_rust_interop_catalog`, `sifr_structural_identity`, `sifr_sysroot`) to `sha2 0.11.0`; the single `0.10.9` node (`Cargo.lock:6900`) is consumed only by external crates. All 7 maintained fixture locks carrying `hybrid-array` are at `0.4.14`, and the core-language fixture lock shows the full coherent 0.11 chain (`block-buffer 0.12.1`, `crypto-common 0.2.2`, `digest 0.11.3`, `const-oid 0.10.2`, `generic-array`/`version_check` dropped). The one untouched fixture lock with `sha2 0.10.9` (`ecosystem_backend_certification/examples/backend_feature_package/Cargo.lock:978`) is reached only from `sqlx-core`/`sqlx-macros-core`/`sqlx-postgres` — correctly left alone.

**Vendor regeneration — verified against crates.io, not just self-consistency.** I downloaded the official crates and confirmed: `hybrid-array-0.4.14.crate` = `707114b52a…f6b` and `sha2-0.11.0.crate` = `446ba71750…5f4`, matching the lock checksums exactly. `diff -r` of the extracted official tarball against `vendor/hybrid-array` is **identical** (excluding `.cargo-checksum.json`), and all 24 listed file checksums recompute correctly with no unlisted or missing files.

**New certification test.** `crates/sifr_stdlib_manifest/tests/hash_sha2_dependency_versions.rs` asserts the workspace version, the absence of any `sha2_*` alias, the exact catalog pin, and that every first-party lock edge is `sha2 0.11.0` — and it is registered in `verification/areas/coverage_matrix/data/cargo_metadata_classification.json:255` with the same shape as the Item 12 ICU analogue.

### Non-blocking observations

1. **Duplicated hex encoder in four crates.** `crates/sifr_driver/src/digest.rs` and `crates/sifr_package/src/digest.rs` are byte-identical, `crates/sifr_sysroot/src/digest.rs:74` is the same body, and `crates/sifr/src/digest.rs:3` inlines the same loop. `sifr` and `sifr_driver` both already depend on `sifr_sysroot`, so three of the four could reuse one exported helper (`sifr_package` does not, so full unification isn't free).
2. **Redundant wrapper.** `crates/sifr_package/src/python/bridge_inventory/filesystem.rs:227` keeps a `pub(super) fn sha256_hex` that only delegates to `crate::digest::sha256_hex`, while sibling modules import the canonical one directly — two names for one operation.
3. **Test will break on the planned Item 20 work.** `hash_sha2_dependency_versions.rs:67` asserts the literal edge string `"sha2 0.11.0"`. Cargo emits the bare name `"sha2"` once only one version remains, so upgrading SQLx (which drops the `0.10.9` node) will fail this assertion rather than pass cleanly.
4. **`0.10.9` consumers are broader than stated.** Besides the three `sqlx-*` crates, `polars-plan` also depends on it (`Cargo.lock:5266`). Still one node, still external-only — but the item's "only where SQLx externally requires it" phrasing understates it.
5. **Weak vectors for the non-256 digests.** `crates/sifr_stdlib/tests/api_behavior.rs:284-286` asserts only output *length* for SHA-224/384/512; the E2E fixture `verification/areas/stdlib_parity/fixtures/audits/stdlib/10_hash_encoding.sifr:9` likewise checks only `len(h)`. SHA-256/MD5/SHA-1 do have exact vectors. Pre-existing, untouched by this diff, but a 0.11 core rewrite is exactly the change those length-only assertions would not catch.
6. **Active issue not updated in this diff** — consistent with the separate record-commit convention used for Item 12 (`cc50f46e1`), so I read this as deferred rather than missing.

One scope note on my own coverage: I did not run any build or gate, per the read-only constraint, so compile/runtime confirmation rests on the supplied gate evidence plus the API-level checks above against the actual vendored `hybrid-array` sources.

yaseralnajjar:
## Verdict: SATISFIED

This batch introduces **no new mechanism defect**.

### Delta verification (7e5cb36 → 24b6369)

**Digest-owner consolidation** — `sifr_sysroot::sha256_hex` is now the single shared encoder (`crates/sifr_sysroot/src/digest.rs:11`, re-exported at `crates/sifr_sysroot/src/lib.rs:12`), and `sha256_file` routes through it rather than duplicating the `lower_hex(&Sha256::digest(..))` pair. Both duplicate modules are gone (`crates/sifr/src/digest.rs`, `crates/sifr_driver/src/digest.rs`) with their `mod` declarations removed. All former call sites now import the sysroot symbol; both crates already declare `sifr_sysroot` (`crates/sifr/Cargo.toml:24`, `crates/sifr_driver/Cargo.toml:21`), so no new dependency edge. The deleted `sifr_driver::digest::lower_hex` had no remaining users — its only in-crate hex use is the local helper in `rust_interop_sqlx_offline_tests.rs`.

**Package-owned encoder** — `crates/sifr_package/Cargo.toml` has no `sifr_sysroot` edge, so keeping `crates/sifr_package/src/digest.rs` private is correct rather than a residual duplicate. Its `lower_hex` is still used directly by `binding_authoring.rs:103` and `bridge_resolution.rs:56`, so nothing is now dead. The removed `filesystem::sha256_hex` wrapper has no remaining referents; `bridge_inventory/mod.rs:3` imports `crate::digest::sha256_hex` directly and the trimmed `use filesystem::{...}` list matches actual usage. Exactly two first-party encoders remain, each justified.

**Lock-edge assertion** — `hash_sha2_dependency_versions.rs:69-77` derives the expected rendering from the count of locked `sha2` packages. The lock currently holds 0.10.9 and 0.11.0 (`Cargo.lock:6900`, `:6911`), so the versioned branch is live today; when Item 20 removes external 0.10.9 (held only by `sqlx-core`/`sqlx-macros-core`/`sqlx-postgres` 0.8.6 and `polars-plan` 0.54.4), Cargo renders bare `sha2` and the test still passes. All seven first-party lock edges are `sha2 0.11.0`.

**Known-answer vectors** — SHA-224/384/512 of `abc` in `crates/sifr_stdlib/tests/api_behavior.rs:284-309` match the FIPS 180-4 published digests byte for byte; the functions under test return `Vec<u8>` (`crates/sifr_stdlib/src/hash.rs:17-30`), so the comparisons are well-typed.

**Base..candidate still satisfies Item 13** — workspace `sha2 = { version = "0.11.0" }` (`Cargo.toml:135`), catalog pin `=0.11.0` (`crates/sifr_rust_interop_catalog/Cargo.toml:43`), no version-named alias, every first-party edge on 0.11.0, and residual 0.10.9 confined to external crates, which is the documented Item 20 boundary.

### Non-blocking observations

1. `crates/sifr/Cargo.toml:29` still declares `sha2 = { workspace = true }`, but after `crates/sifr/src/digest.rs` was deleted there is no `sha2::`/`Sha256` use anywhere in that crate (the only `sha2` token left is a feature-name string literal in `crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs:46`). It is now an unused direct dependency that also keeps a phantom `sifr -> sha2 0.11.0` lock edge alive. No lint denies it, so it does not fail any gate.
2. In the single-version branch, `expected_edge = "sha2"` accepts a bare edge without checking the locked version is 0.11; the 0.11 guarantee there rests entirely on the sibling manifest test. Asserting the sole locked version starts with `0.11` would make that branch self-sufficient.

Verification here was static (diff, symbol/reference, and lock-graph analysis); I ran no builds or tests, so the compile/test claims rest on the focused post-remediation evidence you reported.

yaseralnajjar:
Post-review disposition for final candidate `b992ace751235b69ade3acd6ff6038de32e0d2e6`:

- removed the unused direct `sifr -> sha2` edge
- made the lock-edge regression fail unless SHA-2 0.11.0 is present
- reran the focused manifest and CLI tests, workspace Clippy, formatting, HIR maintainability, diff hygiene, and the 900-line guardrail

The remediation review of `24b6369835ec2cab7594370a36e7883acadf606c` was SATISFIED and reported no new mechanism defect. These were its two non-blocking observations. Per the phase rule, no third Opus review will run.

yaseralnajjar:
Final exact-SHA gate evidence for `b992ace751235b69ade3acd6ff6038de32e0d2e6`:

- The sole create-PR gate completed all functional steps it reached. The cold runtime-platform step passed functionally but exceeded its 120,000 ms blocking budget at 227,896 ms, so the one-shot run exited nonzero as recorded in `target/validation_lane_reports/create-pr.latest.json`.
- The sole merge gate exited 0. All 35 lane steps passed. Runtime platform completed in 24,251 ms, including `binary_file_io_capability.sifr` in 3,651 ms.
- The merge gate passed 76/76 ignored generated-build tests, 1,140/1,140 `sifr_codegen` tests, and 698/698 E2E fixtures.
- Merge wall time was 6,459.31 s. Its wall-time and group-skew findings are advisory. The repository rule excludes a first cold-cache run as host-sensitive performance evidence.
- Opus initial and remediation reviews both returned SATISFIED. No further review or gate run is allowed or needed for this item.

This exact candidate is ready to merge.

PR #3517, opened 2026-08-25
chore(deps): upgrade Base64 to 0.23

yaseralnajjar:
## Summary
- upgrade the maintained Base64 dependency and vendor source from 0.22.1 to latest stable 0.23.1
- retain the official 0.22.1 vendor package required by external locked dependencies
- disable default features and enable only `std`, excluding the new default-on `simd-unsafe` feature
- use the canonical prelude engine constants and extend strict trailing-bit and padding error coverage
- certify direct dependency policy, first-party lock edges, and complete Base64 vendor closure

## Stable-source audit
- crates.io index: 0.23.1, not yanked, published 2026-08-04
- official 0.23.1 archive SHA-256: `ac07cdecf99051d9a5238b80f35af32cdeba5b336e55d957b318b50137e18da5`
- official retained 0.22.1 archive SHA-256: `72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6`
- upstream 0.23 adds more preconfigured constants, clearer invalid-last-symbol errors, custom padding symbols, and optional SIMD engines
- Sifr selects the safe scalar implementation because the SIMD feature is the only upstream unsafe-code source

## Validation
- Base64 dependency and vendor-closure certification: 3/3
- full sifr_stdlib_manifest suite: pass
- Base64 stdlib API behavior: pass
- compiler bootstrap and stateless crypto codegen checks: pass
- focused generated E2E: 14/14 fixtures
- coverage readiness: 4/4
- required workspace Clippy: pass with warnings denied
- changed certification-test Clippy: pass with warnings denied
- formatting, HIR maintainability, diff hygiene, and 3,247-file size guard: pass
- both official Base64 vendor packages pass every recorded file checksum
- an offline locked generated-style dependency probe resolves both Base64 0.22.1 and 0.23.1 from the checked-in vendor tree
- resolved Base64 0.23.1 feature graph contains only std and alloc; simd-unsafe is absent

A broader all-test-target Clippy probe found unchanged test-harness `expect` lint debt outside this dependency item. The required workspace Clippy owner check and the changed certification-test Clippy check pass.

No compatibility adapter, legacy path, fallback, or unsafe Base64 feature was added.


yaseralnajjar:
**Verdict: NOT SATISFIED**

**Blocking findings:**

- regression: `vendor/base64/Cargo.toml:16` (and the absence of any `vendor/base64-0.22.x/` directory), criterion "vendor content and generated dependency behavior agree with the root manifest and root lock". The checked-in `vendor/` tree is the sysroot vendor registry that generated builds consume via source replacement (`crates/sifr_driver/src/build/cargo_manifest.rs:60-73`, `scripts/distribution/build_release_artifacts.sh:210-266`). The candidate rewrote the single `vendor/base64` package in place from 0.22.1 to 0.23.1, but four packages already vendored in that same closure still require `base64 ^0.22` and cannot accept 0.23: `vendor/pem/Cargo.toml:57-60` (non-optional, `version = "0.22.0"`), `vendor/hyper-util/Cargo.toml:154-156`, `vendor/rustls-platform-verifier/Cargo.toml`, `vendor/serde_with/Cargo.toml`. `hyper-util` and `rustls-platform-verifier` are direct optional dependencies of `sifr_stdlib` (`crates/sifr_stdlib/Cargo.toml:20,33`), so they are in the resolve graph of every generated program crate regardless of which stdlib features are enabled, and `Cargo.lock:652-657` still pins `base64 0.22.1` for them. With `CargoVendorMode::SysrootOnly` the directory source is the only registry, so resolution now fails with no matching `base64 0.22.1` package. This is the exact case the tree already handles elsewhere by keeping both versions (`vendor/sha2` 0.11.0 + `vendor/sha2-0.10.9`, `vendor/digest` 0.11.3 + `vendor/digest-0.10.7`, `vendor/hashbrown` 0.17.0 + three suffixed dirs). Correction: add the official `base64` 0.22.1 package back as `vendor/base64-0.22.1/` with its `.cargo-checksum.json`, exactly as `cargo vendor` emits it, so the vendor closure matches both lock edges; confirm by resolving a generated crate against the vendor dir with `--offline --locked`.

- in-scope omission: `crates/sifr_stdlib_manifest/tests/base64_dependency_version.rs:69-112`, criterion "exact-version/feature certification agrees with vendor content". The new certification asserts the workspace spec, the 0.23.1 lock entry, first-party lock edges, and `vendor/base64/Cargo.toml` version, but never asserts that every `base64` version present in `Cargo.lock` for vendored packages exists in the vendor tree — which is why the gap above passed all listed evidence (per `verification/areas/rust_interop/README.md:66`, no gate claims sysroot-vendor coverage). Correction: extend the certification to require a vendored directory for each locked `base64` version reachable from `sifr_stdlib`'s dependency closure.

**Follow-up findings:**

- infrastructure issue: there is no gate anywhere that resolves the sysroot vendor closure offline, so any vendor/lock divergence introduced by a dependency-train item is invisible to the PR and merge gates. A single `cargo metadata --offline --locked` against a synthesized generated-crate manifest with the vendor source replacement would close this class permanently.
- pre-existing issue: `vendor/` mixes plain and version-suffixed directory names by "newest wins," so an in-place upgrade of the plain directory silently drops the previous version. Nothing records which vendored consumers depended on the displaced version.
- suggestion: base64 0.23 changed `DecodeError::InvalidLastSymbol` to include the decoded value, so `error.to_string()` output in `crates/sifr_stdlib/src/base64.rs:14,29,108,117` changes text for that case. Parity tests only assert `is_err()`; if any user-facing docs or fixtures quote the old message, they are worth a grep.

The non-vendor portion of the change is otherwise sound: `default-features = false, features = ["std"]` correctly excludes `simd-unsafe` (`vendor/base64/Cargo.toml:41-47`, `std = ["alloc"]`), and `BASE64_STANDARD`/`BASE64_URL_SAFE` are exact re-exports of `general_purpose::STANDARD`/`URL_SAFE` (`vendor/base64/src/prelude.rs:19-22`), so encode/decode parity and the Base32/Base32hex paths are unaffected, with no compatibility or fallback path added.

yaseralnajjar:
Exact-SHA remediation review for `96e6a72f18be675080fa2ed020d898712ef73a1c`:

## Exact-SHA remediation review — Item 14 (Base64)

Reviewed `da4228e41` → `96e6a72f1` (two commits: `e31728a4a` upgrade, `96e6a72f1` remediation). Read-only; no repository state was modified.

### Change surface (non-vendor)
- `Cargo.toml`: `base64 = { version = "0.23.1", default-features = false, features = ["std"] }`
- `Cargo.lock`: adds `base64 0.23.1`, retains `0.22.1` (and pre-existing `0.13.1`), `sifr_stdlib` edge moved to `0.23.1`
- `crates/sifr_stdlib/src/base64.rs`: `general_purpose::STANDARD/URL_SAFE` → `prelude::{BASE64_STANDARD, BASE64_URL_SAFE}`
- `crates/sifr_stdlib/tests/api_behavior.rs`: added non-canonical-padding / trailing-bit rejection vectors
- `crates/sifr_stdlib_manifest/tests/base64_dependency_version.rs`: new, 3 tests, 220 lines
- coverage-matrix classification: new `base64_dependency_version` test target, inserted in declaration (alphabetical) order

### Blocker 1 — 0.22.1 removed from vendor: corrected
`vendor/base64-0.22.1` is byte-identical to the base commit's `vendor/base64` tree except that `.cargo-checksum.json` gained cargo's newer `$comment` key; the `files` map and `package` hash are identical. I recomputed SHA-256 for every file in both `vendor/base64` (0.23.1) and `vendor/base64-0.22.1`: all match, no missing entries, no unlisted extra files. Both `package` hashes match their `Cargo.lock` checksums (`72b3254f…` / `ac07cdec…`).

The requirement is real and now satisfied: recomputing the sysroot vendor closure from `Cargo.lock` (first-party ∪ vendored packages) yields exactly three Base64 edges — `pem 3.0.6 → 0.22.1`, `hyper-util 0.1.20 → 0.22.1`, `sifr_stdlib → 0.23.1`. Both `pem` and `hyper-util` are vendored, so 0.22.1 is genuinely load-bearing for `SysrootOnly` builds.

### Blocker 2 — certification did not assert vendor coverage: corrected
`vendored_base64_packages_cover_vendored_and_first_party_lock_edges` enumerates `vendor/*/Cargo.toml`, builds the `(name, version)` identity set, selects lock packages that are first-party or vendored, derives their Base64 edge versions, asserts the set is exactly `{0.22.1, 0.23.1}`, and asserts each required version exists in vendor. That is the correct closure boundary: package-mode builds (`PackageOwned`) take the package's own lock as authoritative and may use non-vendored registry packages, so the vendor tree only needs to cover the sysroot closure — which is what the test derives. This is consistent with the observed facts that `spm_precompiled → base64 0.13.1` (lock, not vendored) and the `advanced_data_runtime` fixture lock's `0.13.1` entry are both outside the vendor tree today.

Directory layout is correct on both consumers: Cargo's directory source identifies packages by manifest, not directory name, and the first-party validator `vendor_contains` (`crates/sifr_driver/src/build/cargo_resolution.rs:429`) probes `name-version` then `name` with manifest-identity plus checksum verification. The `base64` + `base64-0.22.1` naming also matches the repo's existing convention (`bitflags` + `bitflags-1.3.2`, `block-buffer` + `block-buffer-0.10.4`).

### 0.23 migration soundness
- `prelude::BASE64_STANDARD`/`BASE64_URL_SAFE` re-export the same `general_purpose::STANDARD`/`URL_SAFE` consts, so the rename is behavior-neutral; every call site in `base64.rs` routes through them, including `base64_decode_opts`, and no `general_purpose::` usage remains anywhere first-party.
- `default-features = false, features = ["std"]` correctly drops 0.23's default-on `simd-unsafe` (vendored `Cargo.toml` `default = ["std", "simd-unsafe"]`), keeping the scalar engine and avoiding an `unsafe` SIMD dependency — the right call given the project's unsafe/panic-free posture, and the test asserts it with the reason inline.
- MSRV 1.71.0 vs workspace `rust-version = "1.98"`: fine.
- 0.23's `InvalidLastSymbol` message change is not observable: no first-party test, snapshot, or E2E fixture asserts Base64 error text (fixtures only assert `len(e.message) > 0`).
- Only one Base64 declaration exists workspace-wide (`base64 = { workspace = true, optional = true }` in `sifr_stdlib`); no stale `0.22` pin remains in first-party code or codegen.
- `vendor/base64-0.22.1` is outside the file-size guardrail's roots (`crates`, `scripts`, `verification`, `demos`, `.github/workflows`), so the added vendor files cannot trip it.

### Non-blocking observations
1. The coverage assertion is Base64-specific. Nothing asserts the general invariant "every dependency of every vendored package is itself vendored." Sibling per-dependency tests share that shape, so this is consistent — but a single generic vendor-closure guardrail would retire this whole class of defect and is worth a later phase item.
2. `required_base64_versions == {"0.22.1", "0.23.1"}` is a hard-coded expectation. It fails closed (loudly, when a vendored dep drops or adds a Base64 edge), which is acceptable for a certification, but it is a maintenance tripwire rather than a derived invariant.
3. The coverage test asserts presence and version identity of vendored Base64 packages but not their `.cargo-checksum.json` `package` hash against the lock checksum. Cargo and `vendor_contains` both enforce this at build time, so it is hardening, not a gap.
4. Tests 1 and 3 mix a fixed `include_str!("../../../vendor/base64/Cargo.toml")` path with dynamic vendor discovery; if the tree were ever re-vendored to `base64-0.23.1`, test 1 would fail to compile. Loud and trivially fixable, not a correctness risk.
5. Stale-vendor edge case: a vendored directory whose version diverges from the lock is silently skipped by `is_vendored_package`, so its Base64 edge would not be counted. Such a tree cannot build (lock/checksum mismatch), so this is not reachable as a silent failure.

No new mechanism defect found. Both original blockers are corrected at the mechanism level, not papered over.

VERDICT: SATISFIED

yaseralnajjar:
Exact-SHA gate record for `96e6a72f18be675080fa2ed020d898712ef73a1c`:

- create-PR gate: one execution only. All executed validation was functionally green; exit 124 came solely from the cold `area_runtime_platform` timing budget (`206.884s > 120s`) after the required 60.6 GiB target cleanup. No rerun.
- merge gate: one execution only, exit 0. All functional steps passed, including 698/698 E2E fixtures (`127353b213e16688`), both Base64 versions in the full feature graph, offline lock-mode builds, and installed/source sysroot boundary equivalence. Overall warm-time target was advisory because the cleaned target caused broad cache misses. No rerun.

Opus remediation review: SATISFIED. No new mechanism defect.

PR #3519, opened 2026-08-25
Upgrade num-bigint to 0.5.1

yaseralnajjar:
Item 15 of the latest-stable convergence phase.

- upgrades every Sifr-owned num-bigint edge to 0.5.1 with explicit std-only features
- isolates BigDecimal's incompatible upstream 0.4 line at 0.4.8
- replaces both vendor trees from official crates.io packages and certifies their checksums
- adopts BigInt::ZERO and BigInt::ONE in first-party integer code
- updates maintained locks, generated dependency plans, and feature-tree snapshots

Focused validation is green: affected frontend/lowering/runtime/stdlib suites, dependency-plan authority E2E, generated-demo freshness, decimal Sifr run, standalone idiomatic Rust demo compile, workspace Clippy, formatting, HIR/file-size/coverage guardrails, and exact lock/vendor certification.

The required exact-SHA Opus review and the single create-PR/merge gates will be posted before merge.


yaseralnajjar:
## Review — Item 15, Num BigInt 0.5 (`c0e2a66` → `037b24c`)

Verification I ran (read-only, on the exact diff):
- Vendor authenticity: computed SHA-256 of every file in both vendor trees against their `.cargo-checksum.json` (54/54 and 52/52 match, no extra/missing files), and cross-checked the two package checksums against the live crates.io sparse index — `0.5.1 = 93e7820b…` and `0.4.8 = c89e69e7…` are exact, and both are the latest non-yanked releases in their lines (0.4.7 and 0.5.0 are yanked).
- All five first-party direct dependents (`sifr_frontend`, `sifr_ir`, `sifr_lowering`, `sifr_runtime`, `sifr_stdlib`) resolve to `num-bigint 0.5.1` in the root lock; `0.4.8` appears only under externally owned edges (`bigdecimal`, `ciborium`, `der-parser`, `parquet`).
- Every tracked non-vendor `Cargo.lock` with a num-bigint node is at 0.5.1, with 0.4.8 co-resident only in the two locks that pull BigDecimal/Parquet; no `0.4.6` reference survives anywhere outside `vendor/` and archived plans.
- `BigInt::ZERO`/`BigInt::ONE` substitutions are value-identical at every site (const-eval div-by-zero, floor-div/mod, truthiness, shift/pow budget guards, `non_negative_u32`, runtime floor-div adjust); no limit constant, radix parse path, or serde surface is touched (num-bigint's `serde` feature is not enabled before or after, and no first-party type serializes `BigInt`).
- Generated code is unchanged and version-agnostic: `demos/decimal_conversions/emitted.rs` reaches BigDecimal's integer via `into_bigint_and_scale()` without naming the 0.4 type, which is why only the hand-written `idiomatic.rs` needed the `bigdecimal::num_bigint::BigInt` alias — and it aliases rather than converts, with `.to_i64()` working through the single shared `num-traits 0.2.19`.
- Feature parity: old `default` == new explicit `std` (the four feature-tree snapshots confirm `i128`/`num-integer/std`/`num-traits/std` are all retained).
- New certification test is registered in the right coverage-matrix package section and follows the existing `*_dependency_versions.rs` precedent; all touched hand-maintained files are ≤ 886 lines.

Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- infrastructure issue: `crates/sifr_stdlib_manifest/tests/num_bigint_dependency_versions.rs:129` (`collect_lock_paths`) walks the whole workspace excluding only `.git`, `target`, and `vendor`, so it also reads gitignored generated projects — `sifr_output/Cargo.lock` and anything under `tmp/` are picked up today. A stale local or mid-suite generated lock makes the assertion fail for reasons unrelated to the maintained inputs. Restricting traversal to tracked/maintained lock paths (or skipping gitignored roots) would make the gate hermetic.
- suggestion: same test, `expected_edge` at line 86 — when a lock contains exactly one num-bigint node the expected dependency string degrades to bare `"num-bigint"`, and the only version constraint is membership in `{0.4.8, 0.5.1}`. A first-party fixture lock that regressed to a 0.4.8-only resolution would pass. Asserting the first-party edge resolves to 0.5.1 specifically (rather than to "the single node, whatever it is") closes that hole.
- pre-existing issue: `demos/decimal_conversions/idiomatic.rs` is not compiled by any gate — the dual-owner num-bigint typing added here was validated only by an ad-hoc temporary Cargo project. The idiomatic demo corpus has no CI compile step, so this file can silently rot on future dependency moves.

yaseralnajjar:
Exact-SHA review and gate evidence

Candidate: `037b24c64c5ac5ac516efbac6d86b3e0c3154a0b`
Base: `c0e2a66f4537bd70f36addbb17b0567715fc8ba4`

- Claude Opus exact-SHA review: SATISFIED. Review: https://github.com/sifr-lang/sifr/pull/3519#issuecomment-5405668574
- Create-PR gate: one run only. All functional areas reached passed. The runtime-platform lane exceeded its 120 s blocking budget on the first cold-cache build (`204572 ms`, with `binary_file_io_capability.sifr` at `159464 ms`), so the gate reported a timing failure and stopped after that lane. Per the phase rule, this gate was not rerun. Report: `target/validation_lane_reports/create-pr.latest.json`.
- Merge gate: one run only, on the same exact SHA. PASS. All functional steps passed, including 698/698 E2E cases (signature `127353b213e16688`), full workspace Cargo tests, generated-code checks, distribution/release, sysroot release, package management, stdlib parity, regressions, fuzz/property, and ecosystem validation. Runtime-platform was warm and passed in `3643 ms`. Advisory-only wall-time and group-skew notices remained. Report: `target/validation_lane_reports/merge.latest.json`.
- Merge lane: `6539.20 s` wall time, `16587.72 s` CPU, `2.2 GiB` max RSS, zero swaps.

PR #3523, opened 2026-08-25
build: upgrade lsp-server to 0.10

yaseralnajjar:
## Summary

- upgrade the single LSP Server dependency from resolved 0.7.9 to latest stable 0.10.0
- adopt 0.10 typed response outcomes in the canonical server response path
- replace the official vendored package and certify its exact manifest, lock edge, archive hash, and package checksum
- add protocol serialization tests that prevent invalid result/error response states

## Official audit

- crates.io latest stable: https://crates.io/crates/lsp-server/0.10.0
- upstream source: https://github.com/rust-lang/rust-analyzer/tree/1b7da4272d8d27c78774f42a6e1ea66a4c1fe984/lib/lsp-server

## Validation

- `cargo test -p sifr_lsp` (77 passed)
- `cargo test -p sifr_stdlib_manifest --test lsp_server_dependency_version` (3 passed)
- developer-tooling LSP smoke and semantic editor suites (8 variants passed)
- coverage matrix (5 variants passed)
- changed production and certification target Clippy with warnings denied
- `cargo fmt --all --check`
- all 14 vendored package file checksums and the official archive hash
- `git diff --check`

Compiler inputs changed. The one create-PR gate and one merge gate will run only after exact-SHA Opus approval.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- None.

Verification performed (exact SHAs 67f613e1b → 42f0dcf15):
- **Response semantics.** `crates/sifr_lsp/src/server.rs:222-232` is the single construction site; every production path routes through it — shutdown (`server.rs:74`, `Ok(Value::Null)`, byte-identical to the old `new_ok(id, ())` which serialized `result: null`), enqueue rejection (`:125`), missing-scheduled-body internal error (`:164`), normal dispatch success/error (`:183`), and pre-dispatch cancellation (`:200`). `requests::handle` already returns `LspResult<Value>` (`crates/sifr_lsp/src/requests/mod.rs:20`), so no serialization is lost; the change also drops upstream `new_ok`'s data-dependent `to_value(..).unwrap()`.
- **No parallel path.** `grep` for `new_ok|new_err|response_result` across `crates/` finds no first-party use of the split constructors, and `Response` appears nowhere in `sifr_lsp/src` outside `server.rs`. Error codes still come from `lsp_server::ErrorCode` via `LspError` (`crates/sifr_lsp/src/errors.rs:9-49`), so `-32602/-32603/-32800/-32601` are unchanged. No adapter, shim, or legacy branch added.
- **Test coverage of shape.** The new unit tests assert typed access plus exact JSON (`{"id":1,"result":null}` / `{"id":2,"error":{...}}`); a both-fields or neither-field response is now unrepresentable by type. On the wire, `verification/areas/developer_tooling/lsp_protocol.py:52-68` rejects any response carrying `error` for success calls and requires an `error` object for `request_error`, and the smoke suite exercises shutdown (`lsp_protocol_smoke.py:244+`) and a real error code (`:238-243`). A regressed result/error shape would fail these.
- **Evidence agreement.** Root manifest `Cargo.toml:106` = `0.10.0`; `crates/sifr_lsp/Cargo.toml:16` inherits via workspace; root lock has exactly one `lsp-server` entry at `0.10.0` with checksum `3ee25a31…7656` and unchanged transitive deps; `vendor/lsp-server/.cargo-checksum.json` package field matches; I recomputed all 14 vendored file SHA-256s independently — all match, with no extra or missing files; `.cargo_vcs_info.json` sha1 `1b7da42…` matches the cited upstream release commit. crates.io confirms `max_stable_version = 0.10.0`, not yanked, with cksum `3ee25a31…7656`. Only `vendor/lsp-server` exists (no stale versioned duplicate). Coverage classification adds `lsp_server_dependency_version` as `test_fixture`/`merge`, alphabetically placed and matching the sibling dependency-certification entries.
- **Evidence vs. criteria.** Focused evidence covers response handling, protocol/transcript/editor smoke, certification, coverage, lint, format, and guardrails; compiler-input gates are correctly deferred to exact-SHA approval.

Follow-up findings:
1. `vendor/lsp-server/src/msg.rs:82-95` — 0.10 makes incoming `Response` deserialization strict: a client response with neither `result` nor `error` now fails `Message::read` (an I/O `InvalidData` on the reader thread) where 0.7's `Option` fields silently accepted it. Sifr sends no server→client requests (`server.rs:114` discards `Message::Response`), so this is unreachable in practice today; worth remembering if outbound requests (e.g. `client/registerCapability`, `workspace/applyEdit`) are added later.
2. `crates/sifr_lsp/src/server.rs:241-278` — the new unit tests pin `invalid_params` and `internal` codes but not `request_cancelled` (`-32800`), which is the code used by three of the five call sites. Wire coverage exists in `verification/areas/developer_tooling/lsp_protocol_stress.py:133`; a cheap unit assertion would localize a regression.
3. `vendor/lsp-server/src/req_queue.rs:21-25` — 0.10 adds `ReqQueue::has_pending`; Sifr uses its own `request_queue`, so nothing to adopt. Noted only so a future audit doesn't read this as missed functionality.
4. The pre-existing `unnecessary_to_owned` warning in an unchanged `sifr_lsp` test target is out of scope per the item instructions and is not an Item 17 regression.

yaseralnajjar:
Exact implementation candidate: `42f0dcf15048171376ccdd4124001b1b215dfa09`

- Opus exact-SHA review: SATISFIED — https://github.com/sifr-lang/sifr/pull/3523#issuecomment-5409418083
- Create-PR gate: run exactly once. All functional checks passed. The cold runtime step took 210.102s against the 120s warm budget, so the runner exited 124. It was not rerun.
- Merge gate: run exactly once on the reviewed SHA. Exit 0; wall time 6662.74s.
- Exhaustive compiler evidence: 1142 codegen tests passed.
- E2E evidence: 698 passed, 0 failed; report signature `127353b213e16688`.
- The merge runner reported only advisory cold-cache timing and batching guidance.

No remediation review or second gate was required.

PR #3527, opened 2026-08-25
build: upgrade Rusqlite to 0.40.2

yaseralnajjar:
Latest-stable Item 19.

- Upgrades the exact Rusqlite line from 0.39.0 to 0.40.2 and libsqlite3-sys to 0.38.2.
- Disables unused defaults and keeps only the bundled native SQLite provider.
- Exercises safe quoting for a SQL-shaped savepoint name from the 0.40 security fix.
- Removes the unused hashlink, sqlite-wasm-rs, and VFS dependency chain.
- Adds exact manifest, feature, lock, checksum, runtime, and native-trust certification.

Official evidence:
- crates.io checksum: 23f2a97da3e3873c73cb2a2e71b35c40ff95e0b1eefa8d72d8499a6928c3b5b3
- upstream tag commit: e88f112bef7899234a497baed5cc3c3d553deeb8

Focused validation:
- Rusqlite certification: 3/3
- Rust interop area: 10/10; matrix self-tests: 239
- lifecycle runtime and alias rejection runtime: pass
- package management: 4/4
- coverage matrix: 5/5
- sifr_stdlib_manifest tests: pass
- locked offline metadata and fixture build: pass
- catalog and fixture Clippy with warnings denied: pass
- formatting, documentation links, HIR, diff, and file-size guardrails: pass

Exact candidate: dd1e9b55c88ae60d21fea4335f155fe08cde53b4


yaseralnajjar:
## Verdict: SATISFIED

Reviewed at exact SHAs (base `4c1aaf088`, candidate `dd1e9b55c`), read-only. No repository mutation; the only writes were two `/tmp` scratch copies used for a formatting comparison.

### Mechanism verification (independent, against the registry sources)

- **Provider policy is unambiguous and single-path.** `rusqlite-0.40.2/Cargo.toml:362` declares `libsqlite3-sys 0.38.2` as a **non-optional** dependency under `cfg(not(all(target_family="wasm", target_os="unknown")))`, while `libsqlite3-sys`/`sqlite-wasm-rs` are optional only on the wasm target (lines 325–332). So `default-features = false, features = ["bundled"]` yields exactly one FFI provider (`bundled` → `libsqlite3-sys?/bundled` + `modern_sqlite`) with `cache` (→ `hashlink`) and `ffi-sqlite-wasm-rs` (→ `dep:sqlite-wasm-rs`) both off. That matches both locks exactly. No compatibility, legacy, or parallel path exists.
- **Dropping `cache` is safe.** No first-party or fixture code uses `prepare_cached`/statement-cache APIs (`git grep prepare_cached` hits only `sifr_driver`'s own artifact-cache helpers). The catalog declares `rusqlite` optionally with no Rust code behind it.
- **The savepoint proof is real, not decorative.** `rusqlite-0.39.0/src/transaction.rs:253,314` interpolate the raw name: `execute_batch(&format!("SAVEPOINT {name}"))` / `RELEASE {name}` — the fixture's name `sifr; DROP TABLE evidence; --` would execute the `DROP` inside `execute_batch`. `rusqlite-0.40.2/src/transaction.rs:254,315` route through `cmd("SAVEPOINT", false, name)` / `cmd("RELEASE", …)`, which quotes the identifier. The surviving-table query is value-bound, not merely error-bound: `resources.rs:427` reads the row into `sqlite_value`, which lands in the summary asserted as `sqlite=sqlite` at `examples/resource_lifecycle_runtime/src/main.sifr:54` and `positive/resource_close_aclose_matrix.sifr:48`. Under 0.39.0 the run fails either at the savepoint step or at the query — a genuine negative control.
- **Trust is still exact.** `libsqlite3-sys-0.38.2/Cargo.toml:18` is `links = "sqlite3"`, so the unchanged `sifr.toml` grants (`rust-build-scripts = ["aws-lc-sys","libsqlite3-sys"]`, `native-links = [… , "sqlite3"]`) remain exact and sufficient — no widening, no new grant. Toolchain is 1.98.0, well past the `cfg_select!` stabilization that forced the old 0.39.0 pin.
- **Lock diffs are minimal and consistent.** Root lock: only rusqlite version/checksum, `libsqlite3-sys` 0.37.0→0.38.2, and removal of `hashlink 0.11.1`, `sqlite-wasm-rs 0.5.5`, `rsqlite-vfs 0.1.1`. Fixture lock additionally collapses the now-single-version `foldhash`/`hashbrown` edges — correct Cargo output. Both locks hold exactly one `rusqlite` and one `libsqlite3-sys` line, and the checksums match the stated official values. No orphan or dangling references remain (`git grep sqlite-wasm-rs|rsqlite-vfs` hits only the new test's negative assertions).
- **Migration is complete.** All four declaration sites of the version/feature policy are updated (`crates/sifr_rust_interop_catalog/Cargo.toml:39`, fixture `Cargo.toml:12`, `_scenario_opaque_resources.py:72`, and the self-test mutation text), plus every policy record: `_matrix_inventory.py:134`, `rust_interop_fixture_matrix.json:131,243`, both `fixture.json` files, `internal_docs/rust_interop_architecture.md:2028`, `plans/phases/39_rust_interop.md:368`. `blocking_diagnostics` has no Cargo manifest (diagnostic-only), so its `fixture.json` policy is the whole surface. No stale `0.39.0` remains outside historical records.
- **Negative-control wiring is sound.** The new `default-features` mutation's expected substring `"rusqlite must set default-features = false"` is genuinely produced by `_scenario_async_reqwest.py:204-208`, and the previously-unchecked `default_features=None` for rusqlite is now `False`, so drift is actually enforced. `vendor/` is the sysroot-owned tree and vendors neither rusqlite nor libsqlite3-sys, so no re-vendoring was owed.
- **No panic/error regression.** The new code is `?`-propagating with `map_err` context; `and_then(rusqlite::Savepoint::commit)` consumes the savepoint so no drop-rollback interaction. No `unwrap`/`expect` added to fixture runtime code. New test file is 212 lines; guardrail intact.

### Non-blocking observations

1. **Lock certification is weaker than the precedent set one item earlier.** `crates/sifr_stdlib_manifest/tests/rusqlite_dependency_version.rs:105-133` locates packages by *name + version*, so a future duplicate `rusqlite 0.39.0` or `libsqlite3-sys 0.37.0` line would satisfy all three tests. The reqwest sibling written for the immediately preceding item asserts both single-line uniqueness per lock (`reqwest_dependency_version.rs:117-124`, `"must resolve one reqwest line"`) and that first-party `sifr*` packages' edges name the maintained line (`:95-106`). Cheap remediation: add a `filter(name == "rusqlite").len() == 1` assertion per lock and the same for `libsqlite3-sys`.
2. **The upstream tag commit `e88f112b…` is not recorded anywhere** (`git grep` returns nothing). The reqwest test anchors its release commit to `vendor/reqwest-0.13.4/.cargo_vcs_info.json`; rusqlite has no vendor copy, so there is no equivalent anchor. Fine as-is, but the provenance datum in the item's evidence has no durable home unless the follow-up record commit carries it.
3. **Two out-of-scope but behavior-preserving drive-bys in `resources.rs`**: the `Arc`/`atomic` + `sifr_runtime::interop` import reorder, and two nested-`if` → let-chain rewrites at `:566-578`. I confirmed the reorder fixes pre-existing style-edition-2024 drift (`rustfmt --style-edition 2024 --check` fails on the base copy, passes on the candidate), and I verified the let-chains preserve side-effect ordering — `task.finish().await` and `file.remove()` are still invoked unconditionally before the `cleanup_error.is_none()` guard, so first-error-wins semantics are unchanged. Harmless, but it is unrelated cleanup absorbed into a dependency item.
4. **Records follow the split convention, as expected.** `plans/issues/active/ad-hoc-latest-stable-release-convergence.md:109` still shows baseline `rusqlite 0.39.0` — consistent with the base commit itself being `docs: record Reqwest 0.13 upgrade`. The follow-up record should retire the now-obsolete rationale still asserted in `plans/issues/archive/rust-interop-runtime-ecosystem-certification.md:633-635` ("`0.40.1` selects `libsqlite3-sys 0.38.1`, whose build script uses unstable `cfg_select!`") and echoed in the active review `plans/reviews/active/rust-interop-certification-5-review-round17.md:83`; 0.38.2 on toolchain 1.98.0 resolves it.
5. `internal_docs/rust_interop_architecture.md:1364-1373` describes the opaque-resource bridge's exercised evidence but does not mention the savepoint-quoting proof, while `rust_interop_compatibility_matrix.json:159` and the fixture README do. Optional one-line alignment.

yaseralnajjar:
Item 19 exact-SHA gate evidence for `dd1e9b55c88ae60d21fea4335f155fe08cde53b4`:

- Exact-SHA Opus review: SATISFIED — https://github.com/sifr-lang/sifr/pull/3527#issuecomment-5414072098
- Focused certification: `sifr_stdlib_manifest` Rusqlite policy/runtime tests 3/3; full crate tests; Rust interop 10/10 plus 239 self-tests; generated lifecycle and alias-rejection runtimes; package, coverage, metadata, Clippy, formatting, documentation, HIR, and file-size checks all passed.
- Create-PR gate was invoked exactly once. All functional checks passed through runtime platform 28/28 with one expected skip. The process exited 124 only because the first cold-cache runtime-platform lane took 232147 ms against a 120000 ms blocking timing budget after the required 49 GiB target cleanup. It was not rerun.
- Merge gate was invoked exactly once on this SHA and passed with exit 0. The final E2E corpus passed 698/698. The report is `target/validation_lane_reports/merge.latest.json`; wall time was 7097.75 s after cold-cache rebuild. The warm-time budget miss is advisory only, with zero functional failures.

This is the approved and fully validated implementation candidate.

PR #3529, opened 2026-08-25
Upgrade SQLx certification to 0.9

yaseralnajjar:
Closes Item 20 of the latest-stable convergence phase.

- upgrades SQLx to exact 0.9.0
- replaces the removed combined runtime/TLS feature with runtime-tokio and tls-rustls-ring-webpki
- keeps the shared catalog PostgreSQL-only while the independently locked backend fixture owns query-macro certification
- validates exact official checksums, active feature edges, offline query metadata, and loopback execution

Exact candidate: fa88364ba1e39186d2af70a929d7c5df205c9144

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2483,13 +2483,12 @@ checksum = "d817e038c30374a4bcb22f94d0a8a0e216958d4c3dcde369b1439fec4bdda6e6"
 
 [[package]]
 name = "etcetera"
-version = "0.8.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "136d1b5283a1ab77bd9257427ffd09d8667ced0570b6f938942bc7568ed5b943"
+checksum = "de48cc4d1c1d97a20fd819def54b890cadde72ed3ad0c614822a0a433361be96"
 dependencies = [
  "cfg-if",
- "home",
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3094,11 +3093,11 @@ dependencies = [
 
 [[package]]
 name = "hashlink"
-version = "0.10.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7382cf6263419f2d8df38c55d7da83da5c18aef87fc7a7fc1fb1e344edfe14c1"
+checksum = "824e001ac4f3012dd16a264bec811403a67ca9deb6c102fc5049b32c4574b35f"
 dependencies = [
- "hashbrown 0.15.5",
+ "hashbrown 0.16.1",
 ]
 
 [[package]]
@@ -3130,20 +3129,11 @@ checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
 name = "hkdf"
-version = "0.12.4"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7b5f8eb2ad728638ea2c7d47a21db23b7b58a72ed6a38256b8a1849f15fbbdf7"
+checksum = "4aaa26c720c68b866f2c96ef5c1264b3e6f473fe5d4ce61cd44bbe913e553018"
 dependencies = [
- "hmac 0.12.1",
-]
-
-[[package]]
-name = "hmac"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c49c37c09c17a53d937dfbb742eb3a961d65a994e6bcdcf37e7399d0cc8ab5e"
-dependencies = [
- "digest 0.10.7",
+ "hmac",
 ]
 
 [[package]]
@@ -3991,10 +3981,7 @@ version = "0.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c943259e342f1e06ff2da7a83eabdfe7f92ce10262688dbf1895ff0b3e6e4652"
 dependencies = [
- "bitflags 2.13.1",
  "libc",
- "plain",
- "redox_syscall 0.9.0",
 ]
 
 [[package]]
@@ -4614,7 +4601,7 @@ checksum = "2621685985a2ebf1c516881c026032ac7deafcda1a2c9b7850dc81e3dfcb64c1"
 dependencies = [
  "cfg-if",
  "libc",
- "redox_syscall 0.5.18",
+ "redox_syscall",
  "smallvec",
  "windows-link 0.2.1",
 ]
@@ -4807,12 +4794,6 @@ name = "pkg-config"
 version = "0.3.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f6b464fbc74e149a392436b17d523f769e057cb6877f6a5c4618bc6f11800548"
-
-[[package]]
-name = "plain"
-version = "0.2.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "planus"
@@ -5468,7 +5449,7 @@ dependencies = [
  "byteorder",
  "bytes",
  "fallible-iterator 0.2.0",
- "hmac 0.13.0",
+ "hmac",
  "md-5 0.11.0",
  "memchr",
  "rand 0.10.2",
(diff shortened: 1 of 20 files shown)


yaseralnajjar:
Reviewed the full diff `4f0bba8fb..d727e73c3` read-only, plus independent verification of the lock claims.

Independent checks I ran:
- Exact lock-identity diff: fixture lock minus root lock is exactly the five claimed identities (`sqlx-mysql 0.9.0`, `sqlx-sqlite 0.9.0`, `flume 0.12.0`, `spin 0.9.9`, `libsqlite3-sys 0.37.0`); `chacha20 0.10.0`, `getrandom 0.4.2`, `whoami 2.1.2`, `wasip2 1.0.3+wasi-0.2.9`, `ring 0.17.14`, `rustls 0.23.43`, `webpki-roots 1.0.9` now match root exactly. The round-1 blocker is resolved.
- The exception set in `_scenario_lock_checks.py:13-44` equals that five-identity set with no surplus, and the new `allowed_fixture_only - scenario_packages` loop makes stale tuples fail closed.
- `cargo tree --locked --offline -e features` in the fixture: no `sqlx-mysql`, `sqlx-sqlite`, `flume`, `libsqlite3-sys`, or legacy alias (`runtime-tokio-rustls`/`tls-rustls`/`tls-rustls-ring`) in the active graph; `runtime-tokio` + `tls-rustls-ring-webpki` + `postgres` + `macros` (offline) all active.
- Root lock `sqlx 0.9.0` edges are exactly `sqlx-core`, `sqlx-macros`, `sqlx-postgres`; no MySQL/SQLite driver in root; `libsqlite3-sys` stays at 0.38.2 for Rusqlite.
- Removing the legacy-alias loop in `sqlx_dependency_version.rs:144` is safe: `assert_eq!(features, expected_features)` is exact and neither `CATALOG_FEATURES` nor `FIXTURE_FEATURES` contains an alias.
- No maintained surface still names `0.8.6` or `runtime-tokio-rustls` (only archived/historical `plans/` review records, and the new negative assertion list). All touched files are well under 900 lines.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- mechanism-level defect (new, for later-item recording): `verification/areas/rust_interop/checks/_scenario_lock_checks.py:13` — at base `4f0bba8fb` the fixture lock was an exact subset of the root lock (0 differing identities), so root-only `cargo fetch --locked --offline` (`_crate_catalog.py:44`) transitively guaranteed the fixture's offline resolve. The five exempted identities are now outside that guarantee. They are never compiled, so no `.crate` download is needed, but they are real edges of `sqlx 0.9.0` in the fixture lock (`sqlx` deps list includes `sqlx-mysql`/`sqlx-sqlite`), and `--locked` only succeeds because dependency resolution reproduces them — which requires their sparse-index summaries (`sqlx-mysql`, `sqlx-sqlite`, `flume`, `spin`, `libsqlite3-sys 0.37.0`) in the registry index cache. Root resolution never activates those optional deps, so a cache primed only by root `cargo fetch` plausibly lacks those index entries and the fixture's `--locked --offline` resolve would fail. I could not falsify this here without a clean `CARGO_HOME` (needs network). Worth recording as a targeted clean-cache experiment plus, if confirmed, an explicit fixture-level prefetch/index-coverage step.
- suggestion: `_crate_catalog.py:12` `CATALOG_FEATURE_OVERRIDES` hardcodes the catalog's SQLx vector separately from `_matrix_inventory.py:135` `EXPECTED_FEATURE_POLICIES`. It is mutation-tested, but the two lists must now be edited in tandem with no check binding one to the other plus the documented reason; deriving the override as "matrix policy minus `macros`" would keep the divergence self-explaining.
- pre-existing issue: archived and active `plans/reviews/**` records still cite `sqlx 0.8.6` and `runtime-tokio-rustls`. They are historical review artifacts, not maintained policy surfaces, so no change is required.

yaseralnajjar:
I inspected the exact diff, both lock graphs, and the new checks.

## What I verified

- Catalog (`crates/sifr_rust_interop_catalog/Cargo.toml:46`) and fixture (`.../backend_feature_package/Cargo.toml:6`) both pin `=0.9.0`, `default-features = false`, split `runtime-tokio` + `tls-rustls-ring-webpki`; fixture retains `postgres` + `macros`. No `runtime-tokio-rustls` survives except as negative assertions.
- Root `Cargo.lock` contains no `sqlx-mysql`/`sqlx-sqlite`, keeps `libsqlite3-sys 0.38.2` for Rusqlite, and `sqlx-core` resolves `rustls` + `webpki-roots 1.0.9` with `ring` (no aws-lc in the fixture graph). Non-SQLx root-lock churn is all net removal of SQLx 0.8 transitives (`hmac 0.12.1`, `md-5 0.10.6`, `rand 0.8.6`, `whoami 1.6.1`, `webpki-roots 0.26.11`, `windows-* 0.48.5`) plus one benign `tempfile → getrandom` edge re-unification.
- Fixture lock confirms the stated conflict empirically: `macros` adds `sqlx → sqlx-mysql, sqlx-sqlite` lock edges pulling `libsqlite3-sys 0.37.0`, which cannot coexist with Rusqlite's `0.38.2` native link.
- New mutation cases and the `sifr_stdlib_manifest` test are consistent with the existing per-crate test pattern and dev-deps.

## Finding

The root-lock-subset allowlist is not narrow. Diffing the two locks, exactly eight fixture identities are absent from the root lock; five (`sqlx-mysql`, `sqlx-sqlite`, `flume 0.12.0`, `spin 0.9.9`, `libsqlite3-sys 0.37.0`) are the genuinely inactive SQLite/MySQL graph. The other three are plain patch drift on **actively compiled** PostgreSQL-path crates — root has `chacha20 0.10.0`, `getrandom 0.4.2`, `whoami 2.1.2`; the fixture pins `0.10.1`, `0.4.3`, `2.1.3`. All three are semver-compatible with the fixture's requirements, so they could have been pinned back.

Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `verification/areas/rust_interop/checks/_scenario_lock_checks.py:15` (`chacha20 0.10.1`), `:21` (`getrandom 0.4.3`), `:57` (`whoami 2.1.3`), criterion "Exact fixture-only lock identities are narrowly allowlisted" and "preserve hermetic no-database operation". These three are not query-macro MySQL/SQLite edges — they are activated `sqlx-postgres`/`rand` dependencies that the root lock already carries at compatible lower patches (`0.10.0`, `0.4.2`, `2.1.2`). The comment above the set misdescribes them, and allowlisting them breaks the property the subset check exists to guarantee: `cargo fetch --locked` at the repo root (`verification/runner/sifr_verify/cargo_setup.py:9`, the only cache-warming step before `CARGO_NET_OFFLINE=true`) fetches only root-lock identities, so on a cold cache the fixture's offline build has no source for `whoami 2.1.3`, `chacha20 0.10.1`, or `getrandom 0.4.3`. The inactive SQLite/MySQL entries do not have this problem because they are never compiled or downloaded. Correction: `cargo update --precise` these three back to the root-lock identities in the fixture lock, regenerate, and delete their three tuples from `SQLX_QUERY_MACRO_FIXTURE_ONLY_LOCK_PACKAGES`, leaving only the five inactive-driver identities.

Follow-up findings:
- suggestion: only one of the eight allowlist entries (`sqlx-sqlite`'s checksum) is mutation-tested in `_scenario_backend.py:145`; nothing detects a stale/unused allowlist tuple after a future lock regeneration. A "every allowlisted identity must appear in the fixture lock" assertion would keep the set from silently accumulating.
- suggestion: `assert_dependency` in `crates/sifr_stdlib_manifest/tests/sqlx_dependency_version.rs:147` checks the legacy-feature list after an exact `assert_eq!` on the full feature vector, so the loop is unreachable-by-construction; harmless, but it reads as coverage it does not add.
- suggestion: `CATALOG_FEATURE_OVERRIDES` in `verification/areas/rust_interop/checks/_crate_catalog.py:13` silently exempts `sqlx` from the matrix feature policy. Worth a self-test case pinning the override's contents so the catalog/matrix divergence cannot widen unnoticed.

yaseralnajjar:
Item 20 gate evidence

Final candidate: `ff2c7c093214d80d5146497b1cf795d68c0f00f7`

- Exact-SHA Opus review: initial review on `fa88364ba1e39186d2af70a929d7c5df205c9144` found the fixture/root cache-identity drift; remediation review on `d727e73c3870851ac1a8dce4c797b995c3737d4d` was SATISFIED. Per the two-review cap, no third review was run for the later coverage-classification-only commit.
- Create-PR gate: the one allowed invocation ran on `d727e73c3870851ac1a8dce4c797b995c3737d4d`. All SQLx/Rust-interop checks passed, then the coverage matrix rejected the new `sqlx_dependency_version` test target because its classification was absent. The gate was not rerun.
- Readiness after classification fix: the Rust-interop matrix and all 241 policy self-tests passed on the final candidate, including mutation checks for the SQLx inactive-lock allowlist and workspace catalog feature policy.
- Merge gate: the one allowed invocation ran on final candidate `ff2c7c093214d80d5146497b1cf795d68c0f00f7` and passed with exit 0. Wall time: 7,531.60s. Full E2E: 698 passed, 0 failed; report signature `127353b213e16688`; 173 groups; 0/173 cache hits.
- Merge-gate advisories: warm wall-time budget exceeded and group skew was high. These are performance telemetry advisories; all blocking checks passed.

Second-review follow-up deferred to the final closure item as required: validate the five inactive SQLx fixture-lock sparse-index summaries with a clean `CARGO_HOME` experiment.

PR #3531, opened 2026-08-25
Upgrade Itertools certification to 0.15

yaseralnajjar:
## Summary
- advance the maintained workspace Itertools policy to 0.15.0 with explicit std-only features
- vendor the official 0.15.0 release while retaining DataFusion 54's checksum-locked 0.14.0 source until Item 22
- certify new 0.15 APIs, lock ownership, vendor hashes, Ruff alignment, and iterator parity

## Validation
- cargo test -p sifr_stdlib_manifest --locked
- cargo test -p sifr_format --locked
- cargo test -p sifr --locked -- --skip test_e2e_pass
- selected 10-fixture iterator/itertools E2E pass corpus
- cargo clippy --workspace -- -D warnings
- cargo fmt --check
- Python maintainability and file-size guardrails
- git diff --check

Draft pending exact-SHA Opus review and project gates.

PR #3533, opened 2026-08-26
build: upgrade Arrow 59 and DataFusion 55

yaseralnajjar:
## Summary
- upgrade the coupled Arrow ecosystem to 59.2.0 and DataFusion ecosystem to 55.0.0
- adopt DataFusion 55's borrowed fill_nan API in the maintained advanced-data bridge and propagate catalog errors
- certify exact analytical package families, checksums, lock edges, runtime behavior, and policy mutations

## Validation
- Arrow/DataFusion dependency certification: 3 passed
- Itertools dependency certification: 4 passed
- advanced-data fixture cargo check --workspace --locked: passed
- advanced-data Rust-interop matrix: 2 variants and 242 mutation cases passed
- exact generated positive and negative advanced-data tests: passed
- stdlib manifest suite: passed
- coverage readiness: 4 variants passed
- Rust interop: 10 variants passed
- Python Arrow examples/runtime focused certification: passed
- cargo test -p sifr --locked -- --skip test_e2e_pass: passed
- cargo clippy --workspace -- -D warnings: passed
- cargo fmt --check: passed
- HIR maintainability and file-size guardrails: passed
- git diff --check: passed

Exact candidate: 61392c034d6c30216cc5a6bd2ceb67cfa01e925b

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,9 +177,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "58.3.0"
+version = "59.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "378530e55cd479eda3c14eb345310799717e6f76d0c332041e8487022166b471"
+checksum = "61d285d16bce7d0be61912f7928342b673067b6b7d7ef6cc179258ba7de1fecf"
 dependencies = [
  "arrow-arith",
  "arrow-array",
@@ -198,9 +198,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-arith"
-version = "58.4.0"
+version = "59.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a41203398f0eaa6f7ec8e62c0da742a21abf282c148fc157f6c35c90e29981a"
+checksum = "757ef1836251e88222542a7da2623bc1c9cb9e20afefa6db2c41e79991cd91d4"
 dependencies = [
  "arrow-array",
  "arrow-buffer",
@@ -212,9 +212,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-array"
-version = "58.4.0"
+version = "59.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ae33dad492b7df00a217563a7b0ef2874df68a0deea1b1a3acf628152f7f7a69"
+checksum = "bc9a4a4b2b5ecd0e04df03471661cb61f28bed3c7fd50994715129b01b2edb97"
 dependencies = [
  "ahash 0.8.12",
  "arrow-buffer",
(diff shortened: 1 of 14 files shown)


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- suggestion: `crates/sifr_stdlib_manifest/tests/arrow_datafusion_dependency_versions.rs:188` asserts only `BRIDGE_SOURCE.contains(".map_err(display_error)?")`, which is already satisfied by the pre-existing `register_table(...).map_err(display_error)?` at `record_batch.rs:107-108`. The intended guarantee — that `table_exist("input")` propagates instead of collapsing to a value — rests solely on the negative `!contains("unwrap_or(false)")` check. Anchoring the assertion to the `table_exist` expression itself (e.g. the `.table_exist("input")\n        .map_err(display_error)?` chain) would pin the actual error path.
- suggestion: `verification/areas/rust_interop/checks/_scenario_advanced_data.py:26` adds the bare token `'map_err(display_error)?'` to `ADVANCED_DATA_SCENARIO_TOKENS`; it was already present in the file before this item, so it adds no detection power. The mutation self-test list gained a `fill_nan` drift case but no case reverting `table_exist` propagation to `unwrap_or(false)`, so that regression is only caught by the manifest test's negative substring check.
- suggestion: the `nanvl` plan invariant is checked twice — once as a hard error in `record_batch.rs:141-143` and again in the `observe_state` mismatch condition at `record_batch.rs:170`. Since `build_state` already rejects a plan without `nanvl`, the stored-plan re-check in `observe_state` can never fire; keeping one of the two would be simpler without weakening the evidence.
- pre-existing issue: root `Cargo.lock` still carries `itertools 0.13.0` alongside 0.14.0 and 0.15.0. It is untouched by this diff and outside Item 22 (which only owned DataFusion's direct edge, now on 0.15.0, with Object Store externally owning 0.14.0 — both correctly asserted in `itertools_dependency_version.rs:347-368`).
- pre-existing issue: `plans/issues/active/ad-hoc-latest-stable-release-convergence.md:112` still lists the analytical-stack row in its pre-upgrade state. Per repo convention (`docs: record Itertools 0.15 completion (#3532)` followed `#3531`), completion is recorded in a separate post-merge docs commit, so this is not an omission of the candidate.

Verification notes on the mechanism itself: both locks converge all 14 Arrow-family crates and Parquet on 59.2.0 and all 31 DataFusion-family crates on 55.0.0, with the audited checksums present in each; the only package churn is the removal of `thrift`/`ordered-float`/`integer-encoding` (dropped by Parquet 59) and the fixture lock picking up `base64 0.23.1` plus the second `itertools` entry — all transitive, no hand-edited versions, no unrelated bumps. `DataFrame::fill_nan(&ScalarValue, &[&str])` is genuinely new in DataFusion 55 (54.1.0 exposes only `fill_null`), so the adoption is real rather than a rename; the bridge's plan assertion and the `table_exist` error propagation both return typed errors with no `unwrap`/`expect` added.

yaseralnajjar:
Create-PR gate evidence for exact candidate `61392c034d6c30216cc5a6bd2ceb67cfa01e925b`:

- One authorized create-PR gate invocation: `scripts/run_all_tests.sh --profile create-pr`
- Exit: `124`; elapsed: 1,152.22 seconds
- Functional result: all executed steps passed. Rust interop passed 10/10 variants; coverage matrix 4/4; diagnostics 5/5; Python interop 19/19; runtime platform 28/28 with one declared capability skip.
- Blocking result: the cold-cache runtime-platform step took 219.467 seconds against its 120-second wall-time budget, after all functional variants passed.
- Resource result: max RSS 1.4 GiB; swaps 0.
- The gate began and ended on the exact reviewed SHA.

Per the phase contract, this one-shot gate will not be rerun.

yaseralnajjar:
Merge gate evidence for exact candidate `61392c034d6c30216cc5a6bd2ceb67cfa01e925b`:

- The one authorized merge gate invocation, `scripts/run_all_tests.sh`, exited `0`.
- Wall time: 6,809.08 seconds; max RSS: 2.4 GiB; swaps: 0.
- Every functional step passed, including exact Arrow/DataFusion certification, 10 Rust-interop variants, 25 Python-interop variants, 76 generated-build integrations, sysroot release qualification, and all other merge-profile areas.
- E2E: 698/698 fixtures passed across 173 cold-cache groups with signature `127353b213e16688`.
- Advisory observations only: warm wall-time budget exceeded, generated/e2e cache footprint, and fixture-group skew.
- The gate began and ended on the exact Opus-approved SHA. It will not be rerun.

PR #3541, opened 2026-08-26
Upgrade CFFI and Cryptography

yaseralnajjar:
## Item 26

Advance CFFI first from 1.17.1 to 2.1.1, validate the intermediate ABI state with Cryptography fixed at 45.0.7, then advance Cryptography to 50.0.1.

Adopt CFFI 2.1's cffi.gen_src source generator and Cryptography 50's stable X.509 verification API. The certificate fixture validates a trusted server path and rejects the same certificate for the wrong hostname. No legacy or fallback path is retained.

Generalize the stable Python release audit to cover nine completed package updates and selected PyPI artifact hashes. Register the new crypto ABI suite in every delivery profile.

## Validation

- Sequential intermediate CFFI callback and library checkpoint passed before Cryptography changed.
- Final CFFI callback, stable audit, crypto ABI feature, tier-1, callback policy, and compiled library suites pass.
- CFFI source generation, Fernet, stable X.509 path verification, wrong-host rejection, Certifi trust loading, and OpenSSL 4.0.2 wheel loading pass.
- Live PyPI audit reports nine packages current with nine matching artifacts.
- Both lock checks, runner/profile self-tests, coverage/profile matrices, Ruff, JSON, HIR maintainability, file-size, and diff checks pass.

No compiler input changed, so the phase rules prohibit Sifr create-PR and merge gates for this item.


yaseralnajjar:
### Exact-SHA Opus review — candidate 0ac31643c4e7b2357694836980f5f15c97bc83d5

Verdict: SATISFIED

Blocking findings:
- none

Verification I performed on the exact SHAs (read-only): confirmed the lock diff touches only `cffi 1.17.1→2.1.1` and `cryptography 45.0.7→50.0.1` plus the regenerated `pycparser` marker and the two direct specifiers, with no incidental transitive drift; confirmed the audited artifact filenames/SHA-256s in `data/latest_stable_python.json` (`cffi:661c298b…`, `cryptography:b8f852c6…`) are byte-identical to entries in `uv.lock:116` and `uv.lock:189` and that `dependency_versions.validate_project` machine-checks version + filename + `sha256:` digest per package, with the four mutation self-tests still targeting existing packages (`polars`, `torch`, `schwifty`, LocalStack); confirmed the rename left zero stale `latest_stable_minor_train` / "Item 25" references and that `crypto-abi-features` is registered in exactly the same seven locations as its sibling suite and appears exactly once per delivery profile under `python_interop`; confirmed installed env reports cffi 2.1.1, cryptography 50.0.1, OpenSSL 4.0.2 25 Aug 2026; ran `runner/crypto_abi_features.py` (exit 0, arg order matches `cffi._cffi_gen_src` `read-sources` signature: module_name, cdef, csrc, output); ran the Cryptography bridge and the CFFI callback bridge under `-W error::DeprecationWarning` — both clean, so the adopted `x509.verification` path and the ABI-mode `ffi.callback` path use non-deprecated current APIs (forward-only, no shim or fallback anywhere in the diff). The issue-table edit lowers the recorded baseline from `cryptography 46.0.0` to `45.0.7`, which the base-SHA lock confirms as the version actually installed — a correction toward accuracy, not a rewrite as a newer version.

Follow-up findings:
- suggestion: `verification/areas/python_interop/runner/crypto_abi_features.py:35` validates `cffi.gen_src` output by substring containment only. Generated C source is never compiled or loaded, so a syntactically broken but substring-matching emission would pass. A later item could compile the generated module to close the loop; ABI-mode callbacks are currently the only compiled CFFI coverage.
- pre-existing issue: `requires_python` in `data/latest_stable_python.json` (all nine packages, not just the two new ones) is recorded but never machine-checked against `pyproject.toml`'s `requires-python = "==3.14.7"`; the audit would not catch a package whose published metadata drops 3.14.
- suggestion: `pyproject.toml` now mixes exact-lower pins (`cffi>=2.1.1`, `cryptography>=50.0.1`) with the file's prevailing loose bounds (`boto3>=1.34,<2`, `certifi>=2024,<2027`). The tight form is stronger for forward-only intent; normalizing the convention repo-wide would be worth a dedicated item rather than per-package drift.
- suggestion: the suite name `crypto-abi-features` exercises only CFFI; the Cryptography 50 API adoption lives in the library-examples bridge. If a later item adds Cryptography-specific feature coverage, folding it into this suite (or renaming to `cffi-abi-features`) would keep the name honest.
- infrastructure issue: `plans/issues/.../ad-hoc-latest-stable-release-convergence.md:198` still shows Item 26 as `pending` and carries no Item 26 evidence block. This matches the Item 25 precedent (separate record commit, e.g. `f122c42c0`), so it is not an omission for this PR — but the record item must capture the sequential intermediate-state evidence, which exists only in the PR narrative today, not in a durable file.

PR #3547, opened 2026-08-26
build: update NumPy and Pandas stable lines

yaseralnajjar:
## Summary

- update both maintained NumPy locks to 2.5.2, then update Pandas to 3.0.5
- adopt NumPy descending sort and Pandas 3 string, copy-on-write, public-module, and column-expression behavior
- extend exact release auditing and all offline delivery profiles

## Validation

- sequential NumPy feature, dataframe, buffer, and DLPack checks
- Pandas 3 feature and compiled dataframe checks
- compiled Arrow stream certification after the Pandas 3 public-module migration
- uv lock checks for both projects
- Python interop runner self-tests and focused suites
- profile and area schema checks
- Ruff formatting and selected correctness checks
- git diff --check

Exact-SHA Opus review and the one permitted create-PR and merge gates will follow on the approved compiler-fixture candidate.


yaseralnajjar:
SATISFIED

**Scope and mechanism verified (read-only; `git status` clean throughout).**

Version ownership across both locks
- `demos/m12_dlpack_demo/uv.lock` and `verification/areas/python_interop/uv.lock` both resolve `numpy 2.5.2`, single package entry each (no marker split remains; base was 2.4.6 / 2.5.1). `python_interop` lock resolves `pandas 3.0.5`.
- Locked cp314 macOS arm64 wheel digests match `data/latest_stable_python.json` exactly: numpy `24b9dc2e…dbda8`, pandas `53730687…8007c`. sdist upload-times align with the recorded `released_at`.
- `uv lock --check --offline` passes for both projects (`Resolved 90` / `Resolved 31`), so pyproject floors `numpy>=2.5.2,<3` / `pandas>=3.0.5,<4` and the locks are consistent.
- `dependency_versions.py` now owns both lock owners: `EXPECTED_PROJECTS["dlpack-demo"] = {numpy, torch}`, `release_map` enforces `releases.keys() == union(projects)`, and `validate_repository` runs `validate_project` over both pyproject/lock pairs, checking version *and* audited artifact filename+hash. Ran it: `projects=2 packages=18 locks=2 images=2 mutations=6`. The new sixth mutation degrades **the dlpack lock's** numpy to 2.5.1 and asserts rejection — a genuine second-owner negative, not a copy of the primary one.
- No `numpy 2.4.6` / `2.5.1` / `pandas 2.3.3` residue anywhere outside the phase plan's baseline table and unrelated vendor text.

Feature coverage is real, not a false positive (executed against the area venv, `numpy 2.5.2` / `pandas 3.0.5`)
- `runner/numeric_dataframe_features.py` passes and each probe fails closed: `np.sort(..., descending=True)` → `[3,2,1]`, `np.argsort(..., descending=True)` → `[0,2,1]`, `str(dtype) == "str"` with `pd.isna` on the missing entry, `pd.col("value")*2` → `[4,6,10]`, plus exact `importlib.metadata.version` pins. Removing any new API raises `TypeError`/`RuntimeError` rather than silently passing.
- The copy-on-write proof is a real mutation, not a no-op: `cities.iloc[0] = "bergen"` observably changes the Series (`bergen`) while `frame.iloc[0,0]` stays `"oslo"`. Under pandas 2 semantics the assertion would fail. Same proof is embedded in the compiled bridge `fixtures/pandas_arrow/python_bridges/pandas_example.py`, whose `raise RuntimeError` propagates through `pandas_full_example.sifr`; the stdout marker is unchanged, so the compiled evidence contract is untouched while the new behavior gates it.
- `fixtures/numpy_buffer/python_bridges/numpy_example.py` adds a `descending=True` flatten-sort assertion to the compiled buffer path with the marker unchanged.
- Both fixtures and the new runner produce zero warnings under `python -W error`, so no deprecation is being absorbed.

Arrow fixture expectation change is a fact, not a mask
- `arrow_declaration_compiled.sifr` now expects `pandas_stream.producer_module() == "pandas"`. Confirmed empirically: `pd.DataFrame.__module__ == "pandas"` under 3.0.5 (pandas 3 public-module identity). The assertion still fails if identity drifts, so this is the "public module identity exercised" criterion rather than a weakened expectation. `arrow_evidence.py` derives `producer_module` dynamically (`type(value).__module__`), `arrow_capsule_contract.json` keys off root `pandas`/capsule names, and no stored certification, contract, or evidence JSON still hardcodes `pandas.core.frame`. The only remaining `pandas.core.frame` string is a synthetic exporter class in `crates/sifr_runtime/src/python/arrow_ops/tests.rs:412`, which is self-contained and correctly untouched.

Wiring, profiles, docs, guardrails
- `manifest.json` adds the `numeric-dataframe-features` adapter suite (`expect_exit_code: 0`, no diagnostic formats, no skip predicate); `runner.py` registers the command in `COMMAND_ARGS` and in `AREA_PROJECT_COMMANDS`, so it executes via `uv run --project … --locked` — lock drift fails the case rather than silently re-resolving.
- Suite added to `create-pr`, `merge`, `nightly`, `release`. `validate_selected_area_suites` passes for all five profiles; area runner self-test (`runner/run.py --self-test`) passes.
- README changes are accurate (six mutations covering both locks; NumPy descending sort; pandas str dtype / CoW / `pd.col`).
- No fallback, version branch, try/except-import, or skip introduced anywhere in the diff. Python 3.14.7 remains the sole lane (`requires-python` untouched in both manifests). All changed first-party files are far below 900 lines (largest: `runner.py` at 397; new module 42).

Non-blocking observations
1. `numeric_dataframe_features.py` does not itself assert `pd.DataFrame.__module__ == "pandas"`; public-module identity is proved only by the compiled Arrow fixture. Adding one line there would keep the proof in the fast adapter suite too.
2. Unlike its sibling `redis_service_features.py`, the new suite doesn't install `warnings.simplefilter("error", DeprecationWarning)`. Verified clean under `-W error` today, so nothing is hidden, but the sibling's stricter posture would be cheap to match.
3. The audit's `requires_python` fields for numpy/pandas are recorded but not machine-checked against the 3.14.7 lane — pre-existing pattern for all 18 packages, not introduced here.
4. Sequential ordering (NumPy in both locks first, then Pandas) is squashed into one commit `6c2a6ac7f`, so it rests on the stated staged evidence rather than commit topology; the phase rule explicitly permits merging only the coherent final unit.

yaseralnajjar:
Exact-SHA validation evidence for `6c2a6ac7f577d60e9858d77efa46c3f3d5787145`:

- Opus review: SATISFIED (one review, no remediation).
- Create-PR gate: ran once. All functional checks passed, including 24 Python interop variants with zero failures. The command exited 124 only because the required private-target cleanup made the cold runtime-platform lane exceed its warm 120-second budget (178,734 ms). No rerun.
  - log SHA-256: `281f61dfb4504535456666f6dcae12a247006476ce2661a8777073f862d52677`
  - report SHA-256: `aaf1055ec7bc3ec6ab690dec8e23ab82ae03977483b83f2f429bfd160bf75fe3`
- Merge gate: ran once and exited 0. All lanes passed, including 698 E2E fixtures with zero failures. Cold-cache wall time was advisory only.
  - log SHA-256: `cc62e3fd564a07d76c08194f53d7fa54429ec17630f42c58377db1a974c7fc8f`
  - report SHA-256: `0972960098cb50be748d9b4e66f2ee91e8af8dd618a4364554b940cdc9c883de`

No gate was rerun.

PR #3551, opened 2026-08-26
build: update Kafka Python stable release

yaseralnajjar:
## Summary

- update kafka-python from 2.3.2 to the latest stable 3.0.11 release
- delete the removed `api_version_auto_timeout_ms` configuration with no legacy path
- certify the 3.x generated DescribeCluster protocol schema in the compiled callback example
- add exact stable-audit, lock-artifact, offline callback, and real Redpanda service evidence

## Focused validation

- unchanged 3.0.11 offline callback passed before feature edits
- live Kafka initially failed on the removed timeout configuration; after its deletion, the real Redpanda producer/consumer/callback round trip passed
- final stable audit: 2 projects, 20 packages, 2 locks, 2 images, 8 negative mutations
- final offline callback example passed with the exact 3.0.11 generated-schema marker
- final live suite passed all 6 services with 0 failures and 0 skips
- Python environment, package, callback, live policy, runner self-tests, uv lock, pinned Ruff 0.16.4, Cargo format, HIR, file-size, diff, and no-pre-v1 compatibility checks passed

Compiler fixture content changed, so the exact reviewed candidate will receive one create-PR gate and one merge gate.


yaseralnajjar:
Exact-SHA Opus review for `dbdbd42915dd45fe0255681c224266dd08f453ea`
against base `2808a84cda557623a278703d6ae59223b172879d`.

- Prompt SHA-256: `99c6c868d01c67724eba2bc26bbed1070cf4536e5402a62e7bad3c67e7f00fb6`
- Response SHA-256: `496b52730760af83bc0027c0b5cb980cae0fffb37a016cc921c278c0120d919c`
- Review round: initial and only

SATISFIED

## Verification performed

Opus inspected the exact range (13 files, one commit) and independently reran
the stable dependency audit. It returned `projects=2 packages=20 locks=2
images=2 mutations=8`.

It executed the offline bridge directly. The result matched the exact
`version=3.0.11:schema=DescribeClusterRequest-v2:fields=3:ack=42` marker in all
four owners. It also changed the three generated-schema input values and
confirmed that decode returned those values. This proves the round trip is not
a default-value echo.

It confirmed the installed 3.0.11 distribution metadata, Python requirement,
universal-wheel tag, upload time, and wheel hash. These values agree with the
stable audit and uv lock.

It found no remaining removed timeout option, Kafka Python 2.x constraint, or
stale callback marker. It confirmed the live `resources=zero` marker is gated
by actual before-and-after resource diagnostics. Both callback bridges rethrow
worker errors on the declaring thread, fail on timeout, require exactly one
result, and cannot access an empty result list.

## Blocking findings

None.

## Non-blocking observations

1. The old Redpanda broker image is outside the current service-image stable
   audit. This is a separate convergence mechanism for Item 35.
2. A timed-out worker thread is non-daemon and can outlive the call. This is a
   pre-existing pattern and did not violate this item's fail-closed criterion.
3. Pinned Ruff created small import-order changes in two touched files.
4. Two declarative Kafka contract names remain pre-existing and have no
   implementation. They are outside this client-release item.

yaseralnajjar:
Exact-SHA gate evidence for `dbdbd42915dd45fe0255681c224266dd08f453ea`:

- Create-PR gate: invoked exactly once. Every functional check passed, including 24 Python interop variants and 28 runtime-platform variants. The command exited 124 only because the cold runtime-platform area took 183329 ms against its 120000 ms warm-run budget. It was not rerun.
  - log SHA-256: `6750a0cb26ae68ffc6f50b573f34c360453c4e3a1a2ac270e8211ac4d22de5cd`
  - report SHA-256: `675d0b303353c26d2d60d627bca03b7bd990fbaba3a378de0ad1cae7191fce56`
- Merge gate: invoked exactly once. Kafka and every completed functional area passed, including 30 Python interop variants, 183 diagnostics variants, 30 runtime-platform variants, 42 developer-tooling variants, generated-code quality, performance, and 67 of 68 distribution variants. It exited 1 solely because `plans/releases/single-maintainer-approval-waiver.json` expired at `2026-08-27T00:00:00Z`. The failing case was `schema-v2-preview-epoch-bootstrap`; the gate stopped before later crate and E2E stages. It was not rerun.
  - log SHA-256: `b9e004ebc8b2ff494cdab8f8642a0e28759e88d4b0fbc8e652c1520a85fe1069`
  - report SHA-256: `cbdb39d9047da8206ebcba10e2c6856beaffaec314e4487413580b67a67181cd`

This failure is owned by `plans/issues/active/ad-hoc-distinct-release-reviewer-restoration.md`. Its acceptance requires a distinct human release reviewer and protected-environment configuration. Item 31 will remain draft and unmerged. The waiver will not be extended, bypassed, or converted into a fallback.

PR #3554, opened 2026-08-27
M2: unify test and build materialization

yaseralnajjar:
## Summary

- route generated binary builds and `sifr test` through one Cargo-project materializer and executor
- preserve Cargo lock modes, resolution preparation, hermetic environment, invocation tracing, bridge sources, namespace layout, and native-link checks for tests
- execute tests from invocation-owned copies so reusable source-cache entries remain immutable

## Validation

- `cargo test -p sifr_driver test_runner --no-fail-fast`
- `cargo test -p sifr_driver test_run_tests_reuses_cached_workspace_for_unchanged_project -- --nocapture`
- `cargo test -p sifr test_command_preserves_frozen_cargo_resolution_mode`
- `cargo test -p sifr_driver build::generated_cargo_project::tests`
- `cargo test -p sifr_driver test_run_tests_support_module_named_main_imports_root_owned_unions`

The optional all-codegen/all-driver run was stopped after the codegen suite completed because unrelated cold nested Cargo probes remained in contention; it is not claimed as passing evidence. Full compiler gates and merge remain deferred behind the phase-wide human-review dependency recorded by M1.

### DIFF
--- a/crates/sifr/src/check_and_package_commands.rs
+++ b/crates/sifr/src/check_and_package_commands.rs
@@ -384,11 +384,15 @@ pub(super) fn emit_success_message(diagnostic_format: DiagnosticFormat, message:
     }
 }
 
-pub(super) fn cmd_test(dir: &Path, diagnostic_format: DiagnosticFormat) -> i32 {
+pub(super) fn cmd_test(
+    dir: &Path,
+    lock_mode: sifr_package::CargoLockMode,
+    diagnostic_format: DiagnosticFormat,
+) -> i32 {
     let mut provider = DiskSourceProvider::new();
     let run_result = match run_with_panic_boundary(
         "internal compiler panic during test command execution",
-        || run_tests(dir, &mut provider),
+        || run_tests(dir, &mut provider, lock_mode),
     ) {
         Ok(result) => result,
         Err(internal) => return render_diagnostics(&[*internal], diagnostic_format),
--- a/crates/sifr/src/cli_model_and_entrypoint.rs
+++ b/crates/sifr/src/cli_model_and_entrypoint.rs
@@ -325,6 +325,15 @@ pub(crate) enum Commands {
         /// Directory containing test files (default: current directory)
         #[arg(default_value = ".")]
         dir: PathBuf,
+        /// Require Cargo.lock to remain unchanged
+        #[arg(long)]
+        locked: bool,
+        /// Disallow network access while resolving Rust dependencies
+        #[arg(long)]
+        offline: bool,
+        /// Combine --locked and --offline
+        #[arg(long)]
+        frozen: bool,
     },
     /// Manage a standalone Sifr installation
     #[command(name = "self")]
@@ -619,7 +628,16 @@ fn run_cli(cli: Cli) -> i32 {
         Commands::Lsp { stdio, parent_pid } => cmd_lsp(stdio, parent_pid),
         Commands::Trace { file } => cmd_trace(&file, diagnostic_format),
         Commands::Emit { file } => cmd_emit(&file, diagnostic_format),
-        Commands::Test { dir } => cmd_test(&dir, diagnostic_format),
+        Commands::Test {
+            dir,
+            locked,
+            offline,
+            frozen,
+        } => cmd_test(
+            &dir,
+            lock_mode_from_flags(locked, offline, frozen),
+            diagnostic_format,
+        ),
         Commands::SelfCommand(args) => cmd_self(&args, diagnostic_format),
     }
 }
--- a/crates/sifr/src/mode_resolution_tests.rs
+++ b/crates/sifr/src/mode_resolution_tests.rs
@@ -14,6 +14,25 @@ use std::sync::{Mutex, MutexGuard, OnceLock};
 
 static CWD_TEST_LOCK: OnceLock<Mutex<()>> = OnceLock::new();
 
+#[test]
+fn test_command_preserves_frozen_cargo_resolution_mode() {
+    let cli = Cli::try_parse_from(["sifr", "test", "tests", "--frozen"])
+        .expect("test command should parse");
+    let Some(Commands::Test {
+        dir,
+        locked,
+        offline,
+        frozen,
+    }) = cli.command
+    else {
+        panic!("expected test command");
+    };
+    assert_eq!(dir, PathBuf::from("tests"));
+    assert!(!locked);
+    assert!(!offline);
+    assert!(frozen);
+}
+
 struct CurrentDirGuard {
     previous: PathBuf,
     _lock: MutexGuard<'static, ()>,
--- a/crates/sifr_codegen/src/lib_test_project_codegen.rs
+++ b/crates/sifr_codegen/src/lib_test_project_codegen.rs
@@ -21,6 +21,7 @@ pub struct TestProjectCodegenResult {
     pub project_union_prelude: String,
     pub used_stdlib_modules: HashSet<String>,
     pub required_features: HashSet<StdlibFeature>,
+    pub interop: crate::InteropBuildPlan,
 }
 
 /// Generate support modules and root-level test bodies under one project union policy.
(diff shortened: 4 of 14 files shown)


yaseralnajjar:
## BLOCKERS

The refactor itself is sound and most of the M2 contract is met (evidence at end). Three regressions introduced by this diff:

### 1. `sifr test` silently swallows any test output line that parses as JSON
`crates/sifr_driver/src/test_runner/execution.rs:95,112` — `GeneratedCargoCommand::Test` now adds `--message-format=json-render-diagnostics` (needed for native-link evidence), and `user_facing_cargo_stdout` compensates by dropping **every** stdout line that `serde_json::from_str::<Value>` accepts. Cargo forwards the test binary's stdout verbatim, so a test that prints `42`, `true`, `null`, `"done"`, or a JSON payload (plausible given `sifr.json` is a stdlib module) has that line deleted from the user-visible output. Before this diff stdout was passed through unmodified. The filter needs to key on cargo's message shape (a JSON *object* carrying `"reason"`), not on "is parseable JSON". No test covers the filter — the runner tests assert only `success`.

### 2. Every `sifr test` is now a full cold Cargo build
`crates/sifr_driver/src/test_runner/execution.rs:52,84` — `target` was removed from `required_paths`, `cargo test` runs in a `TestInvocationWorkspace` under `temp_dir()` that `Drop` deletes, `copy_cached_project` copies only `Cargo.toml` and `src/`, and `run_generated_cargo_command` does `env_remove("CARGO_TARGET_DIR")` with no `build.target-dir` override. So no build output survives an invocation: a warm run recompiles `sifr_runtime`/`sifr_stdlib` and all vendored deps from scratch every time. Previously `cargo test` ran inside the committed cache entry and reused `target/`. The milestone requires the reusable entry stay immutable, not that build output be discarded — a per-cache-key target directory outside the immutable entry (or copying `target` into the invocation workspace) satisfies both. Related: `cache_report.status_line()` still reports warm/cold, which now describes only source materialization and reads as "reused build" when nothing was reused.

### 3. Native-link validation is now enabled for Python-runtime builds that previously skipped it
`crates/sifr_driver/src/build/generated_cargo_project.rs:156` gates on `should_validate_native_link_evidence(interop) || !additional_trusted_native_links.is_empty()`, and `materialize.rs:136` populates `additional_trusted_native_links` from `python_runtime`. `python_runtime` is set independently of Rust interop (`project_codegen.rs:242`), whereas the old `should_validate_native_link_evidence(&generated_project)` looked only at the interop plan. A Python-interop project with an empty Rust interop plan now has every build-script `linked_libs` entry checked against `{python links} ∪ {sysroot links}` and can fail with `SIFR-RUST-TRUST-0001` where it previously built. This is a binary-build policy change outside M2's scope; the gate should stay `should_validate_native_link_evidence(interop)` with the extra names only widening the trusted set.

## What is satisfied

- One authority: `generated_cargo_project.rs` owns manifest generation, bridge sources + `pub mod __sifr_bridge;` injection (now correctly inserted after leading inner attributes, which `#![cfg(test)]` lib roots require), support/namespace module layout with the `__sifr_support_main.rs` alias, and the single `run_generated_cargo_command` covering lock flags, `sysroot_cargo_config_args`, `prepare_cargo_resolution`/`assert_unchanged`, hermetic env, `record_cargo_invocation`, and native-link evidence. Both callers go through it; the old duplicate paths in `materialize.rs` and `execution.rs` are gone.
- Test resolution: `CargoResolutionPolicy::for_test_scope` mirrors `package_cargo_resolution_policy` (nearest-ancestor lock + sysroot lock + vendor dir) and is folded into the test cache key via `cargo_resolution_cache_key_fragment`, so constrained and normal test builds cannot share an entry.
- Cache immutability of the *entry* holds: the lockfile is prepared in the invocation copy, and the tests assert no `target/` and no `Cargo.lock` in the cache root across cold, warm, and frozen runs, plus a `test-build` trace entry carrying `--frozen`.
- CLI: `--locked/--offline/--frozen` are plumbed through `cmd_test` via the shared `lock_mode_from_flags`, with a parse test asserting `--frozen` survives.
- `generated_cargo_project.rs` is 433 lines, under the 900-line guardrail.

No changes made.

yaseralnajjar:
## SATISFIED

Reviewed the complete `3f618e3b..16024325` diff (13 files, +962/−480) statically; no gates re-run (you reported them green, and I made no changes).

**Blocker 1 — user stdout visibility: resolved.**
`user_facing_cargo_stdout` (`test_runner/execution.rs:117`) drops a line only when it parses as JSON **and** is an object containing `reason`. Scalars (`42`, `true`, `null`, `"done"`), reason-less objects, and plain text pass through; `cargo_stdout_filter_preserves_user_json_values` pins exactly that. Cargo's message objects all carry `reason`, and `json-render-diagnostics` still renders compile errors to stderr, which is written separately.

**Blocker 2 — cache immutability + warm artifacts: resolved.**
Cache `required_paths` is now `Cargo.toml` + `src/lib.rs` only; on miss the project is materialized in the staging root and committed. Execution copies the entry into a per-invocation temp workspace (`TestInvocationWorkspace`, dropped afterwards) and runs Cargo with `CARGO_TARGET_DIR = <entry>.target` (`test_target_directory`). Cache dir names are 16 hex chars (`workspace.rs:196`), so `with_extension("target")` is always a sibling `<hash>.target` with no collision or name aliasing, and `prepare_cached_artifact` only ever resolves the exact key path, so the sibling is never mistaken for an entry. `Cargo.lock` and `target/` therefore land in the temp copy / external dir, which the test asserts for cold, warm, and frozen runs.

**Blocker 3 — Python-only builds: resolved.**
`run_generated_cargo_command` gates on `should_validate_native_link_evidence(interop)` alone (`generated_cargo_project.rs:344`); `execution.additional_trusted_native_links` is only unioned *inside* that branch, so Python link names can only widen trust. `materialize.rs:136` supplies them from `python_runtime`, and the moved unit test still covers both the gate and the widening.

**Contract coverage:** single authority (`materialize_generated_cargo_project`) used by build and `sifr test`; shared manifest (`generate_dependency_cargo_toml_with_interop`, identical args on both paths so the cache-key manifest matches the written one), bridge sources, support/namespace layout with `main` → `__sifr_support_main.rs` remapping (including `main.child` → `__sifr_support_main/child.rs`), resolution + `assert_unchanged`, `lock_mode.cargo_arg()`, hermetic env, `record_cargo_invocation("test-build")`, native-link checks. `CargoResolutionPolicy::for_test_scope` mirrors `package_cargo_resolution_policy`, and `sysroot_root/Cargo.lock` matches `SysrootPaths::cargo_lock` (`layout.rs:79`), so `--frozen` has a real authority. Cache key now includes `cargo_resolution_cache_key_fragment`. CLI reuses the existing `lock_mode_from_flags`. Test interop is now real (`TestProjectCodegenResult.interop`) instead of `default()`; incidentally `declare_bridge_module` inserts `pub mod __sifr_bridge;` *after* leading `#![…]`, which is what makes bridges viable under `#![cfg(test)]` lib.rs.

### NEW_MECHANISM (record as a later item, not a blocker)
Test execution copies the cached project into a **fresh** temp dir every invocation (`create_invocation_workspace` uses pid+nanos). Cargo keys unit fingerprints and `-C metadata` by package path, so on a warm run the `sifr_tests` crate itself always recompiles and each invocation deposits a new set of artifacts/fingerprint dirs into the persistent `<hash>.target`, which grows without bound. Dependencies (the dominant cost) do stay fresh, which is why the test's `shared_target.is_dir()` assertion passes, but "compiled artifacts survive warm invocations" holds only for deps. Related: when an entry is invalidated, `final_root` is removed while its `<hash>.target` sibling is orphaned. A stable per-cache-key execution dir (e.g. `<hash>.exec` beside `<hash>.target`) would give full warm reuse and bounded growth while keeping the source entry immutable.

Non-blocking notes: `nearest_ancestor_file` is now duplicated in `cargo_resolution.rs` and `entrypoint_resolution.rs`; `execution.rs` hardcodes `"__sifr_support_main.rs"` instead of reusing `artifacts::SUPPORT_MAIN_RUST_FILE`; and `sifr test` combining `@rust` interop with `import sifr.python` would see an untrusted `python3.x` link, which mirrors the pre-existing gap for non-package builds (no `python_runtime` metadata) rather than being introduced here.

PR #3555, opened 2026-08-27
M3: enforce verification gate integrity

yaseralnajjar:
## Summary
- enforce absolute subprocess deadlines with process-group cleanup
- make every validation profile step resolve to a blocking budget
- expand compiler/tooling filesystem and split-brain authority scans
- add seeded negative self-tests for each mechanism

## Validation
- verification runner and hardening self-tests
- performance frontend-syntax-guardrails adapter: 4 passed
- developer-tooling typescript-go-transfer adapter: 2 passed

Stacked on M2 draft PR #3554. No compiler source changed, so Sifr create-PR and merge gates do not apply.


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `verification/areas/performance/check_split_brain_guardrail.py:38` only matched fully-qualified `sifr_syntax::parse_module(` calls and missed imported/bare calls, including the formatter's production consumer. Detect imported forms and classify the formatter exactly or route it through frontend.
- in-scope omission: `verification/runner/sifr_verify/profile_commands.py:109` did not create the descendant marker parent, so a clean target could make the seed fail independently and falsely pass cleanup. Create the parent or use a temporary directory.

Follow-up findings:
- pre-existing issue: two other hardening command runners still use direct-child-only timeout cleanup.
- infrastructure issue: detached subprocess groups need terminal signal propagation.
- suggestion: make the profile schema positivity rule self-describing if the local schema validator gains numeric-bound support.
- suggestion: use a local matched flag in the split-brain scan.
- suggestion: expose syntax-only classification rationales in guard output.
- pre-existing issue: production Rust `bin` paths remain excluded from direct-read inventory.

yaseralnajjar:
Verdict: NOT SATISFIED

Prior blocker verification:
- Imported/bare `parse_module` use is detected for explicit imports, the formatter is classified exactly, the imported form is seeded, and the local matched flag is used.
- The deadline self-test creates its marker parent before spawning the descendant.

Blocking findings:
- in-scope omission: the direct-read root set omits compiler-side `sifr_stdlib_manifest` and `sifr_sysroot`, leaving their stdlib/sysroot manifest, digest, directory, and path-probe reads outside the inventory. Add both roots, inventory the resulting production sites, and pin the roots in the self-test.

Follow-up findings:
- suggestion (new mechanism-level defect): wildcard or aliased `sifr_syntax` imports can still evade the split-brain import rule; no current production site uses them.
- infrastructure issue: `SIFR_VERIFY_DISABLE_STEP_BUDGETS` also disables the absolute deadline.
- infrastructure issue: new cold-cache blocking budget headroom is not yet measured.
- pre-existing issue: the recorded remaining hardening runners, signal propagation, and production-bin inventory carve-out remain open.

PR #3556, opened 2026-08-27
docs: make architecture inventory authoritative

yaseralnajjar:
## Summary
- replace stale architecture topology and API claims with current implemented boundaries
- generate the workspace crate/dependency map and validation-profile map from committed authorities
- add mutation-tested documentation guardrails for crate, profile, link, and machine-path drift

## Validation
- `python3 verification/areas/documentation/check_architecture.py`
- `python3 verification/areas/documentation/check_architecture.py --self-test`
- full documentation verification area (3 suites)
- verification runner self-test
- Ruff check; new checker Ruff format check
- `git diff --check`

Stacked on #3555. Human integration is intentionally deferred while the phase is staged sequentially.


yaseralnajjar:
Exact-SHA Opus review for `058780a5b505500fd04a80eca71aa467dfe037d8`: **NOT SATISFIED**.

Blockers: generated dependency map included two dev-only edges; crate/path guard missed bare phantom crate names, `crates/sifr`, and nonexistent in-crate paths; one stale CI paragraph still claimed `cargo test` was the full gate and omitted workspace Clippy scope. All were included in the single remediation batch.

yaseralnajjar:
Final allowed remediation review for `0cb9720cb80e66bc2be3c73e78206106cd998bd1`: **NOT SATISFIED**, with every original blocker verified fixed.

New mechanism defect: five legacy **Implementation responsibilities** blocks still mix shipped work and unlabeled future assignments into the current-architecture document. Per the phase rule, this is recorded under M12 and no third M4 review will run.

Post-remediation validation passed: architecture guard and all mutations, full documentation area, profile/runner self-tests, Ruff, JSON, file-size, HIR, and whitespace checks. No `crates/**` files changed, so compiler gates are not applicable.

PR #3557, opened 2026-08-27
M5: enforce structural generated Rust safety

yaseralnajjar:
## Summary
- replace unrestricted codegen verbatim nodes with opaque origin-bearing compiler fragments
- structurally validate fragments and complete generated Rust, including import/capability analysis
- reject invalid or forbidden Rust at the final generated-project materialization boundary
- remove generated option unwraps and panic-prone large-integer constant construction exposed by the new checks

## Validation
- cargo test -p sifr_codegen (1,145 passed)
- cargo test -p sifr_driver generated_cargo_project
- generated_code_quality:smoke (5 variants, 0 failures)
- cargo fmt --check
- raw-code, HIR maintainability, and file-size guardrails

Stacked on #3556. Integration remains deferred under the phase record.


yaseralnajjar:
Exact-SHA Opus review for `b5060ba1984f4ba9b9bf8964c95f86f6826854c5`: **NOT SATISFIED**. Blocking mechanism: the final forbidden-call detector treated legitimate user functions/methods named `unwrap` or `expect` as compiler-owned, allowing valid source to trigger the codegen assertion before the driver's structured materialization diagnostic. One consolidated remediation is in progress. The review also identified nested lint-suppression attributes, structured call/macro IR rejection, attribute-fragment import parsing, and exact-integer limit preservation as non-blocking gaps being folded into that batch. Full SHA-keyed evidence is preserved outside the Git tree.

yaseralnajjar:
Final allowed remediation review for exact SHA `74fb63bcf00a3d46b7ec03878065ca9bb10a0426`: **NOT SATISFIED**. It confirmed the provenance mechanism and all other M5 criteria, but reproduced the same original false-positive in the recursive argument-lowering path (for example `values.append(wrapper.unwrap())`). Per the no-third-review rule, that residual original site is being corrected with the exact regression and a full HIR-name-flow audit; there will be no additional Opus round. SHA-keyed review evidence is preserved outside Git.

yaseralnajjar:
M5 final implementation candidate: `5644505a6badeb51b39e38df82b3f8972c545265`. The final-review residual original name-flow defect was corrected without a third review, as required; the direct recursive-argument reproduction and all 1,149 codegen tests pass. The sole create-PR run exposed a stale `DEFAULT_MAX_INTEGER_DIGITS` retained-root entry, which was removed and whose owning guard passes. The sole merge run on the final SHA passed M5 guardrails, runner foundations, and 10/10 Rust interop variants, then stopped on the already-deferred M4 current/future architecture taxonomy defect; that is recorded for M12. No gate was rerun.

PR #3562, opened 2026-08-27
M10: harden cache identity and lifecycle

yaseralnajjar:
## Summary

- replace correctness-critical 64-bit cache identities with SHA-256 and full-key verification
- derive the frontend compiler identity from source/build revision
- add bounded cache status and clean commands with safe dry-run and symlink handling
- split near-limit modules and document the cache contracts

## Validation

- broad non-E2E-pass tests for sifr, driver, frontend, package, and LSP
- formatting, file-size, maintainability, frontend-cache, package-manager, and TypeScript-Go transfer guards

Human integration is intentionally deferred while the phase stack is completed.


yaseralnajjar:
**Verdict: NOT SATISFIED**

**Blocking findings:**

- in-scope omission: `crates/sifr_compiler_services/src/python_runtime.rs:106-108` and `:129-131` (with the paired omission at `crates/sifr_driver/src/build/project_codegen.rs:226-240`), criterion "close the deferred M1 follow-up: remove silent empty serialization fallback from Python certification cache identities". Neither file is touched by the candidate. `set_arrow_certifications`/`set_dlpack_certifications` still do `serde_json::to_string(&certifications).unwrap_or_default()`, and `apply_package_runtime_metadata` still guards each fragment with `if !metadata.arrow_certification_identity().is_empty()`, so a serialization failure silently drops the certification from the artifact cache key and a build with different certifications reuses the cached binary. Correction: propagate the serialization error as a diagnostic (or make the identity construction infallible by digesting the typed fields directly) and push the certification fragment unconditionally.

- in-scope omission: `crates/sifr_driver/src/test_runner/execution.rs:83-85`, criterion "replace fresh per-invocation test execution roots with stable per-cache-key roots". `execute_test_runner_project` still calls `TestInvocationWorkspace::create()` (which is `create_invocation_workspace("test_runner")`, `crates/sifr_driver/src/build/workspace.rs:8-43`: `sifr_test_runner_<pid>_<nanos>`) and copies the cached project into it on every run; the root is deleted on `Drop`. Correction: run `cargo test` from a root derived from the test-runner cache key (the cache entry workspace itself or a stable `<key>` sibling) so warm reuse does not depend on a per-invocation copy.

- in-scope omission: `crates/sifr_driver/src/build/workspace.rs:269`, criterion "reclaim their external Cargo target when the source-cache key is invalidated". The test target dir is the cache-key sibling `<key>.target` (`crates/sifr_driver/src/test_runner/execution.rs:134-136`), but the invalidation path only does `remove_dir_all(&final_root)`; the same holds for the pending-artifact drop and concurrent-winner paths. Every key rotation orphans a full Cargo target directory that nothing links back to the dead key, and `sifr cache clean` treats `<key>` and `<key>.target` as unrelated entries so it can drop the key and keep the orphan. Correction: reclaim `final_root.with_extension("target")` whenever `final_root` is removed on invalidation.

**Follow-up findings:**

- suggestion: `docs/cli/cache.mdx` is the only `.mdx` under `docs/` not registered in `docs/docs.json` navigation (all other pages, including every `docs/cli/*.mdx`, are listed), so the new page is absent from the sidebar even though `docs/cli/overview.mdx` advertises the commands.
- suggestion: `crates/sifr_driver/src/build/rust_interop_probe_cache.rs:125-136` adds authoritative-lock and trusted-vendor paths to the probe key using raw `display()`, while the sibling fix in `prepared_lock_path` switched to `normalized_policy_path`. This re-introduces the same "safe but unstable miss when a path becomes canonical" class the item is closing, in a newly added key input.
- suggestion: the same `lower_hex`/`sha256_hex` loop is now duplicated in six places (`sifr_frontend/build.rs`, `cache_fingerprint.rs`, `python_input_fingerprint.rs`, `graph/digest.rs`, `rust_interop_digest.rs`, `workspace.rs`, `package_discovery.rs`, `formatter_cache.rs`, plus the e2e harness). A single shared helper would keep the encoding provably identical.
- suggestion: `COMPILER_SOURCE_ROOTS` in `crates/sifr_frontend/build.rs:5-13` excludes `crates/sifr_format`, yet `crates/sifr/src/formatter_cache.rs:498` uses `CompilerFingerprint::current()` as the formatter cache's compiler identity. A formatter-only change therefore does not invalidate formatter cache markers. (Strictly better than the pre-change key, which had no fingerprint at all — hence not blocking.)
- pre-existing issue: `crates/sifr_source/src/lib.rs:287-292` still computes a 64-bit FNV `stable_hash`, consumed by `SourceMap::source_hash` (`crates/sifr_diagnostics/src/source_map/mod.rs:97,133`). It is in-memory only, not a persisted cache identity, so it is outside this item's "correctness-critical persisted identity" scope.
- infrastructure issue: `sifr cache clean` has no locking against a concurrent build using the same `$TMPDIR/sifr_generated_artifact_cache` entries; a clean during an in-flight build can remove a workspace or shared probe target underneath it.

yaseralnajjar:
I reviewed the remediation diff (`ed6c7c2e6..2e2ad86fb`) against the prior findings and the item's criteria.

Verified addressed:
- **Finding 1** — `python_runtime.rs:102-126` now returns `Result<String, serde_json::Error>`; `project_codegen.rs:226-240` always emits both fragments and maps failure to a `BUILD_MATERIALIZATION_FAILURE` diagnostic. No remaining callers of the old `&str` accessors.
- **Finding 2** — `execution.rs:167-188` publishes a stable `<key>.execution` sibling by atomic rename from a staging dir, reuses it when complete, and `test_runner.rs:192-236` asserts warm reuse of both `.execution` and `.target`.
- **Finding 3** — `workspace.rs:270-311` invalidation now removes `.execution`, `.target`, and the source root, and returns a diagnostic on failure; lifecycle grouping added in `cache_lifecycle.rs`.
- Policy-path normalization (`rust_interop_probe_cache.rs:138,146` via the now-`pub(super)` `normalized_policy_path`), `sifr_format` in `build.rs` compiler source roots, and `docs/docs.json` nav registration are all present.

One regression in the newly added lifecycle grouping blocks.

Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_driver/src/build/cache_lifecycle.rs:330-338`, criterion "Add bounded cache status/clean lifecycle commands" (and the grouping half of prior finding 3). `group_cache_candidates` adds companions with `path.with_extension(extension)` without excluding the path itself. For a `test_runner` candidate that *is* a companion whose owner stem is absent from the candidate set (an orphaned `<key>.target` or `<key>.execution` — exactly the leftovers the pre-remediation invalidation produced and that `cache clean` exists to reclaim), the `continue` guard at line 326-328 does not fire, so the loop pushes the path a second time: `paths = [X.target, X.target]`. Consequences: (a) `measure_entry` walks the same tree twice, so `sifr cache status` reports roughly double the bytes for those entries; (b) `clean_cache_at` line 184-188 calls `remove_path` twice on the same path, and `remove_path` (line 352-353) propagates the second call's `NotFound` from `symlink_metadata`, so `sifr cache clean` fails with an io error instead of reclaiming. When both orphans exist without the owner, two overlapping entries are also emitted, double-counting and double-removing across entries. The initially reviewed SHA treated each candidate as its own single-path entry and cleaned orphans without error, so this is a behavior regression introduced by the remediation, on the migration path from pre-remediation caches. The new test at line 433-462 only covers the owner-present case and cannot catch it. Correction: group `test_runner` candidates by their stem (`path.with_extension("")`) and emit one deduplicated path group per stem, so a companion is never paired with itself and orphaned companions of a missing owner form exactly one entry; add coverage for orphaned `<key>.target` / `<key>.execution` with no owner directory.

Follow-up findings:
- pre-existing issue: `crates/sifr_driver/src/build/rust_interop_probe_cache.rs:117` still feeds `backend.cargo_manifest_path.display().to_string()` into the probe cache key un-normalized, while sibling policy paths on lines 138/146 now use `normalized_policy_path`. It causes only spurious misses (different spellings of the same existing manifest), not false hits, and is outside the stated normalization criterion.
- suggestion: with a stable per-key `.execution` root, two concurrent `sifr test` invocations sharing one cache key now run Cargo in the same project directory rather than in isolated invocation roots. Cargo's build-directory lock serializes the compile, but lockfile resolution is not covered by it; worth confirming same-key concurrency explicitly in a later item (cross-key parallel isolation is already covered).

yaseralnajjar:
Gate evidence for final implementation candidate `2e2ad86fb80ee916542738967935039c157fa18e`:

- The single create-PR gate ran once.
- The single merge gate ran once.
- Both passed generated-demo freshness, dependency/ownership, method dispatch, unsafe ABI, codegen invariant, sysroot/stdlib, driver, runner-foundation, and all ten Rust-interop variants.
- Both stopped at coverage on the unclassified new `sifr_frontend` custom build-script target and the two existing M11 delivery-taxonomy lines in `internal_docs/architecture.md` (975 and 1459).
- Neither gate was rerun.

The M10 staged handoff and deferred follow-ups are recorded by record-only commit `a46fd50c4d09b14cf14ed012c3106e9da13def73`.

PR #3563, opened 2026-08-28
test: add sustained compiler fuzzing

yaseralnajjar:
## Scope

- add six coverage-guided cargo-fuzz targets for parser, lowering, ownership, codegen validation, diagnostics, and project graphs
- add semantic properties for normalization, narrowing, incremental/full equivalence, and deterministic codegen
- rename deterministic fuzz smoke to mutation smoke
- wire non-blocking sustained fuzzing into nightly/release profiles and CI evidence

## Validation

- targeted semantic-property tests passed
- fuzz/property area passed: 43 variants
- all six fuzz targets built offline and passed the 45-second nightly sustained lane
- workspace and fuzz-project strict Clippy passed
- coverage readiness, architecture, runner self-test, profile plans, formatting, and size/maintainability guards passed

## External existing failure

- pre-v1 regression closure guard reports the unchanged performance trend manifest SHA as stale; benchmark and trend inputs are not changed by this item

Stacked on the M10 branch. Human review/merge checkpoint is deferred by request.


yaseralnajjar:
## Review of M11 (base `a46fd50c4` → candidate `b62fe3afa`)

**Verdict: NOT SATISFIED**

**Blocking findings:**

- in-scope omission: `verification/runner/sifr_verify/hardening/coverage_fuzz.py:115` (with `.github/workflows/local-first-validation.yml:92`), criterion "Wire a real sustained lane into nightly/release policy". The per-target subprocess timeout is `seconds + 120`, and the timed subprocess is `cargo +nightly fuzz run`, which performs the sanitizer-instrumented build of the whole fuzz project inside that same window. The `sustained-fuzz` CI job runs only `cargo install cargo-fuzz` and `cargo +nightly fetch` before invoking the runner — there is no `cargo +nightly fuzz build` step and no target-dir cache — so on a cold runner the first target must compile 178 packages (`verification/fuzz/Cargo.lock`) with coverage+ASan instrumentation within 165 s (nightly) or 240 s (release). That build does not finish in that budget, so `subprocess.TimeoutExpired` fires, the target is recorded as `coverage-fuzz-failed`/exit 124, and because the suite is non-blocking (`verification/areas/fuzz_property/runner.py`, `blocking = name != "sustained-fuzz"`) the area exits 0 and the CI step is green having fuzzed nothing. The passing local evidence does not cover this: it ran after `cargo +nightly fuzz build` had already populated the fuzz target dir, i.e. warm only. Correction: build before timing — add a `cargo +nightly fuzz build --fuzz-dir verification/fuzz` step to the `sustained-fuzz` job (or perform the build as a separate untimed subprocess in `run_coverage_fuzz_target`) so the manifest budget bounds fuzzing time only; alternatively make the timeout margin an explicit manifest field sized for a cold instrumented build.

**Follow-up findings:**

- pre-existing issue: the unchanged `verification/areas/performance/data/trend/current.json` stores a stale `benchmark_manifest` SHA. Neither file is touched by this candidate; record it in its owning issue.
- infrastructure issue: tool unavailability and real fuzz findings are indistinguishable in the area result. A missing `cargo-fuzz` subcommand, an offline registry miss on the excluded fuzz lockfile, and a genuine crash all surface as a single non-blocking `coverage-fuzz-failed` mismatch (`coverage_fuzz.py:100-103`). A local `--profile nightly`/`--profile release` run on a host without `cargo-fuzz` will therefore report six silent non-blocking failures. A distinct `tool-unavailable` mismatch class (and a summary line naming it) would keep "the lane could not run" separable from "the lane found something".
- suggestion: `verification/fuzz/Cargo.toml` declares `serde_json` but no fuzz target or `src/lib.rs` uses it; drop the dependency.
- suggestion: `verification/areas/fuzz_property/manifest.json` still declares `"resource_classes": ["default-local"]` while `nightly.json`/`release.json` now select the area with `["default-local", "long-running"]`. Nothing cross-checks the two today (`profiles.validate_selected_area_suites` only validates suite names), so this is drift rather than a failure.
- suggestion: the new `cargo-test` property entries ignore `repeat_runs` entirely (`property_and_fuzz.py:130-137` returns after a single `run_cargo_property`), and the minimum was relaxed to 1 for that command. The four new semantic properties therefore get no rerun-determinism comparison, unlike every other `property` entry. Either honor `repeat_runs` for `cargo-test` or state in `verification/policy/fuzz_property.md` that the rerun contract is inapplicable there.
- suggestion: `semantic_property_codegen_is_deterministic` compares two `compile()` calls in one process, so it cannot observe per-process hash-seed or environment-dependent ordering. A second comparison against a separately spawned process would make the determinism claim stronger.
- suggestion: `verification/fuzz/fuzz_targets/project_graph.rs` creates and `remove_dir_all`s a fixture tree on every iteration, putting filesystem IO in the hot loop and cutting exec/s sharply; hoist fixture creation to a one-time init and mutate only the in-memory metadata JSON per input.
- suggestion: `valid_program` varies only 24 integer literals and `ownership_program` only three statement shapes (`verification/fuzz/src/lib.rs`), so libFuzzer's coverage feedback has very little structure to steer. An `Arbitrary`-driven grammar would make the `codegen_validation` and `ownership` targets meaningfully coverage-guided rather than template-parameterized.
- suggestion: `plans/phases/29_verification_hardening.md:54,68` still name the `fuzz-smoke` suite, and `crates/sifr/tests/e2e_support/harness_behavior_tests.rs:254` retains a "smoke fuzz/property" comment, after the accurate rename.

yaseralnajjar:
Verdict: SATISFIED

The original blocker is closed at both layers. `verification/runner/sifr_verify/hardening/coverage_fuzz.py:67-73` runs a single `build_fuzz_project` preflight before any target, with its own 1200 s infrastructure timeout, a dedicated `fuzz_project_build` case (`instrumented-build` variant), and an early return on failure — so the per-target `-max_total_time` and `seconds + 120` subprocess timeouts (lines 150, 163) now start only after instrumented compilation has completed. `.github/workflows/local-first-validation.yml:92-93` adds a blocking `CARGO_NET_OFFLINE=true cargo +nightly fuzz build --fuzz-dir verification/fuzz` step after `cargo install cargo-fuzz` and `cargo +nightly fetch`, so a cold runner that cannot build fails the job instead of silently burning the fuzz budget on compilation. Build and run use the same cargo-fuzz default profile, cwd, and env, so the preflight artifacts are the ones the timed runs execute; the failure case is counted in `total_variants`/`total_failures` while the suite stays non-blocking (`verification/areas/fuzz_property/runner.py:124`), which preserves the "broad findings non-blocking until minimized" criterion. Manifest, policy, lane doc, and profile contracts are unchanged and remain consistent with the six-target contract.

Blocking findings:
- none

Follow-up findings:
- suggestion: `verification/runner/sifr_verify/hardening/coverage_fuzz.py:100-113` captures build stdout/stderr but discards it; a failed preflight emits only `fuzz-project-build-failed` and an exit code in the machine-readable area result, so the non-blocking lane's most common failure mode is the least diagnosable. Consider recording a bounded tail of the build output in the variant.
- infrastructure issue: `verification/areas/fuzz_property/manifest.json` declares `timeout_seconds: 900` for the area, while the release lane can now reach 1200 s build plus 6 × 120 s fuzzing. No runner code enforces the area-level `timeout_seconds` today (it appears only in `selftest.py` fixtures), so this is declared-budget drift rather than a live failure, but the number no longer describes the lane. The release profile's blocking `default_step_budget` of 2 400 000 ms leaves roughly 8 minutes of headroom in the worst case.
- suggestion: `verification/areas/fuzz_property/sustained_lane.md` still describes CI as installing cargo-fuzz and fetching the lockfile "before the offline run" and does not mention the new blocking prebuild step or the runner preflight. The doc is not wrong, only incomplete after the remediation.
- pre-existing issue: `verification/runner/sifr_verify/release_evidence.py:292-301` raises `GovernanceError` on duplicate variant labels, and all six sustained targets share the label `coverage-guided`. This predates M11's remediation and only matters if the sustained suite is ever routed through release evidence collection; the new `instrumented-build` label is unique and does not worsen it.

yaseralnajjar:
M11 gate summary for implementation candidate 7664b902bf4697ca20bc39c683cdb617d26032e2:

- Final remediation Opus review: SATISFIED.
- Create-PR gate: ran once and failed in diagnostics:rules after generated-demo freshness, compiler guardrails, runner foundations, Rust interop, and coverage readiness passed.
- M11-owned failure: verification/fuzz/fuzz_targets/diagnostics.rs uses the forbidden SIFR-TYPE-0001 catch-all.
- Pre-existing diagnostics failure: active SIFR-INTERNAL-0003 is absent from the code catalog and baseline-coverage registry.
- The create-PR gate was not rerun. The merge gate did not run after the create-PR failure.
- Both allowed review rounds are exhausted. Follow-ups and the exact handoff are recorded in the phase ledger.

Full create-PR output is preserved outside the reviewed tree under the candidate SHA.

PR #3578, opened 2026-08-29
test(codegen): make emitted Rust gates fail closed

yaseralnajjar:
## Summary
- discover and lock every maintained generated-Rust entrypoint/source family
- replace formatter mutation and broad Clippy suppression with exact owner-bound debt signatures
- add lexical safety scanning, negative seeds, recursive freshness, and governed semantic audit anchors
- remove 15 non-authoritative legacy Rust snapshots and 3 orphan emitted files

## Validation
- generated-code quality full profile: 9 variants, 0 failures (91 generated projects)
- audit inventory and mutation self-tests
- Python/JSON syntax checks
- file-size, HIR, driver maintainability, docs-link, and diff hygiene checks

No compiler source changed, so phase rules omit Sifr create-PR and merge gates for this item.


yaseralnajjar:
# Item 1 Remediation Review (Round 2, final)

**Range reviewed:** `d7a08f12a` → `b0fb5c204` (full base-to-candidate diff, 44 files). Draft PR #3578.

---

## 1. Verdict

**SATISFIED**

---

## 2. Original blocker resolution

| # | Blocker | Status | Evidence |
|---|---|---|---|
| 1 | Freshness gate conflated emit failure with stale output | **Resolved** | `verification/areas/generated_code_quality/inventory_gates.py:40-56` — `classify_freshness_result` accepts only two shapes: `rc==0 and lines == [FRESHNESS_SENTINEL]`, or `rc==1 and stale and len(stale)==len(lines)`; everything else raises. This is exactly matched by the producer: `scripts/check_demo_emitted_freshness.py:63-66` returns the compiler's non-zero code with **empty stdout** on emit failure, which now lands in the `raise` branch. Self-test at `inventory_gates.py:58-81` covers success, stale, `(1,"")`, `(1,"compiler failed")`, `(0,"")`, `(0,"stale: …")`, `(2,sentinel)`. Verified passing by execution. |
| 2 | Semantic-anchor self-tests did not exercise all validation mechanisms | **Resolved** | `check_emitted_rust_audit_inventory.py:422-515` now mutates every branch of `validate_semantic_anchor`: metric extra-field and **missing baseline key** (426), text extra-field/revision/escaped-path/**uncovered-evidence path** (430-466), empty `contains`, unreadable baseline anchor, search_count extra-field, invalid field shapes, **uncovered search root** (498), **valid-shape count mismatch** (505), **unknown kind** (510). `python3 check_emitted_rust_audit_inventory.py --self-test` → PASS on this SHA. |
| 3 | ERQ-016 dropped the real finite `itertools.count` cap | **Resolved** | Inventory ERQ-016 is retitled "Infinite count iteration is capped at a finite prefix", mechanism names `count_from(start, step, 10000)`, anchored `{"kind":"text","revision":"current","path":"stdlib/sifr/itertools.sifr","contains":"return count_from(start, step, 10000)"}`. Confirmed live at `stdlib/sifr/itertools.sifr:276`. Owner remains Item 5. |
| 4 | Invalid-`islice` behavior anchored to a fixture comment | **Resolved** | Split into new ERQ-033, anchored to the implementation: `stdlib/sifr/itertools.sifr:256-258` (`if stride <= 0: / stride = 1 / actual_stop = actual_start`), confirmed present. Owner Item 5. Duplicate-ID, ID-format, and item-coverage validation all still pass with the new row. |

---

## 3. Blocking findings

**None.** I found no candidate-introduced mechanism defect and no in-scope acceptance omission.

Independently re-verified on the candidate worktree (read-only):
- audit-inventory validation + full self-test: PASS (33 findings, 29 actionable, 4 rejected)
- surface-inventory validation + self-test: PASS (9 entrypoint classes, 8 discoveries, path-set SHA-256 pinned → discovery cannot shrink silently)
- debt self-test + selection-id validation: PASS
- checked-in snapshot safety debt exact match: PASS over **262** `demos/**/emitted.rs`

Secondary claims spot-checked and confirmed:
- **Format-before-check removed** — `generated_code_quality.py:664-685` runs only `cargo fmt -- --check`; the mutating `cargo fmt` call (ERQ-005's anchor text) is gone.
- **Blanket allows removed** — the 68 `-A` `GENERATED_CLIPPY_ARGS` are deleted; `STRICT_CLIPPY_ARGS` (`quality_policy.py:14-43`) denies warnings and enables pedantic/nursery plus the safety lint set.
- **Hard source-literal checks re-enabled** — `SOURCE_FORBIDDEN_POLICY_IDS` covers unwrap/expect/panic/todo/unimplemented/unsafe/allow, and none of these policy ids appear anywhere in `generated_quality_debt.json`, so any occurrence fails the exact comparison. Item 8 is correctly absent from `safety.owner_items` (`[3, 4, 11]`).
- **Negative seeds** — 18 seeds pinned by digest; 12 safety policies, 3 Clippy lints (each asserted by *lint code*, not just non-zero exit), 1 rustfmt, 2 determinism.
- **Symlink-safe path resolution** — all four modules derive `REPO_ROOT` from `Path(__file__).resolve()` and compare via `resolve().is_relative_to(repo.resolve())`.
- **ERQ-020 flip to `confirmed`** is factually supported: `crates/sifr_codegen/src/methods/decimal.rs:23-34` emits `.with_prec(28).unwrap_or_else(|| <default ctx>)`.
- No `crates/**` file changed, so omitting the Sifr create-PR/merge gates matches Execution Rule 4.

---

## 4. Non-blocking follow-ups

1. **Three companions were deleted rather than regenerated.** `demos/graph_isolation/single_file/emitted.rs`, `demos/temp_workspace_isolation/parallel_runs/{a,b}/emitted.rs` each still have a live sibling `main.sifr` and were hand-written idealized stubs, not real emission. Removal is explicitly permitted by the acceptance criterion, but regenerating would have been equally cheap and kept three surfaces in the pinned 265-file set instead of 262. Suggest revisiting at Item 8/12.
2. **ERQ-025 is now a stale actionable row.** Its mechanism ("Fifteen demo `main.rs` files coexist…") was discharged by this very candidate (15 `main.rs` deleted), yet it remains `confirmed`/`owner_item: 8`. Its evidence resolves only through baseline history. Item 8 should close it as already-satisfied rather than re-derive work.
3. **Gate breadth vs. discovery breadth.** `surface_inventory.json` inventories all 9 entrypoint classes (1,479 discovered paths), but corpus/rustfmt/Clippy/panic-scan/determinism still execute only over the 91 `corpus_manifest.json` positive entries. Discovery "covers" the classes as the criterion words it, but they are not all *exercised*. Worth an explicit statement of intent in the phase record before Item 12.
4. **Checked-in `demos/**/emitted.rs` get freshness + safety scanning but no rustfmt/Clippy check.** ERQ-002's residual formatting debt lives in `rustfmt.owner_items: [8]` while ERQ-002 itself is owned by Item 1 — the ownership split is defensible (gate vs. producer) but is not written down anywhere; record it.
5. **Minor code nits (no reachable failure):** `quality_policy.py:161` has a silent `else path.name` fallback in `scan_files` that would collide signatures if ever reached; `SMOKE_ENTRY_IDS` is duplicated in `runner.py:21` and `inventory_gates.py:37`; `safety.owner_items` includes Item 3, so a future `unwrap` debt row would pass `validate_debt_owners` (it would still fail the exact comparison, which is the real gate).

---

## 5. Closure recommendation

Close Item 1 and merge. All four round-1 blockers are fully and verifiably resolved at `b0fb5c204`, every Item 1 acceptance checkbox is backed by an executing, fail-closed mechanism that I re-ran independently, and no new in-scope mechanism defect was introduced. Under the no-third-review rule, carry follow-ups 1-4 into the phase record as owned notes for Items 8 and 12; item 5 is optional cleanup.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `verification/areas/generated_code_quality/inventory_gates.py:125` — criterion "checked-in generated output must be fresh or removed as non-authoritative" / gates must fail closed. The new freshness gate accepts `returncode in {0, 1}` and derives staleness only from `stale: ` stdout lines. `scripts/check_demo_emitted_freshness.py:64` returns the failing `sifr emit` exit code, and a user-diagnostic emit failure exits 1 with no `stale: ` line, so a demo whose emission breaks yields `stale=[]`, `orphans=[]`, `actual=None`, and the gate reports "freshness passed". Correction: distinguish the two exits — require the script's success sentinel (`all generated demo companions are fresh`) when no `stale: ` lines are present, or have the script exit with a distinct code for emit failure and raise on it.
- in-scope omission: `verification/areas/generated_code_quality/check_emitted_rust_audit_inventory.py:440-448` — criterion "the audit-inventory self-test covers every validation branch". The `stale_search_count` mutation uses `patterns: ["**/*.py"]`, which is rejected by the field-shape check, so it asserts `search_count fields are invalid` and never reaches the `search count mismatch` branch at line 316 (verified: the same mutation returns only the field-shape error). The metric-key branch, the `semantic_anchor.path must be covered by finding evidence` branch, the evidence-covered-search-root branch, and the unknown-`kind` branch also have no mutation case, and the semantic-anchor validator is the largest new validation surface added by this item. Correction: add mutations that hit each branch, including a valid-shape `search_count` with a wrong `expected_count`.
- regression: `verification/areas/generated_code_quality/emitted_rust_audit_inventory.json:199-201` (ERQ-016) — the Item 0 locked ledger recorded "count-style generation has a finite implementation cap and islice-style paths do not uniformly reject invalid negative or zero-step arguments"; the candidate rewrites the mechanism to the islice clause only and drops the finite-cap claim and its evidence. The cap is still present and reproducible (`stdlib/sifr/itertools.sifr:273-276`, `count()` returns `count_from(start, step, 10000)`), and it violates the locked contract clause "No finite cap can stand in for an infinite iterator", so Item 5 could close against a mechanism narrower than the recorded defect. The change is also unrecorded: `plans/issues/active/ad-hoc-emitted-rust-excellence.md` documents only the ERQ-020 correction. Correction: restore the finite-cap clause (or split it into its own Item 5 finding) anchored at the `count_from(start, step, 10000)` site, and record the amendment in the issue's audit-correction note.
- in-scope omission: `verification/areas/generated_code_quality/emitted_rust_audit_inventory.json:201` — criterion "evidence rows use governed semantic anchors, not path existence alone". ERQ-016's anchor is the comment text `explicit adaptation for non-positive step => empty iterator` in `crates/sifr/tests/e2e/pass/cpython_itertools.sifr:66`, a hand-written comment in a passing fixture. The comment can survive verbatim after the mechanism is fixed, and can be deleted while the defect remains, so the anchor does not track the mechanism. The implementation site exists (`stdlib/sifr/itertools.sifr:256-258`, `if stride <= 0: stride = 1; actual_stop = actual_start`). Correction: anchor to the implementation text, not the fixture comment.

Follow-up findings:
- suggestion: `verification/areas/generated_code_quality/generated_code_quality.py` — `SOURCE_FORBIDDEN_POLICY_IDS = {"allow-attribute"}` narrows the intrinsic emitter-source literal lint from the previous seven-pattern set to one. I re-ran the old pattern set over `codegen_source_files()`: the only additional hit is a `direct-index` false positive on `"\n[dependencies]\n"` in `lib_project_codegen.rs:481`. So nothing is hidden today, but `unwrap`/`expect`/`panic`/`todo`/`unimplemented`/`unsafe` literals in emitter code are now caught only if a corpus program exercises that emitter. Re-enabling those six for the source lint (excluding `direct-index` and `signed-to-usize`) restores the coverage at no cost.
- pre-existing issue: `stdlib/sifr/itertools.sifr:242-258, 273-276` — the finite `count()` cap and the non-positive-step islice adaptation are real Item 5 producer defects; noted only because the ledger edit above touches how they are tracked, not as work for this item.
- infrastructure issue: `verification/areas/generated_code_quality/surface_inventory.py:73` — `path.relative_to(repo_root)` uses the unresolved repo root while the glob root is resolved; if the checkout path ever contains a symlinked component this raises an uncaught `ValueError`. Not reachable in this worktree.
- suggestion: `verification/areas/generated_code_quality/data/generated_quality_debt.json` — `safety.owner_items` lists Item 8, but no safety entry maps to an Item 8 policy (`allow-attribute`). Unused owner entries are permitted by `validate_debt_owners`, yet they weaken the "exact owner-bound debt" property; consider rejecting owner items with no referencing entry.

PR #3580, opened 2026-08-29
Implement exact integer and overflow architecture

yaseralnajjar:
Implements Item 2 of the emitted Rust excellence phase: one canonical exact integer representation, exact arithmetic and checked conversions, consistent typed division/modulo failures, and debug/release differential property evidence.\n\nValidation before review: workspace Clippy with warnings denied; rustfmt; file-size and HIR guardrails; 1,150 codegen tests; 1,044 lowering tests; 84 runtime tests; exact-integer oracle tests in debug and release; 143/143 emitted pass fixtures; CLI unit and build-output integration tests; expected-failure corpus.


yaseralnajjar:
Item 2 initial Opus review — exact candidate `d4aea519efebdf29bad472a9795afcdd72c4f865`, base `907fe8e3c2fbe64c5e6afb4ed5fda047b34dc68b`.

Verified against the exact candidate content (worktree clean at `d4aea519e`), by reading the diff and reproducing behavior with a freshly built `target/debug/sifr`.

Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_lowering/src/lower/expressions/core_and_calls.rs:618-622` (slice-step lowering) together with `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_stepped_slice.rs:107` and `crates/sifr_runtime/src/slice.rs:24`. Criterion: "Floor division, modulo, division-by-zero, overflow/resource-limit, exponentiation, and shift failures are consistent and typed" plus the project rule that the compiler must prevent user-triggered runtime panics. Reason: the candidate replaced the old hand-rolled stepped-slice loop with `SifrSliceIndices::new_known_nonzero`, which `assert!`s a nonzero step, but nothing in lowering proves it. `range(start, stop, step)` did get the proof gate (`bytes_len_range.rs:358` calls `is_proven_nonzero_integer_expr` and emits `SIFR-INT-0005`); the slice path did not. Reproduced:
  ```python
  def main() -> None:
      xs: list[int] = [1, 2, 3, 4, 5]
      n: int = len(xs) - 5
      ys: list[int] = xs[::n]
  ```
  `sifr check` reports "no errors found"; the release binary aborts with `panicked at crates/sifr_runtime/src/slice.rs:24: compiler nonzero slice-step proof was invalid`. At base `907fe8e3c` the same program produced an empty list, so this is a newly introduced user-triggerable panic (it also makes the `assert!` data-dependent rather than a programmer invariant). Correction: gate the step in `core_and_calls.rs:618` exactly as `bytes_len_range.rs:345-366` does — reject an unproven nonzero step with `DiagnosticCode::INT_EXACT_DIVISION_REQUIRES_HANDLING`, or give stepped slicing a typed fallible result and emit a checked constructor; the same emitter serves `list`, `bytes`, and `str`, so all three are affected.

Follow-up findings:
- pre-existing issue: `crates/sifr_lowering/src/lower/expression_operators/integer_semantics.rs:186-212` proves only that the *operands* of `int / int` are float-representable, not the quotient. `1 / 3` therefore type-checks as plain `float` and emits `SifrInt::from_i64(1).to_f64_proven_exact() / SifrInt::from_i64(3).to_f64_proven_exact()` → `0.3333…`, while the same value pair through non-const operands returns `Err(FloatPrecisionLossError)` from `checked_true_div`. The operand-only rule is unchanged from base, but this candidate is what made the unproven path fallible, so the two paths now disagree on identical inputs; `integer_model.md`'s "no silent precision loss" reading wants the quotient proved.
- pre-existing issue: `crates/sifr_lowering/src/lower/integer_failure_diagnostics.rs:11-21` — `exact_int_division_requires_handling` now unconditionally returns `false` with every parameter underscored. Its job moved to the Result-typing path; the dead function and its call sites should be removed.
- suggestion: `statically_safe_bounded_integer_arithmetic` (`integer_semantics.rs:82-121`) uses `literal_exact_integer_value`, which only sees literals and unary `+`/`-`, whereas `is_proven_nonzero_integer_expr`/`proven_exact_integer_value` also resolve const-evaluable module constants. `LIMIT ** 2` with a const `LIMIT` is forced into a `Result` that the doc's const-evaluation rules say should be provable. Purely conservative, not unsound.
- suggestion: `crates/sifr_runtime/src/int.rs:614-620` — the negative-zero branch in `exact_bigint_ratio_to_f64` is unreachable because `BigInt::ZERO` never carries `Sign::Minus`.

yaseralnajjar:
Item 2 sole remediation Opus review — exact candidate `9606b67b84ae5865105415399d647319b455bb99`, base `907fe8e3c2fbe64c5e6afb4ed5fda047b34dc68b`.

Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_lowering/src/lower/expression_operators/integer_semantics.rs:216` (`exact_integer_ratio_is_proven_float_representable`, used at line 203) — criterion "no user-triggered runtime panics / no silent narrowing" and "typed fallibility where exact operations can fail". The remediation replaced the base proof (`is_proven_nonzero_integer_expr(right)` **and** `exact_integer_expr_is_proven_float_representable(left)` **and** `(right)`) with a proof about the *reduced* quotient only. Codegen for an `int / int` expression whose result type is `Float` never uses an exact-ratio runtime path: `lower_integer_true_division` returns `None` for a non-`Result` result type (`crates/sifr_codegen/src/stmt_support_emitter/checked_integer_codegen.rs:419`), `lower_mixed_float_integer_arithmetic` bails when both operands are integers (same file, line 633), so lowering falls through to the per-operand conversion at `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_binop.rs:401-425` (and the equivalent path at `crates/sifr_codegen/src/lower_expr/scalar_operands.rs:27`), emitting `left.to_f64_proven_exact() / right.to_f64_proven_exact()`. `SifrInt::to_f64_proven_exact` panics when the *operand* is not exactly representable (`crates/sifr_runtime/src/int.rs:497-501`). The new proof no longer establishes that precondition, because gcd reduction can cancel a non-representable common factor.
  Failure scenario: 
  ```
  def main() -> None:
      numerator: int = 9007199254740993        # 2**53 + 1
      denominator: int = 9007199254740993
      value: float = numerator / denominator
  ```
  The reduced ratio is `1/1`, so the ratio proof succeeds and the expression is typed `float` (infallible). The emitted Rust calls `to_f64_proven_exact()` on `2**53 + 1`, which returns `None` and panics with `compiler exact-integer float proof was invalid`. A non-degenerate variant such as `numerator = 27021597764222979` (`3 * (2**53 + 1)`) with the same denominator reduces to `3/1` and panics identically. The base commit rejected both (each operand had to be individually representable), so this is a regression introduced by the remediation, and it is the same class of defect — a compiler "proof" assertion reachable from ordinary source — that blocked the first review.
  Correction: make the static proof match what codegen actually emits. Either (a) keep the ratio proof but additionally require `exact_integer_expr_is_proven_float_representable` for both operands before returning `Type::Float`, or (b) route the proven-safe `int / int → float` case through an exact-ratio runtime conversion (the existing `exact_bigint_ratio_to_f64` behind a `true_div_proven_exact` helper) instead of two independent `to_f64_proven_exact` calls, so the emitted code's precondition is exactly the proved property. Add a lowering test for a ratio whose reduction cancels a >53-bit common factor.

Everything else in the remediation checks out: the slice-step panic path is closed at the single lowering site that can construct `HirExpr::Slice` with a step (`crates/sifr_lowering/src/lower/expressions/core_and_calls.rs:618` → `slice_validation::lower_proven_nonzero_slice_step`), and all three `new_known_nonzero` codegen call sites (`stmt_expr_stepped_slice.rs:107`, `leaves_and_plain_calls.rs:487`, `await_and_async_comprehension.rs:437`) are fed only from that site or from the already-guarded `range` lowering; the dead `exact_int_division_requires_handling` helper and the unreachable negative-zero branch are gone; and the module-constant proof reuse in `statically_safe_bounded_integer_arithmetic` stays consistent with the `pow_known_valid` / `shl_known_valid` / `shr_known_valid` emission path (the literal-only shift branch at `stmt_expr_binop.rs:366` is unreachable for exact operands, so broadening to module constants does not strand codegen).

Follow-up findings:
- pre-existing issue: removing the negative-zero branch in `crates/sifr_runtime/src/int.rs:612` makes `0 / -5` yield `+0.0`, while Python yields `-0.0`. The removed branch was indeed unreachable (it tested the numerator's sign, and a zero `BigInt` has `NoSign`), but the correct sign source is the *denominator*. Base behavior was identical, so this is not a regression; worth recording if Python float-sign parity is in scope for a later item.
- suggestion: when `lower_proven_nonzero_slice_step` rejects a step, `lower_subscript` continues with `step = None` rather than propagating `None`, so the recovered HIR describes a step-less slice. This is harmless today because `error_with_code_at` fails the compile, but it can produce a misleading cascading diagnostic on the same expression; propagating the failure would match the `start`/`stop` handling two lines above.

yaseralnajjar:
Item 2 gate evidence and current blocker:

- Reviewed compiler candidate `9606b67b84ae5865105415399d647319b455bb99`: focused validation passed (lowering 1047/1 ignored, codegen 1150, runtime 84, expected-failure corpus, exact-integer oracle 8/8 debug and release, workspace Clippy, fmt, file-size, HIR, diff hygiene).
- The sole create-PR gate was run on `9606b67b84ae5865105415399d647319b455bb99` and failed at demo freshness because three demo sources still assumed the old integer fallibility contract.
- Those sources were corrected, and all 248 changed authoritative companions were regenerated and verified fresh. Candidate became `36144c7cedf5ae7d74f8da0e96edb0832e3d6b6d`.
- The sole merge gate was run on `36144c7cedf5ae7d74f8da0e96edb0832e3d6b6d`. Demo freshness and preceding guards passed; it failed because four newly observed exact-integer runtime roots were absent from the governed retained-intrinsic allowlist.
- The allowlist, its self-test, and manifest schema now pass. Current clean head is `68f6a876ef3a7856072e5615a9e41bd411a8201c`.

Under the explicit no-second-gate rule, neither failed gate has been rerun. This draft must not merge until the user adjudicates whether one fresh create-PR and merge gate may run on the final exact SHA.

yaseralnajjar:
Item 2 final evidence

- Compiler candidate: aa97d2ca6d0da1ec5700b02d3f57ef864a450a53
- Test/governance follow-up: 7b3ba45d25e07adabb820c9f80463534060d42ee (no compiler files changed)
- One create-PR gate on aa97d2c: all reached checks passed; runtime-platform cold rebuild exceeded the 120s blocking-step budget after all 28 variants completed with 0 failures and 1 policy skip. Report: target/validation_lane_reports/create-pr.latest.json.
- One merge gate on aa97d2c: core, static, Rust interop, coverage, CPython differential, and all 30 Python interop variants passed; it stopped on three stale diagnostic baselines. Report: target/validation_lane_reports/merge.latest.json.
- Gate-discovered diagnostics reconciliation: 179 baseline variants executed, 178 passed initially; the sole new fixture wording mismatch was corrected. Exact output and exit-code checks for all four touched fixtures now pass, together with all diagnostic governance checks.
- cargo fmt --check, file-size, HIR, driver maintainability, and diff hygiene pass.

No gate was rerun, per the phase rule. SIFR-INT-0006 retirement is owned by Item 3; dead SIFR-TYPE-0901 warning-path retirement is owned by Item 8.

PR #3582, opened 2026-08-30
SQL M0: lock architecture and dependencies

yaseralnajjar:
## Summary

- lock the schema-first SQL architecture, language contracts, ownership, artifact topology, provider capabilities, and verification inventory
- resolve the latest stable compatible parser, driver, TLS, and SQLite dependency set into the root Cargo lockfile
- add offline contract, baseline, and mutation verification plus the external async-cleanup prerequisite

## Validation

- `python3 verification/areas/sql_platform/runner.py`
- `python3 verification/areas/sql_platform/tools/resolve_dependency_baseline.py --verify-authorities`
- `SYNTAQLITE_SQLITE_VERSION=3053002 cargo check -p sifr_sql_dependency_lock --all-features --locked`
- `cargo metadata --format-version 1 --locked`
- qualification dependency graph forbidden-framework scan
- `python3 verification/areas/documentation/runner.py --suite structure`
- `uv run --project verification python -m sifr_verify --self-test`
- `cargo fmt --check`
- Python compile checks, `git diff --check`, and the file-size guardrail

No compiler source changed, so the phase rule skips the Sifr create-PR and merge gates.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Evidence checked (read-only, on `7f3f6bc2`):
- `Cargo.toml:126` and `Cargo.toml:141` restore the AWS-LC-RS substrate, and the explicit feature lists are exactly the upstream default sets: registry `rustls-0.23.43` `default = ["aws_lc_rs","logging","prefer-post-quantum","std","tls12"]` and `tokio-rustls-0.26.4` `default = ["logging","tls12","aws_lc_rs"]`. No feature is added or dropped relative to base `90762c787`, so nothing outside SQL ownership changes behavior.
- `Cargo.lock` `rustls 0.23.43` dependency set at `7f3f6bc2` is byte-identical to base (`aws-lc-rs, log, once_cell, ring, rustls-pki-types, rustls-webpki, subtle, zeroize`); the prior candidate's only deviation (dropped `log`) is reverted. Base→final lock diff contains no removed TLS lines — only additions from the new qualification crate.
- `crates/sifr_sql_dependency_lock/Cargo.toml` selects `tokio-postgres-rustls/aws-lc-rs` and `mysql_async/aws-lc-rs`, both of which expand to `rustls/aws_lc_rs` + `tokio-rustls/aws_lc_rs` per the registry manifests — no ring edge reachable from SQL providers.
- `verification/areas/stdlib_parity/data/network_http_dependency_snapshots.json:46-47` still expects `rustls/aws_lc_rs` and `tokio-rustls/aws_lc_rs`; preserving the provider is what makes leaving that snapshot untouched correct.
- `internal_docs/sql_architecture.md:1481,1637` name AWS-LC-RS consistently; no residual "workspace-selected"/ring wording remains.
- Re-ran the two owned checkers: `check_contracts.py` (contracts ok: milestones=19 providers=3 domains=6 invariants=30) and both self-tests (12 and 8 mutations) pass. `validate_cargo` enforces `root_features ∪ member_features == qualification.features`, so the qualification JSON and both manifests are consistent by construction.
- Remediated suggestions verified in code: profile mutation selects by `area == "sql_platform"` (`check_contracts.py:456`), dependency mutations select by crate name (`resolve_dependency_baseline.py:350`), architecture baseline matches display name and version on one line and matches `| major | \`tag\` |` rows, tag selection paginates and takes the greatest stable numeric key, and the resolver self-test asserts stable ordering plus prerelease rejection.
- `git diff --check` clean; every changed hand-maintained source file is under 900 lines; no compiler crate source is touched.

Follow-up findings:
- suggestion: `verification/areas/sql_platform/tools/check_contracts.py:363` matches a baseline crate by case-insensitive substring of `display_name` within a line. A display name that is a prefix of another crate's name (`rustls` vs `rustls-platform-verifier`) would match the wrong row if the two versions ever coincide. Anchoring on the first table cell would make the check exact.
- suggestion: `verification/areas/sql_platform/tools/resolve_dependency_baseline.py:262` paginates without a page cap; a release authority that keeps returning full pages would loop unbounded. A page ceiling with a `BaselineError` would keep the refresh terminating.
- suggestion: `verification/areas/sql_platform/tools/resolve_dependency_baseline.py:252` keeps an `or ()` fallback in the `max` key even though candidates are pre-filtered to non-`None` keys; the dead fallback would silently rank a malformed tag lowest instead of failing if that filter is ever relaxed.

PR #3587, opened 2026-08-30
Complete checked emitted failure handling

yaseralnajjar:
## Summary

- replace emitted abort, exit, unreachable, unwrap, expect, and silent fallback paths with typed or structurally infallible control flow
- align exact-integer/float, Decimal/BigDecimal, bytes, Python context, try/finally, and collection proof discharge with compiler-validated invariants
- regenerate authoritative demos and exact generated-quality/diagnostic governance records

## Item 3 acceptance

- generated user paths contain no prohibited failure discharge
- compiler invariants fail before source rendering
- typed errors preserve category, payload, source span where applicable, and timing

## Validation

- 1,156 sifr_codegen tests pass
- workspace Clippy passes with warnings denied; formatting and source guardrails pass
- generated Clippy passes for smoke, representative, and all 91 projects
- full generated determinism, demo corpus, intrinsic panic lint, inventory, diagnostics governance, and 262-demo freshness pass
- broad non-E2E suite reached one corrected Decimal runtime-failure fixture; its exact rerun passes

Exact candidate: 2e3867cbe3546e09a94f391672410808315f3b25

Create-PR and merge gates remain pending until the exact final post-review SHA, per the phase gate-once rule.


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_lowering/src/lower/expression_operators.rs:194-201` (exact-int true division), `:212-227` (mixed float/int), `crates/sifr_lowering/src/lower/expressions/call_builtins.rs:543-547` (`float(int)`), helper `crates/sifr_lowering/src/lower/expression_operators/integer_semantics.rs:292`. Criterion: item scope "complete the inherited Item 2 exact-ratio proof/codegen mismatch" — the fix must not change program meaning. Reason: `proven_exact_integer_literal` substitutes the const-integer *fact* for the operand, but `crates/sifr_lowering/src/lower/integer_const_facts.rs` clears facts only on assignment and on branch merge (`restore_const_integer_state_after_branches`); nothing clears a fact for a name assigned later in an enclosing loop body. Base used these facts only to *prove* safety and still emitted the runtime operand, so the staleness was invisible; materialization turns it into silently wrong numeric output. Verified on the candidate binary: `n: int = 0; while n < 3: v: float = float(n); print(str(v)); n = n + 1` emits `let v: f64 = 0.0;` inside the loop and prints `0 0 0` instead of `0 1 2`; likewise a `for` body doing `q: float = x / 2 ... x = x + 2` emits `let q: f64 = 4.0 / 2.0;` and prints `2 2 2` instead of `2 3 4`, and `x * 1.5` emits `4.0 * (1.5_f64)`. Correction: either restrict `proven_exact_integer_literal` to syntactic literals (`literal_exact_integer_value`, which the same diff already switched to at `integer_semantics.rs:112` for pow/shift) and keep emitting the runtime operand otherwise, or clear const-integer facts for every name assigned anywhere in a loop body before lowering that body; add lowering tests for `while`/`for` bodies mutating a pre-loop int used in `/`, `float()`, and mixed-float arithmetic — neither the demo corpus nor the E2E suites cover this shape.

Follow-up findings:
- pre-existing issue: `crates/sifr_codegen/src/stmt_support_emitter/structured_return_if_while.rs:16` — the none-like return normalization is gated on `TryClosureReturnWrap::Direct`, so a bare `return` inside a `try`/`except` that promotes bindings, in a `Result[None, E]` function, emits `Ok(ControlFlow::Break(Ok(None)))` against `ControlFlow<Result<(), E>, _>` (E0308). Base is broken identically (`Ok(Some(Ok(None)))` vs `Option<Result<(), E>>`), so not a regression, but the new carrier was the natural place to fix it.
- pre-existing issue: `break`/`continue` inside a `try`/`finally` nested in a loop emits `cannot break/continue inside of a closure` (E0267); the closure carrier predates this diff (base `loops_try_finally.rs:712-754`).
- pre-existing issue: `raise` in a function declared `-> int` type-checks clean but emits `return Err(...)` from a non-`Result` signature (E0308) — type-system gap, unrelated to the Item 3 contract.
- infrastructure issue: `plans/phases/34_generated_code_quality_and_production_readiness.md:432,450` still assert `SIFR-INT-0006` "continues to fail closed" after this candidate retires it.
- suggestion: `crates/sifr_codegen/src/ir_validate.rs:277-330` checks `RustExpr::MacroCall` names but not `RustExpr::FormatMacro` (handled at `:343`), which is the exact variant the removed `unreachable!` discharges used; no live path emits it today, but the guard the item adds has a hole in the shape it was built for.
- suggestion: materialized literals leave the original bindings unread (`let a: SifrInt = ...` with no reader), adding to `unused_variables` debt owned by Item 8.

Verified in passing on the candidate: demos contain zero `panic!`/`unreachable!`/`unwrap`/`expect`/`process::exit|abort`; `validate_items` runs before `Renderer::render_file` at `entrypoints.rs:228` and `lib_modules_and_codegen.rs:772`; no dangling `SIFR-INT-0006` references remain in `crates/`, `docs/`, `verification/`; all changed hand-maintained files are under 900 lines (max 895); spot-regenerated `decimal_arithmetic`, `bytes_constructors`, `no_runtime_panics`, `union_narrowing` companions match byte-for-byte; unprovable division and unprovable slice steps produce typed diagnostics with spans (`SIFR-TYPE-0002`, `SIFR-INT-0005`).

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- None.

Evidence checked (read-only, exact SHAs 2e3867cbe → 229c26879):
- Invalidation is ordered correctly at every loop entry: `while_loop.rs:13` (before condition lowering and before `save_const_integer_state`), `control_flow.rs:771` (after the once-evaluated iterable, before the snapshot), `async_for.rs:202` (same shape). All three loop dispatch sites in `statement_dispatch.rs:217-222` are covered; no fourth repeatable-body lowering path exists.
- `invalidate_loop_body_const_integer_facts` (`integer_const_facts.rs:32-44`) reuses `collect_current_function_local_bindings` (`state_collection.rs:506`), which walks Assign/AnnAssign/AugAssign/for target+body+orelse/while/if/with/match/try — i.e. nested compound statements — and stops at nested function bodies.
- Post-loop merge is sound: the cleared snapshot is the baseline for `restore_const_integer_state_after_branches` (`integer_const_facts.rs:55-70`), so a body assignment cannot resurface as a post-loop constant; `async_for` gained the save/merge it previously lacked.
- Walrus was the one remaining assignment form not collected; it is not a hole — `lower_named_expr` binds via `scope.define` and codegen emits a block-scoped `let` (`leaves_and_plain_calls.rs:605`), so it does not mutate the loop-carried binding.
- Regressions added match the blocker's three mechanisms (`value / 2`, `float(value)`, `value * 1.5`) for both while and for, asserting typed-failure diagnostics rather than folded constants; the E2E fixture's expected sequences (`0:0:0`,`0.5:1:1.5`,`1:2:3` / `2:4:6`,`3:6:9`,`4:8:12`) are per-iteration dynamic values, and it follows the existing pass-fixture convention. Guardrail sizes are within limits (largest touched file 873 lines).

Follow-up findings:
- pre-existing issue: `proven_exact_integer_value` (`crates/sifr_lowering/src/lower/expression_operators/integer_semantics.rs:329-333`) falls back to the module-constant map `ctx.const_integer_values` whenever the scope fact is absent. A function-local that shadows a module integer constant and is cleared (by this new loop invalidation, or by the pre-existing clears at `patterns_and_assignments.rs:313` / `aug_assign_lowering.rs:356`) can still fold the module value. The hazard predates this remediation; worth a name-resolution-aware lookup instead of an `or_else` fallback.
- new mechanism defect: nested-function mutation of a loop-carried integer is not modeled — `collect_current_function_local_bindings` records only the nested function's name, so a closure declared and called in a loop body that rebinds a `nonlocal` int leaves the enclosing fact intact. Same class as the fixed blocker but a different mechanism; per the phase rule this belongs to a later item.
- suggestion: the async-for arm of the fix (`async_for.rs:202`) has no dedicated regression; the two new lowering tests cover only while and for.
- suggestion: nested loops re-walk each inner body once per enclosing level (O(depth × body size)); a single pre-pass or memoized collection would avoid the repeated traversal if loop-heavy sources become a lowering cost.

yaseralnajjar:
Gate evidence for the final PR head 6e7c5b32dc9574a40ff5624834daa768613a0b14: the single create-PR gate reached a stale retained-intrinsic governance entry after all prior checks passed; the entry was removed and both the exact checker and its mutation self-test passed, without rerunning the gate. The single merge gate then passed static, core-language, differential, Rust-interop, coverage, and 29 of 30 Python-interop variants. Its sole failure was sqlite-context: suppressible Python-context error flow emitted a unit Suppress arm where the enclosing try carrier required a Result. The review allowance for Item 3 was already exhausted, so this distinct mechanism is assigned to immediately following Item 3A and neither gate is rerun.

PR #3591, opened 2026-08-30
fix: close residual emitted Rust flow proofs

yaseralnajjar:
## Summary

Closes Item 3A of the emitted-Rust excellence phase:

- model suppressible Python context body errors as typed continuation paths in HIR flow and CFG analysis without weakening infallible-return precision
- remove dead context closure fallbacks and preserve boolean literal patterns
- make exact-integer proof lookup binding-aware across module-constant shadowing
- invalidate loop and expression facts for direct, transitive, and callable-alias nested-function mutation effects
- emit captured nonlocal rebindings with mutable outer storage
- add sync/async context, while/for/async-for, shadowing, and nested mutation regressions

## Validation

- `cargo test -p sifr_ir`
- `cargo test -p sifr_lowering -p sifr_codegen`
- `cargo test -p sifr -- --skip test_e2e_pass`
- focused runtime execution of `loop_integer_float_fact_invalidation.sifr`
- focused runtime execution of `async_for_stream_result.sifr`
- focused compiled SQLite Python-context example (`total=71` marker observed)
- `cargo fmt --check`
- Item-owned Clippy passes with only three untouched workspace failures relaxed; strict workspace Clippy currently stops on pre-existing `needless_borrows_for_generic_args` in `annotation_resolution.rs` and `needless_pass_by_value`/`expect_used` in `structural_record_codegen.rs`.

Exact-SHA Opus review and the single create-PR/merge gates will be attached before ready-for-review/merge.

### DIFF
--- a/crates/sifr/tests/e2e/pass/async_for_stream_result.sifr
+++ b/crates/sifr/tests/e2e/pass/async_for_stream_result.sifr
@@ -31,4 +31,18 @@ async def main() -> Result[None, StreamError]:
     assert total == 6
     assert count == 4
     assert stream.current == 4
+
+    dynamic: list[float] = []
+    seed: int = 0
+    second_stream: CounterStream = CounterStream(0, 3)
+    async for _value in second_stream:
+        try:
+            converted: float = float(seed)
+            dynamic.append(converted)
+        except FloatOverflowError as error:
+            assert False, str(error)
+        except FloatPrecisionLossError as error:
+            assert False, str(error)
+        seed = seed + 1
+    assert dynamic == [0.0, 1.0, 2.0]
     return None
--- a/crates/sifr/tests/e2e/pass/loop_integer_float_fact_invalidation.sifr
+++ b/crates/sifr/tests/e2e/pass/loop_integer_float_fact_invalidation.sifr
@@ -1,3 +1,6 @@
+SHADOWED: int = 8
+
+
 def main() -> None:
     while_values: list[str] = []
     value: int = 0
@@ -32,3 +35,31 @@ def main() -> None:
             assert False, str(error)
         value = value + 2
     assert for_values == ["2:4:6", "3:6:9", "4:8:12"]
+
+    nonlocal_values: list[float] = []
+    captured: int = 0
+    while captured < 3:
+        try:
+            converted: float = float(captured)
+            nonlocal_values.append(converted)
+        except FloatOverflowError as error:
+            assert False, str(error)
+        except FloatPrecisionLossError as error:
+            assert False, str(error)
+
+        def advance() -> None:
+            nonlocal captured
+            captured = captured + 1
+
+        advance()
+    assert nonlocal_values == [0.0, 1.0, 2.0]
+
+    SHADOWED: int = 1
+    SHADOWED = SHADOWED + 1
+    try:
+        shadowed_value: float = float(SHADOWED)
+        assert shadowed_value == 2
+    except FloatOverflowError as error:
+        assert False, str(error)
+    except FloatPrecisionLossError as error:
+        assert False, str(error)
--- a/crates/sifr_codegen/src/hir_analysis/queries.rs
+++ b/crates/sifr_codegen/src/hir_analysis/queries.rs
@@ -5,6 +5,8 @@ pub(crate) use value_liveness::*;
 #[cfg(test)]
 mod default_expression_tests;
 #[cfg(test)]
+mod python_context_flow_tests;
+#[cfg(test)]
 mod tests;
 #[cfg(test)]
 mod type_and_operator_tests;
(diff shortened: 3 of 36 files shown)


yaseralnajjar:
Item 3A initial exact-SHA Opus review for `8b7b46cd629e6530d693462e10590ec287b931c3`: **NOT SATISFIED**. Blocking finding: `proven_exact_integer_value` gated module-map fallback on `BindingKind::ModuleConstant`, but imported constants are registered as ordinary module-frame bindings. This regressed exact-value proofs for imported integer constants. Required remediation: distinguish lexical shadowing by whether the visible name resolves to the module frame (for example, `Scope::resolves_to_module_binding`) rather than by binding kind. This is the sole remediation batch for this item. Non-blocking follow-ups from the review will be recorded in the phase ledger for later ownership.

yaseralnajjar:
Reviewed the full base→candidate diff and, in detail, the remediation delta (`8b7b46c..719bd96`).

**Blocker check.** `crates/sifr_lowering/src/lower/expression_operators/integer_semantics.rs:329-335` now gates the module map fallback on `ctx.scope.resolves_to_module_binding(name)` instead of `BindingKind::ModuleConstant`. That helper (`crates/sifr_lowering/src/scope.rs:195`) resolves innermost-first and returns true only when the winning frame is frame 0, so it accepts imported constants (defined into the module frame via `Scope::define` in `lower/imports.rs:43-54` and `private_stdlib_imports.rs:230`) and rejects any function-local or parameter shadow. This is exactly the correction the prior review required, and it is pre-existing shared machinery (also used by `python_interop/callsite.rs:104`, `rust_callback_callsite.rs:65`), not a new one-off predicate.

**Soundness of the widened fallback.** Every writer of `ctx.const_integer_values` records an immutable module-frame name: `remember_module_const_integer` (module-constant lowering only, `module_constants_lowering.rs:68,208`, each followed by `define_module_constant`) and the two import paths. Sifr has no `global` statement, so a module-frame integer name cannot be rebound from inside a function; the widened fallback cannot serve a stale value where the old `ModuleConstant` filter would not have.

**Coverage.** Both directions are pinned: `cleared_local_fact_does_not_fall_back_to_shadowed_module_constant` (local shadow must not inherit the module fact) and the new `imported_integer_constant_retains_exact_value_proof`, which uses a real multi-module `ExternalDefs` with `constant_integer_values` and would have failed under the pre-remediation code. The new `lower_source_with_externals` helper sits alongside the existing support helpers rather than duplicating harness code. The rest of the item (carrier rejoin in `python_context/sync.rs` and `async_context.rs`, `cfg.rs` continuation handling, nested-function `call_effects.rs` propagation, fixtures and snapshots) is untouched by the remediation, so the prior round's assessment of those criteria stands.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr_lowering/src/lower/annotation_resolution.rs:144` carries a `needless_borrows_for_generic_args` Clippy finding, and a broad all-target Clippy probe surfaces further lint debt outside this item; both are untouched by this diff and belong to Item 8.
- suggestion: `ctx.const_integer_values` is keyed by bare name with no binding identity, so its correctness now rests on the invariant that only immutable module-frame names are ever inserted (module constants plus the two import paths) and that Sifr has no `global` rebinding. That invariant holds today but is implicit; a debug assertion at the insertion sites, or keying by `BindingId`, would make it enforced rather than assumed if module-level mutable globals are ever introduced.
- suggestion: fact invalidation (`lower/integer_const_facts.rs`) operates only on `Scope`-carried per-binding facts, so it structurally cannot reach the module map. That is correct given the immutability invariant above, but the two fact sources having different invalidation reachability is worth a short comment at `proven_exact_integer_value` so a future edit does not add a mutable writer to the module map.

yaseralnajjar:
Validation for exact candidate `719bd96ad5b4d11c507b356bd6fece2ab6d4ac3f`: the single permitted `scripts/run_all_tests.sh --profile create-pr` invocation completed all functional checks successfully. It reported a gate failure only for the cold-cache runtime-platform step budget: 230,281 ms versus 120,000 ms; all 28 variants passed with 0 functional failures (1 declared skip). The worktree target had been cleaned per the >20 GiB repository rule, making this the first cold-cache run; AGENTS.md says not to use that run as host-sensitive performance evidence. Per the item rule, this create-PR gate will not be rerun. Report: `target/validation_lane_reports/create-pr.latest.json`.

yaseralnajjar:
Merge-gate evidence for exact implementation candidate `719bd96ad5b4d11c507b356bd6fece2ab6d4ac3f`: the single permitted `scripts/run_all_tests.sh` run passed all Item 3A-owned compiler/runtime/emitted-code checks. Generated Rust passed corpus compilation, panic/abort scan, rustfmt, strict Clippy, determinism, and freshness. The lane had one blocking inventory-record failure: `surface_inventory.json` expects 701 E2E paths while both exact base `f9927f24c9737d8221e16f28c4e28315c02ef0a9` and candidate contain the same 704 paths with identical digest `ef6a17a107fa114027c96eb2947afc71430a781834df64b97df608629dc10b87`; the diff adds/removes no E2E source paths. This is pre-existing Item 8 inventory debt, not an Item 3A regression, and the gate will not be rerun. Report: `target/validation_lane_reports/merge.latest.json`.

PR #3595, opened 2026-08-30
feat: add canonical SQL schema profiles

yaseralnajjar:
## Summary

- add the provider-neutral `sifr_sql_contract` crate with canonical SchemaIR, normalization, fingerprints, diffs, compatible slices, generated profile modules, and runtime manifests
- add package-level named SQL profiles with closed session modes, source containment, stable identities, and archive/publish inclusion
- run checked-in schema sources through the exact locked provider component during package build/check, under the capability-denied component host, then validate and bind the returned authority into the build cache
- add SQL profile architecture documentation and executable verification for offline provider execution, ambient WASI denial, determinism, and contract coverage
- assign the schema-profile SQL verification suites to create-PR, merge, nightly, and release profiles

## Validation

Focused validation passed:

- `cargo test -p sifr_sql_contract` (9)
- `cargo test -p sifr_compiler_component` (14)
- focused driver schema-profile tests (2)
- `cargo test -p sifr_package` (149 passed; the externally unavailable demo-subrepository case was filtered)
- SQL platform schema-profile verification (10 variants)
- coverage matrix
- strict Clippy for the contract, driver library, and package library
- rustfmt, diff hygiene, HIR maintainability, and file-size guardrail

The single create-PR attempt on `04e00c51b4cc5413bcd10675696455bcaec4cf3a` stopped before compilation because the four validation profiles omitted the required `schema-profiles` suite. Commit `99ad2269d9eae60a5319167d033dc86266306c9c` corrects those profile assignments; the profile-assignment matrix passed. Per the phase rule, the create-PR gate was not repeated.

The single merge-gate attempt on `99ad2269d9eae60a5319167d033dc86266306c9c` passed the SQL/compiler static, core-language, CPython differential, Rust-interop, coverage, and preceding Python-interop checks. It stopped at the already-owned Python `sqlite-context` generated-Rust failure. `plans/issues/active/ad-hoc-emitted-rust-excellence.md` Item 3A owns this exact typed-continuation defect. The gate was not repeated.

## Review

- Initial exact-SHA Opus review on `cf8cad6b3735a51c922f662993c624ee94f49578`: BLOCKED. The remediation resolves namespace qualification intent, credential-channel closure, the source-to-provider-to-authority production pipeline, and missing executable verification.
- Sole remediation review on `04e00c51b4cc5413bcd10675696455bcaec4cf3a`: BLOCKED on two later mechanisms: emitted generated-name encoding is not injective, and the claimed compiler profile registry is not yet production-queryable.
- Under the explicit two-review limit, those findings will be recorded for a later SQL phase item; no third review was run.


yaseralnajjar:
I completed the review by reading the governing documents, the full first-parent diff, and the implementation and tests directly. No files were edited.

# Milestone 4 exact-SHA review — `cf8cad6b3735a51c922f662993c624ee94f49578`

Base: `aedf70bf9` (first parent). Scope: `sql_4_schema_profiles` only.

## What is solidly delivered

`sifr_package/src/manifest/sql_profiles.rs` parses named profiles with a closed field set and rejects unknown fields with a credential-specific reason. `graph/derive.rs:191-247` canonicalizes each declared source and proves package containment. `cargo/package.rs:158-165` and `source/include.rs` put schema sources into publish/archive requirements. `sifr_structural_identity::profile_identity` is a domain-separated, length-prefixed nominal hash, so equal schemas under different profile names do not collapse. `sifr_sql_contract` canonical ordering (`BTreeMap`/`BTreeSet` everywhere) makes `schema_fingerprint` order-independent, and `fingerprint.rs:24-30` correctly excludes `source` from the canonical form. `ContextArtifact` is a genuinely provider-neutral transport: the host stays SQL-free, and `validation.rs:133-152` plus `tests.rs:441-448` bind it into envelope validation and the cache identity.

## Blockers

### B1 — Generated profile modules fail for any schema with the same local name in two namespaces
`crates/sifr_sql_contract/src/generated.rs:141-175`

`generated_types` derives the Sifr type name with `local_name` (`generated.rs:238-243`), the last dotted segment only, then errors if any earlier candidate has the same name. A PostgreSQL profile with `app.status` and `public.status` enums — the exact shape the documented example encourages, `search-path = ["app", "public"]` (`internal_docs/sql_schema_profiles.md:33`) — makes `generate_profile_module` return `Err`. There is no per-namespace nesting, no qualified type name, no disambiguation. The whole profile module cannot be produced from a legal schema.

Two secondary defects in the same block: the error kind is `ReservedExport` (`generated.rs:169`) although nothing reserved is involved, so the diagnostic misidentifies the cause; and the acceptance criterion "Compiler-known profile exports cannot collide with generated schema symbols" is implemented as a silent `continue` (`generated.rs:161-163`) with no test for the cross-namespace case at all. `tests/schema_profiles.rs:143-160` only exercises the reserved-name path with a single object.

### B2 — `sql-modes` is an arbitrary compile-time key/value map that reaches the runtime manifest
`crates/sifr_package/src/manifest/sql_profiles.rs:103`, `:206-222`, `:230-237`; `crates/sifr_sql_contract/src/profile.rs:38`, `:90`

`internal_docs/sql_schema_profiles.md:54-57` states: "The parser rejects unknown profile and session fields. Session configuration has no arbitrary option map. Thus, a connection URL, password, credential environment variable, or live-introspection option cannot enter normal compilation." The implementation contradicts this. `sql-modes` is in the allowed profile field list and is parsed by `string_map` (`:458-481`), which accepts **any** key with any non-empty string value. Those pairs land in `SessionContract.sql_modes`, flow through `build_profile_authority` into `RuntimeSchemaManifest.session` (`profile.rs:106`), and are serialized into the schema context artifact's dialect (`build_authority` requires `dialect.modes == session.sql_modes`).

`[sql.profiles.app.sql-modes]` with `database-url = "postgresql://user:secret@host/db"` parses cleanly and is carried end to end. The negative test at `crates/sifr_package/src/sql_profile_tests.rs:21-36` only injects three specific *top-level* keys, so it passes while the vector is open. `sql-modes` is also absent from the profile field list documented in `sql_schema_profiles.md:24-41`, so the field is both undocumented and unconstrained.

### B3 — No compile-time authority path exists; the milestone ships a library, not a pipeline
`crates/sifr_package/src/sql_profiles.rs:19`, `:159-172`; `crates/sifr_sql_contract/src/*`

Milestone 4's purpose is "Establish one exact compile-time schema authority for each named profile." Outside `#[cfg(test)]` and `tests/`, nothing in the workspace calls `resolve_sql_profiles`, `normalize_schema`, `build_profile_authority`, `generate_profile_module`, or `schema_context_artifact`. `sifr_driver`, `sifr_frontend`, and `sifr_codegen` have zero references to `sifr_sql_contract`. Concretely:

- `ResolvedSqlProfile::schema_sources` (`sql_profiles.rs:19`, populated at `:166-170`) is written and never read. The declared `.sql` files are proven to exist and are archived, but their bytes are never opened by anything.
- `SchemaSourceKind` is parsed (`manifest/sql_profiles.rs:117-129`) and stored, then never consumed or cross-checked. `SchemaDocumentKind` is likewise never read inside `normalize_schema` (`normalization.rs:32-56`) and is not carried into `SchemaIr`, so "All three schema-source forms enter the same normalization path" (`sql_schema_profiles.md:88`) is vacuously true — the discriminant has no effect anywhere.
- `generate_profile_module` returns a `String`; it is never written, never registered as an importable `sifr.sql.schemas.<profile>` module, and no frontend resolver knows the path. The criterion "Generated Sifr modules expose profile namespaces … and metadata" is satisfied only as a function return value.
- Nothing digests the schema source bytes into `ProfileAuthority`, so no mechanism ties the `SchemaIr` back to the checked-in files that permanent-01 declares authoritative.

I considered whether this is legitimately deferred to Milestone 7, since a real DDL→objects mapping is provider-owned. It is not: the *dispatch* (read declared sources → send them plus the schema context artifact to the resolved provider component → normalize the response → emit and register the module) is Milestone 4's owned scope, and Milestone 3 already merged a component host and a non-SQL fixture component capable of driving it end to end. As it stands, no compilation of any Sifr program is affected in any way by a `[sql.profiles.*]` block beyond file-existence validation.

### B4 — Verification inventory claims evidence that does not exist
`verification/areas/sql_platform/data/verification_inventory.json:5,6,19,20`

Four rows flip from `planned` to `active`:
- `permanent-02-offline-compilation`, evidence owner "offline build and capability-denial suites" — no offline-build test and no capability-denial test were added. The only related test is a manifest field-name rejection.
- `platform-03-profile-binding` ("ordinary query binds one named profile"), evidence owner "profile identity compile fixtures" — no compile fixtures exist, and query→profile binding is Milestone 6's subject. Marking it active asserts evidence for a capability this commit does not implement.

The milestone's focused-validation list explicitly requires "offline-build and credential-leakage negative tests"; neither exists in executable form.

## Non-blocking findings

1. **Profile fingerprint is host-dependent and path-sensitive.** `crates/sifr_package/src/sql_profiles.rs:45-50` builds `source_files` with `path.to_string_lossy()`, so separators and lossy replacement leak into `build_profile_authority`'s canonical tuple (`profile.rs:121-137`). Same repository, different host OS → different `ProfileFingerprint` and different `RuntimeSchemaManifest`. This also puts file paths inside a fingerprint while `schema_fingerprint` deliberately excludes them (`sql_schema_profiles.md:105`). Store normalized `/`-joined UTF-8 strings, and reject non-UTF-8 paths at parse time.

2. **Generated identifiers are not checked against language keywords.** `generated.rs:245-251` accepts any alphanumeric/underscore name. A legal `CREATE TYPE class AS ENUM (...)` emits `class class(Enum):`, an unparseable module, reported as an opaque downstream failure.

3. **Slice transitive-closure semantics differ from the documentation.** `slice.rs:75-79` uses `or_insert_with(full semantics)`. If an object is both directly requested with two named properties and reachable as a transitive dependency, the narrow set wins, so `sql_schema_profiles.md:118` ("records the complete semantics of each transitive dependency") does not hold for that object.

4. **`verify_compatible_slice` compares dependency sets by exact equality** (`slice.rs:133`). Under `schema-strictness = "compatible"`, adding an unrelated new dependency edge to a required table (a new FK, for instance) fails verification. That is `exact` behavior in the compatible path.

5. **`ExactOverloadSet` encodes an undocumented identity convention.** `slice.rs:160-171` assumes overloads are named `namespace.name.<signature>` and only `SchemaObjectKind::Function`. Operators and casts, both in the object-kind set, have no overload-absence representation. There is no test for this arm.

6. **`package_source` collapses every path provider to the literal `"path"`** (`sql_profiles.rs:152-155`), weakening "exact provider" for local development graphs. `package_id` still disambiguates, so this is cosmetic today.

7. **No artifact count limit.** `validation.rs:133-152` bounds parts, holes, and dependencies against `ComponentHostLimits` but not `context.artifacts`. The only backstop is `max_input_bytes = 4 MiB` (`host.rs:38`, enforced at `:180`), a budget sized for templates. A serialized real PostgreSQL catalog will exceed it; add an explicit artifact limit and revisit the input budget before Milestone 7 depends on it.

8. **`GeneratedProfileModule::lookup_static_symbol` ignores `self`** (`generated.rs:72-79`) and never consults `metadata.schema_symbols`; it is a bare wrapper over `SchemaIr::resolve_symbol`. Either make it a free function or have it verify against the emitted symbol index it exists to guard.

9. **`resolve_symbol` mishandles dot-free identities.** `schema.rs:135-150`: a top-level object such as a catalog `mydb` is matched both exactly and by the `.mydb` suffix rule, so it is reported ambiguous whenever any nested `x.mydb` exists, with no way to name it qualifiedly.

10. **`check_schema_profiles.py` verifies by substring presence** (`:65-75`). It asserts that `pub fn normalize_schema` and the string `"connection URLs, credentials"` appear in the sources; it cannot detect a behavioral regression, and the mutation self-test only mutates the JSON record, never the implementation. This is record validation presented as mechanism validation.

## Verdict

BLOCKED

yaseralnajjar:
## Remediation review — Milestone 4 @ `04e00c51b4cc5413bcd10675696455bcaec4cf3a`

### Prior blocker status

**1. Namespace-colliding generated types — addressed in intent, not sound (see finding 1).** `generated_types` now keys on a full qualified path (`crates/sifr_sql_contract/src/generated.rs:170-198`), reserved-name skipping is gone, and `resolve_symbol` does exact-match-before-suffix (`crates/sifr_sql_contract/src/schema.rs:135-137`). The compiler-known-export half of the criterion is met: every generated name is prefixed `enums__`/`domains__`/`composites__`, so it cannot equal `sql`, `Schema`, `symbol`, etc.

**2. Arbitrary `sql-modes` carrying credentials — resolved.** `sql-modes` is now a `BTreeSet<String>` restricted to `[A-Za-z0-9_-]{1,64}` at manifest parse (`crates/sifr_package/src/manifest/sql_profiles.rs:206-214,450-456`) and re-validated independently in `validate_profile` (`crates/sifr_sql_contract/src/profile.rs:209-221`), with a negative test asserting a `postgresql://user:secret@host/db` mode is rejected and not echoed (`crates/sifr_package/src/sql_profile_tests.rs:40-47`).

**3. Absence of a compile-time source→provider→authority pipeline — resolved.** `prepare_sql_profiles` (`crates/sifr_driver/src/build/sql_profiles.rs:279-357`) is wired into `RootedEntrypointPlan::from_entrypoint` for `PackageProject` (`crates/sifr_driver/src/build/entrypoint.rs:144-149`), so both `build` and `check` run it. It selects one exact `.schema` processor from the locked provider package, resolves the SHA-pinned component, reads and hashes checked-in sources, runs the sandboxed host, validates the response document/kind set (`crates/sifr_sql_contract/src/component.rs:1051-1092`), builds the authority, generates and parses the module, and folds the profile/schema fingerprints into the build-cache identity. Offline-ness is enforced by the host's zero-import rule (`crates/sifr_compiler_component/src/host.rs:189-201`), checked before instantiation.

**4. Verification inventory claims without executable evidence — largely resolved, with one new unbacked claim (see finding 2).** `platform-03-profile-binding` was honestly downgraded `active`→`planned`, and two real cargo cases were added to the manifest and runner (`verification/areas/sql_platform/manifest.json:61-75`, `verification/areas/sql_platform/runner.py:42-47`) backing the offline/positive/property claims.

---

### Findings

**Finding 1 — new-mechanism-defect**
`crates/sifr_sql_contract/src/generated.rs:290` (`emitted_name`), guarded only by `crates/sifr_sql_contract/src/generated.rs:186`

The ambiguity guard compares the structured `path()` vector, but emission flattens that vector with `path.join("__")` and applies `escape_identifier` (keyword → trailing `_`). Neither transform is injective, and no uniqueness check runs on the emitted name. `valid_qualified_name` (`crates/sifr_sql_contract/src/normalization.rs:209-218`) explicitly permits `_` anywhere in a segment, so distinct paths collapse:

- enum `public.a__b` → path `["enums","public","a__b"]`; enum `public__a.b` → path `["enums","public__a","b"]`. Distinct paths, no error, both render `class enums__public__a__b(Enum):` into one module.
- enum `app.class` and enum `app.class_` in the same namespace both render `class enums__app__class_(Enum):`.

The driver's only downstream check is `parse_source` (`crates/sifr_driver/src/build/sql_profiles.rs:345`), which accepts duplicate class definitions, so the later definition silently shadows the earlier one and two distinct schema types collapse into one generated type — the exact failure mode the previous blocker named. The same unchecked escape applies within a type at `crates/sifr_sql_contract/src/generated.rs:131` (enum variants `class` and `class_` both emit `class_ = N`, and `identifier_list_property` de-duplicates only the raw labels) and `:148` (composite fields). `internal_docs/sql_schema_profiles.md:157-158` asserts this name is "collision-free", which the implementation does not guarantee.

**Finding 2 — prior-blocker-unresolved**
`verification/areas/sql_platform/data/schema_profile_qualification.json:38`, enforced only by `verification/areas/sql_platform/tools/check_schema_profiles.py:17`

The remediation added `"compiler-profile-registry"` to `generated_artifacts`. That string appears nowhere in `crates/` or `internal_docs/` — no producer, consumer, or documented contract. The checker "validates" it by comparing the JSON list to a hardcoded copy of the same list, so the assertion is self-referential; unlike `profile_fingerprint_includes`, no token check or test binds it to code. The nearest candidate, `PreparedSqlProfiles` (`crates/sifr_driver/src/build/sql_profiles.rs:250-277`), is not a registry any compiler stage can query: its only non-test reader is `cache_fragment`, and the `module` accessor is `#[cfg(test)]`. This reintroduces the "inventory claim without executable evidence" defect into the same qualification record the remediation was meant to repair.

---

BLOCKED

PR #3597, opened 2026-08-30
feat: add common SQL contracts

yaseralnajjar:
## Summary

- add provider-neutral SQL type, bind, codec, cardinality, effect, and analysis contracts
- add the driver-free `sifr_sql_runtime` ownership and execution boundary
- register structured SQL diagnostics and complete qualification/mutation evidence
- close bounded-decimal, custom-type, SQLite, codec-profile, redaction, and matrix-review gaps

## Validation

- `cargo test -p sifr_sql_contract -p sifr_sql_runtime -p sifr_diagnostics`
- `cargo clippy -p sifr_sql_contract -p sifr_sql_runtime -p sifr_diagnostics --all-targets -- -D warnings`
- SQL-platform area: 14 variants, 0 failures
- coverage-matrix readiness: 4 variants, 0 failures
- file-size and maintainability guardrails pass

## Review status

Two exact-SHA Opus reviews were completed under the phase limit. The second review found three additional M5-local defects; those are corrected at `d33b51374f668dc564aa187910446c16daa59f29`. A policy decision is pending on whether to authorize one exceptional third exact-SHA review of the corrected candidate. The PR remains draft until that decision.


yaseralnajjar:
# SQL Milestone 5 review — `f71d1bb` vs `f83bb85`

**Verdict: REMEDIATE** (7 BLOCKERs, 2 LATER MECHANISM). Full write-up at `/Users/yaseralnajjar/.claude/plans/you-are-the-exact-sha-pure-petal.md`.

## What holds up

Topology is correct: `sifr_sql_runtime` depends only on `sifr_runtime` (`crates/sifr_sql_runtime/Cargo.toml:9-13`), the compiler-only contract crate is not in its graph, and `check_common_sql.py:123-130` pins that surface exactly. Panic containment (`future.rs`, `codec.rs`), `SqlError::Display` redaction, resource accounting, and the pool/connection/transaction/stream send-and-borrow protocol are real and tested.

## BLOCKERS

**B1 — Bounded decimals bind as `Exact`.** `bind.rs:247-261` special-cases only `Text`/`Binary`; `Decimal { precision: Some(5), scale: Some(2) }` with `SifrType::Decimal` returns `Exact`. Binding `12345678.99` is declared infallible but overflows server-side. Fix: `EncodeCheck::DecimalPrecisionAndScale` arm in `constrained_exact`.

**B2 — Custom types cannot round-trip.** `sql_type.rs:215-217` maps `Custom` to the *codec* name; `bind.rs:169-183` requires the registry's `contract.sifr_type`. The commit's own tests assert `Custom{"app.money.v1"}` on read (`tests:186-194`) and `Custom{"app.Money"}` on bind (`tests:329-354`). Reading a `money` column and inserting it back is `Rejected(NominalIdentity)`. `provider.rs:97` already exempts `Custom` from the canonical check to paper over this.

**B3 — Nullable custom result fields skip nullability validation.** Same `provider.rs:97-105` guard, plus `codec_matches` ignores nullability and `null_behavior` is never consulted on read. A `Nullable` custom column typed as non-optional passes `validate`; NULL then surfaces as a `Decode` error for a row the type system called impossible.

**B4 — `SqliteDynamic` has a read contract but no bind rule.** No arm in `special_compatibility`, so binding an `int` to any affinity column is `Rejected(UnsupportedPair)` — only the full 5-member union binds. Since the doc forbids providers widening the relation, SQLite must duplicate the bind table.

**B5 — `CodecRegistry` is keyed globally by `DatabaseType`** (`codec.rs:85-90`) while `ProviderAnalysis::validate` demands a registered codec for every built-in. PostgreSQL binary `int4` and MySQL text `int` collide on `Integer{Signed,Bits32}` → `Err("database type has more than one codec")`. This is the provider-duplication mechanism.

**B6 — Derived `Debug` on `OwnedSqlValue`/`ExecutionRequest`** prints plaintext bound values. Redaction is enforced on `SqlError::Display` and left open on the type that actually holds secrets.

**B7 — The matrix evidence can falsely pass.** `--self-test` mutates only the JSON record, and 7 of the 13 declared `bind_results` are produced by no test. Deleting the sign-mismatch arm (`bind.rs:85-96`) leaves all four profiles green while unsigned/signed binds silently. B1–B5 are all invisible to the gate.

## LATER MECHANISM

- **L1** `Transaction` (`handles.rs:202-228`): `finish` is a dead store, no commit/rollback split, no `Drop` rollback. → transaction/session milestone; add acceptance that a dropped uncommitted transaction issues an explicit rollback.
- **L2** `RuntimeCardinality`/`RuntimeEffectContract` duplicate the contract types (correctly, for topology) but already disagree on the bottom element. → query-substrate milestone; add a lowering round-trip test over the full method × effect cross product.

Non-blocking items (error-kind misclassification in `ExecutionRequest::validate`, inert `preserves_lower_bounds`/`panic_containment`, diagnostic-code literal duplication, ASCII-only `SafeSqlIdentifier`, `account_parameters` overwrite, unfused `ProviderFuture`) are listed in the file.

yaseralnajjar:
# Milestone 5 final review gate — `f7d729ed8`

Reviewed the complete item diff `f83bb85c7..f7d729ed8` (both commits). Read-only; nothing edited or committed.

## Original blockers

**Fixed at root: B3, B5, B6, B7.**
- B3 — `provider.rs:94-110` now applies the null-codec `Reject` check *and* full `canonical_read_type_with_nullability_in` equality to result fields; regression test at `common_sql_contracts.rs:723-747`.
- B5 — `CodecRegistry::for_profile` (`codec.rs:74-108`) filters by server profile; `ProviderAnalysis::validate` pins the analysis to the selected profile (`provider.rs:63`); two-profile test at `:554-593`.
- B6 — hand-written redacting `Debug` for `OwnedSqlValue` (`parameter.rs:39-73`) and `ExecutionRequest` (`provider.rs:128-141`, statement length only, profile dropped); executed at `runtime_contracts.rs:212-215`.
- B7 — the matrix tests now assert the executed family/bind-result sets **equal** the qualification JSON (`common_sql_contracts.rs:300-310`, `:497-503`) with exhaustive matches, and the suite is really wired into `runner.py` / `manifest.json` / all four profiles.

**Not fully fixed: B1, B2, B4.** Details below.

L1 and L2 are correctly still absent.

## Blockers

**N1 — new M5 blocker, introduced by the B1 fix.** `bind.rs:115-123` matches `Decimal|BigDecimal|Numeric` against any bounded `DatabaseType::Decimal` **without comparing `representation`**, and runs before the canonical check at `:73-76`. So `Numeric → Decimal{precision:Some(18), scale:Some(2), representation: Decimal}` returns `Fallible(DecimalPrecisionAndScale)` (accepted), while the same pair with `precision:None, scale:None` returns `Rejected(UnsupportedPair)`. Tightening a column constraint accepts *more* inputs, and a special-value-carrying `Numeric` lands in a finite decimal column with only a precision/scale encoder check. The arm is redundant for every correct pair — `constrained_exact` (`:309-315`) already yields the identical result — so deleting it is the root fix.

**N2 — new M5 blocker.** `bind_compatibility` never normalizes an inline `None` union member, but `array_element_compatibility` does (`split_nested_nullability`, `:291-306`). The two entry points of one closed relation disagree. False green: `InputType{value: Union{Str,None}, nullability: NonNull}` → `SqliteDynamic{Text,Null}` with `NonNull` returns `Exact` (`:196-198`, `:203-247`) — a null-admitting value accepted for a NOT NULL parameter; the same value via the array path is rejected. False red, same root: `:73` compares against the *non-nullable* `canonical_read_type`, so reading a nullable column and re-binding it to that same nullable parameter is `Rejected(UnsupportedPair)`. The docs never say which encoding `InputType.value` must use. The bind matrix has exactly one nullable input and it is the reject case, so the whole nullable axis is unexecuted.

**N3 — unresolved B2 (partial).** `canonical_read_type_in` (`sql_type.rs:240-257`) consults the registry only at top level, then delegates to `canonical_read_type`, whose `Array` (`:191-199`) and `Range` (`:206-212`) recursion has no registry. A `money[]` column with its codec registered in the profile returns `Err(InvalidProvider)`, making every such result field an `InvalidResultField` via `provider.rs:99-110`. Fail-closed, so not unsound, but a registered custom type nested in an array is now unmappable where it previously mapped. The matrix tests `Array` and `Custom` separately and never composes them.

## New later mechanisms

None — every defect is inside M5's own owned surface and fixable without a provider runtime or query substrate.

## Non-blocking

`SqliteDynamic` encodes nullness twice (storage-class `Null` vs `Nullability`); `canonical_read_type_with_nullability` now dead externally; identity-validation mismatch between `RuntimeCodecIdentity::new` and `OwnedSqlValue::Encoded`; decimal `scale > precision` unvalidated. Panic containment, ownership doctests, driver-free dependency surface, the synthetic diagnostics baseline (matches the `sql_3` precedent), the mutation self-test, and the file-size guardrail all check out.

Full write-up with exact line cites and the four required root fixes is in the plan file.

## REMEDIATE

yaseralnajjar:
I reviewed the exact diff `f83bb85c7..d33b51374` — the contract crate (`sql_type.rs`, `bind.rs`, `codec.rs`, `cardinality.rs`, `effect.rs`, `provider.rs`, `diagnostic.rs`), the runtime crate, the matrices/tests, the diagnostics registry and fixtures, and the verification manifest/profiles/checker.

The three prior blocking findings are fixed in this candidate:
- `bind.rs:84-87` now requires canonical representation equality before `constrained_exact` applies precision/scale, so `Numeric` → `Decimal{representation: Decimal}` is `reject-unsupported-pair` (test at `common_sql_contracts.rs:422-430`).
- `bind.rs:58-78` normalizes inline `None`/`None`-unions and `InputType.nullability` once at entry, with matrix cases for literal `None`, inline unions, and the SQLite nullable case.
- `sql_type.rs:170-256` threads the registry through the whole recursion, with a nested custom-array read case at `common_sql_contracts.rs:230-242`.

One issue remains, introduced by the final commit.

Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_sql_contract/src/bind.rs:84` (with `crates/sifr_sql_contract/src/sql_type.rs:212-218`), criterion "provider-neutral SQL types and exact bind mappings for … range … and custom" plus the milestone's read/bind coherence. The final commit made *read* mapping registry-aware through `Range`, but the bind fallback still calls the registry-blind `canonical_read_type`, and `special_compatibility` has no `Range` arm (only `Array`/`List` recurse through the closed relation). Failure scenario: profile `postgresql-18` registers codec `app.money.v1` for `DatabaseType::Custom{public.money}`; a column/parameter of `DatabaseType::Range{element: Custom{public.money}, multirange: false}`. `canonical_read_type_in` now returns `Range{element: Custom{"app.Money"}}` and `ProviderAnalysis::validate` accepts the result field, but `bind_compatibility` with exactly that `SifrType` and that `ParameterType` falls through to `canonical_read_type`, which errors on the custom element, yielding `Rejected(UnsupportedPair)` → `SIFR-SQL-0002`. A value read from the column can never be bound back into a parameter of the identical type. Before this commit both sides rejected, so the asymmetry is new here — it is the same read/bind disagreement class that was blocking in the initial review, closed on the read side only. Correction: use `canonical_read_type_in(&target.database, codecs)` in the fallback at `bind.rs:84` (the top-level `Custom` case is already resolved earlier in `special_compatibility`, so behavior for `reject-missing-codec` / `reject-nominal-identity` is unchanged), or add a `Range`/multirange arm that recurses through `bind_compatibility` like the array arms; add a range-of-custom row to the bind matrix.

Follow-up findings:
- suggestion: `bind.rs:76` — a null or nullable input short-circuits to `Exact` before the SQLite arm, so `SqliteDynamic{storage_classes}` that omits `SqliteStorageClass::Null` still accepts `None` when `ParameterType.nullability` is `Nullable`. Only reachable with a self-contradictory descriptor (nullable parameter whose dynamic storage set excludes `Null`), and `internal_docs/sql_common_contracts.md:61` makes `nullability` canonical, so it is not an omission — but nothing validates the two fields against each other. Either check `Null` membership on the null path or reject the contradictory descriptor in `validate_database_type`.
- suggestion: `codec.rs:128-155` — `validate_contract` does not require a `DatabaseType::Custom` codec's `sifr_type` to be `SifrType::Custom`. A codec declaring, say, `SifrType::Str` for a custom type reads as `Str` (`sql_type.rs:229-237`) but can never bind, since the custom bind arm (`bind.rs:183-197`) requires a `SifrType::Custom` identity match. The registry-aware fallback above also fixes this; alternatively enforce the shape at registration.
- suggestion: `crates/sifr_sql_runtime/src/handles.rs:218-220` — `Transaction::finish` writes `self.active = false` on a value that is dropped immediately; the write is dead. Either drop the field write or have `finish` return a completion result.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

The prior blocker is closed at `crates/sifr_sql_contract/src/bind.rs:84`: the fallback now calls `canonical_read_type_in(&target.database, codecs)`, so it uses the same registry-aware recursion as the read mapping (`sql_type.rs:170-256`, which recurses through `Array`, `Range`/multirange, and resolves `Custom` via `codec_for_database_type`). Read/bind coherence now holds for nested custom types, not just top-level `Custom` (which was already covered by `special_compatibility`). The matrix gains an exact `Range<Custom app.Money>` → `Range<Custom public.money / app.money.v1>` = `Exact` case (`tests/common_sql_contracts.rs:393-408`), pinning the recursive path, and `internal_docs/sql_common_contracts.md:66-71` states the recursive registry-aware rule. The only behavior widening is that a target whose codec maps to a builtin Sifr type is now accepted from that builtin instead of rejected as `UnsupportedPair` — that is exactly the coherence the criterion requires, not a regression; nothing else in `bind_compatibility` changed order or precedence, since all pre-registry pairs still short-circuit in `special_compatibility`.

Follow-up findings:
- suggestion: rejection-reason granularity is coarser inside containers — a `Range<Custom>` target with no codec in the profile falls to `Err(_) => Rejected(UnsupportedPair)` (`bind.rs:86`) rather than `MissingCodec`, and a missing-codec array element is folded into `ArrayElement` (`bind.rs:289`). Diagnostics-only, out of the current criterion; worth considering when provider/diagnostic work lands.

yaseralnajjar:
Create-PR gate attempt on `472cd7d887bd5da9562a5d0500d79ed0f8209e0b` stopped in `diagnostics:rules/code_coverage` because SIFR-SQL-0001 through 0008 lacked typed non-test compiler uses. Per the one-attempt rule, the gate was not rerun. The root fix replaces duplicated string mappings with `DiagnosticCode::SQL_*` references in `sifr_sql_contract`; the affected diagnostics rules area now passes 5/5 variants.

yaseralnajjar:
**Review of f7a3e5a35 (delta from 472cd7d88)**

Checked:
- **No dependency cycle**: `sifr_diagnostics` deps are `ruff_text_size`, `sifr_source`, `schemars`, `serde`, `serde_json`; `sifr_source` depends only on `ruff_text_size`. Neither reaches `sifr_sql_contract`. Cargo.lock delta is exactly the one new edge.
- **Topology/classification**: `sifr_sql_contract` is `first_party_compiler` in `verification/areas/coverage_matrix/data/cargo_metadata_classification.json`, so a diagnostics edge does not cross a runtime boundary. `verification/areas/sql_platform/tools/check_common_sql.py:124-129` constrains only the runtime crate's dependency set (`{"sifr_runtime"}`, no drivers) — the contract crate is unconstrained there, so no data file needed updating.
- **Gate defect actually closed**: `code_coverage.py:104-115` excludes `crates/sifr_diagnostics/**` and any `tests`/`*_tests.rs` path, then requires each active constant to appear in some remaining source. `crates/sifr_sql_contract/src/diagnostic.rs:25-35` is non-test lib source and references all eight `DiagnosticCode::SQL_*` constants, which are all present in `ACTIVE_DIAGNOSTIC_CODES` (`registry.rs:717-724`). No `#[cfg(test)]` stripping issue, no unknown/non-active constant.
- **Mapping fidelity**: variant→constant→string is 1:1 and order-correct against `crates/sifr_diagnostics/src/codes/registry/sql.rs:6-13` (0001…0008), matching the literals removed from `code()`.
- **No API regression**: `code() -> &'static str` is retained and still `const fn` (`DiagnosticCode::code` is `const fn`, `registry.rs:291`); `diagnostic_code()` is purely additive. Existing consumers (`sifr_driver/src/build/sql_profiles.rs`, `sifr_package/src/sql_profiles.rs`) use unaffected surface. `diagnostic.rs` is 77 lines, well under the guardrail.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_sql_contract/tests/common_sql_contracts.rs:601` asserts the typed path for only `ProviderContract`; the remaining seven variants are covered only through the derived `code()` string. Asserting `diagnostic_code()` for all eight (or iterating the variant list) would catch a future constant-level reshuffle directly rather than indirectly. Not required by the item.

yaseralnajjar:
Merge-gate evidence for exact candidate `f7a3e5a3523f9cd90309f457a78ba7220fabe6d5`:

- The single authorized merge-gate run reached and passed repository guardrails, Rust interop (10/10), coverage readiness (4/4), core language, and CPython differential.
- Python interop passed 28/30 variants. The only failures were host-timeout cases `readonly-check-doctor` (300 s) and `binding-authoring` (180 s).
- These same host-variance cases are already documented by the performance-budget owner in `plans/issues/archive/adhoc_performance_budget_host_variance.md` and the Phase 40 evidence ledger. They are unrelated to this milestone's SQL contract/runtime changes.
- Focused validation on this exact SHA remains green: `sifr_sql_contract`, `sifr_sql_runtime`, and `sifr_diagnostics` tests; strict Clippy; diagnostics rules; SQL common qualification; file-size guard.
- Per the phase's one-gate policy, the merge gate was not rerun.

The final exact-SHA Opus review is SATISFIED with no blocking findings.

PR #3601, opened 2026-08-30
Unify checked collection place lowering

yaseralnajjar:
## Summary
- route checked collection reads, writes, deletes, unpacking, and guarded mutations through typed lowering plans
- preserve exact negative-index, missing-key, optional, ownership, and non-mutating membership semantics
- regenerate emitted demo companions and extend focused E2E/codegen coverage

## Validation
- cargo test -p sifr_codegen -p sifr_lowering -p sifr_runtime
- verification/runner/e2e/run_e2e_pass.sh (705 passed)
- cargo fmt --check
- generated-code inventory, demo freshness, maintainability, and file-size guardrails

Draft pending exact-SHA Opus review and final gates.


yaseralnajjar:
## Verdict: **NOT SATISFIED**

The remediation fixes the straight-line form of the prior blocker but not the loop-carried form. Both defects below are in the remediation mechanism itself, reproduced at the exact remediation SHA `a91f43d2b`.

### Defect 1 (blocking, same class as the prior blocker): witnesses are never invalidated across a loop back-edge

Invalidation and re-read only happen along a linear statement sequence: `try_lower_checked_place_mutation_tail_for_ir` (`crates/sifr_codegen/src/checked_place/control_flow.rs:78-140`) is invoked once per statement of the enclosing block (`crates/sifr_codegen/src/stmt_support_emitter/stmt_block.rs:29-35`), removes the witness *after* the mutating statement is lowered (`control_flow.rs:104-106`), and re-guards only the `following` tail. When the mutation is inside a loop body and the witness read occurs *before* it in that body, the hoisted `let Some(__sifr_checked_value_N) = …` stays outside the loop and is re-used on every subsequent iteration.

Reproduced at HEAD (`cargo run -p sifr -- run`):

```python
def for_loop_stale(mut values: list[int]) -> int:
    if len(values) == 0:
        return -1
    total: int = 0
    for i in range(3):
        total = total + values[0]
        values[0] = values[0] + 10
    return total
# main: print(for_loop_stale([1]))
```
Emitted body: `let Some(__sifr_checked_value_0) = …get(0).cloned() else { … };` before the loop, then inside the loop `total = &total + &__sifr_checked_value_0.clone();` and `let __assign_value = &__sifr_checked_value_0.clone() + &SifrInt::from_i64(10);`.
Program prints **3**; CPython prints **33**. Criterion 2 (writes/augmented writes preserve exact semantics) and criterion 1 (one *valid* typed plan) fail.

The same shape with `del values[0]` in the loop also keeps reading the pre-loop witness after the list is emptied, i.e. a read that Python answers with `IndexError` silently yields a stale value.

The dictionary variant does not even compile — the exact E0502 the initial review cited:

```python
def dict_loop_stale(mut mapping: dict[str, int], key: str) -> int:
    if key not in mapping:
        return -1
    total: int = 0
    for i in range(2):
        total = total + mapping[key]
        mapping[key] = 100
    return total
```
```
error[E0502]: cannot borrow `*mapping` as mutable because it is also borrowed as immutable
35 |     let Some(__sifr_checked_value_2) = mapping.get(key) else {   -- immutable borrow occurs here
40 |         total = &total + &(*__sifr_checked_value_2).clone();     -- immutable borrow later used here
45 |                 mapping.insert(__assign_key, __assign_value);    ^^ mutable borrow occurs here
```
(`sifr build` fails with SIFR-BUILD-0005.) So the borrow-lifetime part of the blocker is resolved only for mutations that are direct siblings of the guard, not for mutations reached through a loop body.

### Defect 2 (in-mechanism, criterion 2 — deletion / missing-key): the refresh replays the guard's original missing action

`control_flow.rs:120-127` (and `:460-463`, `:540-548`) rebuilds the fresh read with `witness.missing`, i.e. the exit action that the *membership guard* justified. After a deletion that action no longer models the language semantics:

```python
def dict_del_then_read(mut mapping: dict[str, int], key: str) -> int:
    if key not in mapping:
        return -1
    del mapping[key]
    return mapping[key]
```
emits `let Some(__sifr_checked_value_2) = mapping.get(key) else { return -1; };` — the program returns `-1` where Python raises `KeyError`. The pre-remediation behaviour (stale value) was also wrong, so this is not a regression, but the refreshed plan still does not preserve exact missing-key semantics after deletion.

### Not blocking / out of scope
- Emitted subscript writes use `if let Some(__elem) = values.get_mut(idx) { … }`; here the write is bounds-proven by the surrounding guard, so criterion 4 is not violated by the paths I exercised.
- The rustfmt aggregate exact-debt signature change is Item 8-owned, as stated.
- The parts of the remediation that do work are real: straight-line list/dict exit guards, positive-branch if-let guards, mutable-borrow user calls, and the `fixed_indexing` demo all refresh correctly, and NLL keeps the pre-mutation borrows dead in those shapes.

Required correction: make witness validity account for the loop back-edge — a witness must not survive into a block whose execution can be repeated after any mutation of its dependencies (invalidate before lowering a loop body when the body mutates a dependency, and re-establish the checked read inside the body), and the refreshed read after a mutation must carry the missing action that matches post-mutation semantics rather than replaying the guard's exit.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_codegen/src/checked_place/control_flow.rs:19` (also `:42`, `:307`, `:385`), criterion 2 and 3 — checked-place read witnesses are installed for a whole branch/tail region and are never invalidated when the container is mutated in between, so a read after a write in the guarded region observes the pre-mutation snapshot (list witnesses are `get(..).cloned()`) or produces uncompilable Rust (dict witnesses are live immutable borrows). The witness map is new in this candidate (`git grep checked_place_read_witnesses 7bbb886f4` finds nothing), so these are introduced here, not pre-existing. Verified by emitting/running with the candidate compiler:
  - `def f(mut a: list[int], i: int) -> int: if i < 0 or i >= a.len(): return 0; a[i] = 5; return a[i]` with `a=[1,2,3], i=1` prints `2`; Python/Sifr semantics require `5`. Emitted body binds `__sifr_checked_value_0 = a.get(norm).cloned()` before the write and returns `__sifr_checked_value_0.clone()` after it.
  - Same with `a[i] += 1`: `print(a[i])` and the return both yield `2` while the list is `[1, 3, 3]`.
  - Branch form `if i >= 0 and i < a.len(): a[i] = 5; print(a[i])` prints `2`.
  - Dict forms `if k not in d: return 0` / `if k in d:` followed by `d[k] = 99; return d[k]` fail the Cargo build with `E0502: cannot borrow *d as mutable because it is also borrowed as immutable` (`let Some(__sifr_checked_value_0) = d.get(k)` still live at `(*__sifr_checked_value_0).clone()`), i.e. valid Sifr programs no longer compile.
  - Correction: scope witness validity to the statements that provably precede any mutation of the witnessed container (drop the entry in `lower_checked_read_guards_branch` / `lower_checked_read_branch` / the two exit-guard sites at the first statement that mutates the root path, including aliasing paths such as `HirStmt::SubscriptAssign`, `SubscriptAugAssign`, `Delete`, nested-subscript assigns and mutating method calls), and re-emit a fresh checked read for occurrences after that point. Reject witness installation entirely when the guarded region mutates the container if per-statement invalidation is not tractable in this item.

Follow-up findings:
- suggestion: `crates/sifr_codegen/src/checked_place_mutation.rs:462` — `wrap_sequence_projection`/`wrap_mapping_projection` emit `if let Some(..) = get_mut(..) { .. }` with no `else` arm when `failure` is `None` (statically-proven index). Behaviour is correct only while the lowering-side proof (`guarded_sequence_index_result_type` plus the new `IndexVarNonNegative` guards) is sound; a debug-only `else { unreachable }` or an assert would make any future proof gap fail loudly instead of silently dropping the write (criterion 4 currently holds — verified: out-of-range and negative writes and deletes raise `IndexError` / mutate correctly).
- suggestion: `crates/sifr_codegen/src/hir_analysis/queries/checked_place_reads.rs:6` — `proven_collection_reads` collects reads with no notion of intervening mutation; adding a mutation-aware cutoff here would fix the blocker at its root for every caller rather than at each guard site.

yaseralnajjar:
Exact candidate gate evidence for `a91f43d2bace42c5579d02cf0a9bce57e4962300`:

- The single create-PR gate passed all reached guardrails, Rust interop, coverage, diagnostics, and 23/24 Python-interop variants. It stopped on `callback-examples`: Item 4's replacement of callback vector indexing emitted an underconstrained one-element `try_into()` and Rust reported E0282.
- The single merge gate passed all repository guardrails, Rust interop, coverage, core-language (including full HIR/CFG matrices and 186 audit fixtures), CPython differential, and 29/30 Python-interop variants. It stopped on the same callback-example E0282 regression.
- Neither gate was rerun, per the phase contract.
- The sole remediation Opus review found new mechanism defects in loop-back-edge witness validity and post-deletion missing behavior. Under the no-third-review rule, these findings and the callback decoder regression are assigned to immediate Item 4A before Item 5.

PR #3608, opened 2026-08-30
Close residual checked-place witness lifecycles

yaseralnajjar:
## Summary
- refresh checked-place witnesses at loop back-edges and before repeated while conditions
- invalidate lowering proofs after destructive mutation and loop-carried dependency rebinding
- make callback argument array decoding explicitly typed and panic-free for fixed arities
- suppress unconsumed generated witness guards and regenerate all affected demo companions

## Validation
- focused lowering and codegen tests pass
- aggregate `sifr_codegen`, `sifr_lowering`, and `sifr_runtime` tests pass
- callback examples matrix passes all seven compiled examples
- checked-place E2E fixture passes natively
- generated-code panic scan passes demos, representative E2E, stdlib flows, and negative seeds
- generated demo companions are fresh
- targeted changed-crate Clippy, rustfmt, diff, HIR, file-size, and audit-inventory checks pass

## External validation blockers
- profile resolution for create-pr/merge/nightly/release is currently rejected because each omits required `postgresql-live-differential`
- workspace Clippy is blocked by uninitialized `third_party/libpg_query/18`; changed-crate Clippy passes
- direct full E2E execution bypassing only profile resolution reached 703/705; the two existing serialization-order fixtures fail because isolated generated groups enable `serde_json` default/std without `preserve_order`


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_lowering/src/lower/async_for.rs:202`, criteria 1 and 2. `lower_while` (`while_loop.rs:15`) and `lower_for` (`control_flow.rs:787`) both gained the new `loop_invalidated_sequence_targets` pre-pass, but `lower_async_for` did not, even though it is one of the structured loop entrypoints and its codegen counterpart *was* changed in this commit (`stmt_support_emitter/async_with_and_for.rs:268`). Consequence: a checked-place proof survives the async-for back edge and the codegen loop-carried refresh replays the earlier guard exit — exactly the behavior criterion 2 forbids. Reproduced on the candidate SHA: a program with `if key not in table: return None` dominating `async for value in stream: total = total + table[key]; try: del table[key] except KeyError: pass` compiles and emits `let Some(__sifr_checked_value_0) = table.get(&key) else { return Ok(()); };` at the top of the async loop body; at runtime iteration 2 takes the stale guard exit, so `main` returns early and never prints — silent wrong behavior. The byte-identical sync-`for` version is correctly rejected (`SIFR-TYPE-0005: unsupported operand type(s) for +: 'int' and 'None | int'`). Correction: in `lower_async_for`, next to the existing `invalidate_loop_body_const_integer_facts(ctx, &for_stmt.body)` call, add the same loop over `loop_invalidated_sequence_targets(&for_stmt.body)` calling `ctx.clear_sequence_guards_for_binding` / `ctx.clear_sequence_guards_for_target`, and add a lowering test mirroring `test_loop_delete_invalidates_membership_guard_before_body_lowering` for `async for`.

Follow-up findings:
- pre-existing issue: generated Rust for `crates/sifr/tests/e2e/pass/checked_place_collection_semantics.sifr` is not warning-clean at base either (13 rustc warnings: `non_camel_case_types` on synthesized union enums, `unused_parens` around `is_some_and` closure bodies, `unused variable: short_first/short_last/short_middle`). The two new `for i in range(...)` helpers add 2 more `unused variable: i` warnings of that same pre-existing class. Witness scaffolding itself is exempt from `unused_variables` because the bindings begin with `_`, so criterion 4's "unused witness scaffolding" is enforced only by the new `checked_place_read_witness_uses` telemetry, not by rustc — the demo regeneration (−416 lines) is the real evidence and `scripts/check_demo_emitted_freshness.py` passes.
- suggestion: `checked_place_loop_condition_refreshes_for_ir` excludes condition-refresh keys from the body's loop-carried set via `already_refreshed`, but does not exclude them from `checked_sequence_loop_guards_for_ir`. A key reachable through both paths emits two `let Some(...) = ... else` re-reads per iteration (loop top plus body top). Correct, just redundant.
- suggestion: the loop-carried refresh also fires on the first iteration, where the pre-loop witness is still valid, so every guarded loop pays one extra `.get()` on entry (visible in `control_flow.rs:445-459`).
- infrastructure issue: workspace Clippy and verification-profile resolution are blocked as reported; I did not re-verify those, nor the seven-case `python-interop-callback-examples` matrix (its fixtures need package/bridge resolution and a prepared venv). Criterion 3 is supported by inspection — `python_interop_callbacks.rs:531-552` now annotates `Result<[::sifr_runtime::python::ForeignObject; N], _>` so `try_into` array-length inference is type-directed rather than pattern-driven, is panic-free, and is arity-generic including N=0 — plus the new assertion in `python_interop_direct_tests.rs:344`.
- Independently re-run on the candidate SHA and passing: `cargo test -p sifr_codegen` (1175), `cargo test -p sifr_lowering` (1077), `cargo fmt --check`, `check_file_size_guardrails.py`, `check_hir_maintainability_guardrails.py`, `check_demo_emitted_freshness.py`, and the native `checked_place_collection_semantics` binary. Loop-else `_broke` semantics, `continue`, nested-loop telemetry isolation, post-inner-loop stale reads, and the common `while i < len(values)` / `while len(stack) > 0` patterns all behave correctly under the rewrite. Scratch artifacts I created (`sifr_output/`, `sifr-query-signatures.json`) were removed; the worktree is clean.

yaseralnajjar:
I reviewed the full `505683af2..13fc41d0d` diff and surrounding code, read-only.

## Round-1 blocker: corrected

`crates/sifr_lowering/src/lower/async_for.rs:204-207` now runs the same pre-pass as the sync loops:

- Placement mirrors `lower_for` exactly (`statements/control_flow.rs:786-790`): after iterable lowering and `invalidate_loop_body_const_integer_facts`, before `ctx.scope.push()` / `lower_stmts(&for_stmt.body, ...)`. The iterable is evaluated once, so clearing after its lowering is correct; `while_loop.rs:15-18` correctly clears earlier, before condition lowering.
- Both `clear_sequence_guards_for_binding` and `clear_sequence_guards_for_target` are called, matching sync `for`, so both rebound-dependency proofs and place-membership proofs are dropped.
- `loop_invalidated_sequence_targets` covers the delete/rebind/`clear|pop|popleft|remove` shapes and does not descend into nested `FunctionDef`/`ClassDef`, so the pre-pass is conservative in the right direction.
- Codegen parity is also present: `async_with_and_for.rs:268` now routes async-for bodies through `lower_checked_sequence_loop_body_for_ir`, so async-for gets the same loop-carried witness refresh as `for`.
- `test_async_loop_delete_invalidates_membership_guard_before_body_lowering` (`guarded_index.rs:521-531`) mirrors the sync `test_loop_delete_invalidates_membership_guard_before_body_lowering` and uses an async-generator shape that other lowering tests exercise successfully.

The remediation introduces no regression: the only other changes are two `i` → `_i` renames in the E2E fixture, which are cosmetic (the loop variable is unused in both new fixture functions).

## Complete item against the criteria

1. **Back-edge validity** — met. Loop-carried refresh (`checked_place/control_flow.rs:446-497`) covers `for`/`async for` bodies; `checked_place_loop_condition_refreshes_for_ir` + `checked_place_while_stmt_for_ir` cover re-evaluated `while` conditions by desugaring to `loop { refresh; if !(cond) { break }; body }`. Borrow ordering is safe under NLL: refreshed dict witnesses (`borrowed: true`, `coll.get(k)`) are consumed before the in-body mutation, and post-mutation reads are handled by the pre-existing per-statement refresh.
2. **Invalidation before body lowering for while/for/async for** — met; the same two-clear pre-pass in all three sites, plus `collection_delete.rs:51-53` clearing on `del`.
3. **Callback fixed-array conversion** — met. `python_interop_callbacks.rs:534-548` annotates `Result<[ForeignObject; N], _>` with `N = handles.len()`, making `try_into` type-directed and arity-generic; failure returns `Err` via let-else, no unwrap/expect. Covered by an assertion in `python_interop_direct_tests.rs`.
4. **No unused witness scaffolding** — met. The `checked_place_read_witness_uses` telemetry gates both guard and loop-carried emission; the demo regeneration removes 300+ lines of dead `let Some(__sifr_checked_value_N) = ... else { break };` blocks. Notably those dead bindings also carried a live `break`, so removal is a correctness improvement, not just cleanup.
5. **break/continue/loop-else/nesting** — met. The `while`→`loop` rewrite emits a bare `break` for the condition-false exit, which does not set `_broke`, so the loop-else path (`loops_try_finally.rs:11-22, 63-86`) still runs the else clause on normal exit while user breaks suppress it. `continue` re-runs the refresh and the condition. Nested-loop use propagation is sound: an inner region marks a key locally satisfied only when it re-establishes it on every inner iteration, otherwise the use bubbles to the outer region.
6. **Root-cause, in scope** — met. All touched files stay under the 900-line guardrail.

The three known blockers (workspace Clippy `third_party/libpg_query/18`, verification-profile preflight, serde_json map-order E2E) are infrastructure/pre-existing and correctly out of scope.

## Verdict: SATISFIED

### Non-blocking follow-ups (worth recording as later phase items)

1. **Async-for lacks the `save/restore_sequence_guards` bracket** (`async_for.rs`). `lower_for` (`control_flow.rs:790, 848`) and `lower_while` (`while_loop.rs:24, 50`) restore the guard set at loop exit; `lower_async_for` never does, and sequence guards live on `LowerCtx`, not on the scope, so `ctx.scope.pop()` does not clear them. A guard established inside the body — e.g. `table[key] = v`, which `maybe_record_dict_assignment_guard` records and which `loop_invalidated_sequence_targets` does *not* collect (only `Name`/`Attribute` rebind targets are collected, not subscript assignment) — survives past an `async for` that may run zero iterations, so a following `table[key]` can be treated as proven. Pre-existing asymmetry, but directly adjacent to this item's mechanism.
2. **`missing: None` loop-carried refresh can make a `while` loop non-terminating** (`checked_place/control_flow.rs:472-484`). When a loop-carried witness has no `missing` action — the shape produced by `lower_checked_read_guards_branch`, i.e. a witness established by an enclosing `if k in d:` — the refresh wraps the *entire* loop body in `if let Some(..) { body }`. Inside a `while`, a place that disappears mid-loop skips the whole body including the loop-variable update, so the loop spins forever rather than exiting or diagnosing. This is only reachable through a mutation codegen sees but the lowering visitor does not (e.g. passing the collection to a `mut`-param function that clears it), which the lowering model currently assumes preserves the proof — so it is gated behind a separate pre-existing assumption, but the failure mode changed from a stale value to a hang. Correct fix is to give the loop form's `missing` action (`break` for `while`, `continue` for `for`/`async for`) precedence over the `IfLet` fallback inside loop bodies.
3. **Regression-test strength.** Both `test_loop_delete_...` and the new `test_async_loop_delete_...` assert only `result.is_err()`. Asserting the specific checked-place diagnostic would prevent them from passing vacuously if an unrelated diagnostic fires on the fixture. I could not run the negative control (pre-fix failure) under the review-only constraint, so this is the one claim in the item I verified by code path rather than execution.
4. **No E2E runtime coverage for the async-for path.** `checked_place_collection_semantics.sifr` exercises sync `for` and `while` back edges only; the `async_with_and_for.rs:268` behavior change has unit-level coverage only.
5. **Minor:** `checked_sequence_loop_guards_for_ir` does not skip keys already covered by a condition refresh (unlike `try_lower_checked_sequence_exit_guards_for_ir`, which skips keys already in `checked_place_read_witnesses`), so a read appearing in both the condition and the body can emit two lookups per iteration. Efficiency only.
6. **Nit:** `async_for.rs:204` shadows the outer `target` binding with the loop variable; `lower_for` avoids this by naming the outer one `target_name`.

yaseralnajjar:
Final exact-SHA gate evidence for `13fc41d0d8e4465305b6bd4402f6f0557be91260`:

- `scripts/run_all_tests.sh --profile create-pr` — invoked exactly once; exited 2 in profile preflight because `create-pr` omits required `postgresql-live-differential`.
- `scripts/run_all_tests.sh` — invoked exactly once; exited 2 in profile preflight because `merge` omits required `postgresql-live-differential`.

Item-level validation is green: full `sifr_lowering` suite (1078 passed, 1 ignored), targeted `sifr_lowering`/`sifr_codegen` Clippy with `-D warnings`, native checked-place E2E, generated-demo freshness, formatting, diff check, HIR maintainability, and repository file-size guardrail. The sole remediation Opus review returned SATISFIED.

yaseralnajjar:
Latest-main integration note: after the two allowed Opus reviews and the single permitted create-PR/merge gate attempts were exhausted on reviewed SHA `13fc41d0d8e4465305b6bd4402f6f0557be91260`, `main` advanced with async-cleanup compiler changes that conflicted in async-for emission. The reviewed commits were preserved and `origin/main` was merged as `f8869ebc24647364e3c9d0862d53a18c43030885`.

The semantic resolution retains the new closable async-for cleanup mechanism and applies Item 4A checked-place loop-body refresh to both ordinary and closable async-for paths. Per the phase ceiling, no third Opus review and no second create-PR/merge gate were run.

Focused post-integration evidence:
- `cargo test -p sifr_codegen -p sifr_lowering -p sifr_runtime` — pass (`sifr_codegen`: 1180 tests; all package/doc suites green)
- targeted `sifr_codegen`/`sifr_lowering` Clippy with `-D warnings` — pass
- native `checked_place_collection_semantics.sifr` — pass
- native `async_for_closable_iterator_error_cleanup.sifr` — pass
- generated-demo freshness — pass
- fmt, diff, HIR maintainability, and 900-line file-size guardrails — pass

PR #3611, opened 2026-08-30
feat(sql): implement verified PostgreSQL runtime

yaseralnajjar:
Implements Milestone 9 of the schema-first SQL platform: the common runtime policies and the raw PostgreSQL provider runtime. Adds verified typestate pools, typed session reset, bounded LRU preparation, exact fetch contracts, consuming transactions, savepoints, fresh retry-safe replay, owned and borrowed streams, cancellation cleanup evidence, redacted SQLSTATE errors, wire codecs, and PostgreSQL 13-18 live qualification.\n\nFocused validation passes: common and provider Rust tests, SQL contract and mutation suites, strict Clippy, formatting, file-size and maintainability guards, and the live PostgreSQL 13-18 runtime matrix. No compiler source changed, so full Sifr create-PR and merge gates are not applicable under the phase rule.\n\nKnown unrelated validation issue: the PostgreSQL compiler artifact checksum hashes the complete workspace Cargo manifests and lockfile, so adding this isolated runtime crate causes the existing compiler mutation case to report guest-source drift even though no compiler source or dependency changed. This remains owned by Milestone 18 artifact-authority hardening.


yaseralnajjar:
## Verification performed

I inspected the complete diff (49 files, ~5000 lines) plus surrounding code, ran the offline SQL qualification tools at the candidate SHA, and recomputed the compiler guest-source identity at both SHAs.

**Verdict: NOT SATISFIED**

**Blocking findings:**

- `verification/areas/sql_platform/tools/postgresql_component_inputs.py:16` + `crates/sifr_sql_postgresql/component-artifacts.json:6` — the reported diagnosis is materially wrong, and this is a real gate regression, not a mutation-only artifact. `guest_source_sha256` hashes root `Cargo.lock` and `Cargo.toml` wholesale; adding the workspace member and `futures-core` entry changes both. I recomputed the digest with the *base* blobs for those two files and the *candidate* contents for every other hashed input: it reproduces the recorded `611a8a84…`, so the only drift source is the two root manifests, confirming no compiler source or compiler dependency changed. But the failure is **not** confined to the mutation suite: the plain, non-mutation case `sql-postgresql-compiler-qualification` (`check_postgresql_compiler.py:84-87`, wired at `runner.py:81`) is in the `postgresql-compiler` suite, which is listed in `verification/profiles/create-pr.json`, `merge.json`, `nightly.json`, and `release.json`. At base SHA `check_postgresql_compiler.py` exits 0; at candidate SHA `518244bc2` it exits 1 with `component guest source checksum has drifted`. Consequence: the SQL-platform area gate that M9 itself extends fails on this candidate in the create-pr and merge profiles, so M9 cannot merge green and the next milestone inherits a red gate. Correction: either regenerate/record `component-artifacts.json`'s `guest_source_sha256` for this candidate, or narrow the hashed input set to the compiler-relevant manifest slice as part of this item; deferring the whole thing to M18 leaves a merge-blocking failure in place. (The other three tools — `check_contracts.py`, `check_common_sql.py`, `check_postgresql_runtime.py` — pass.)

Everything else in the criteria list checks out. Exact pins match M0 and `Cargo.lock` (`tokio =1.53.1`, `rustls =0.23.43`, `tokio-rustls =0.26.4`, `tokio-postgres =0.7.18`, `postgres-types =0.2.14`, `tokio-postgres-rustls =0.14.0`, all `default-features = false`, feature allowlist enforced by `check_postgresql_runtime.py:71-89`); the common crate owns pool/session/cache/verification/transaction/limits policy while the provider owns only driver calls; `PostgresPool<Unverified>` exposes no execution methods; `apply_session` + `verify_session` run on connect, on every acquisition (`prepare_acquired_native`) and after every reset; compatible verification compares absence facts (`verification.rs:91-98`, tested at `runtime_policies.rs:28`); execution/fetch/scalar/warm/cardinality contracts and the `first`/`expect_at_most_one` compiler split are implemented and documented; connection task ownership is aborted on drop and terminal APIs consume; transaction states, savepoints, drop-discard, shielded bounded cleanup with `AsyncCleanupEvidence`, transaction-stream poisoning, and fresh-transaction retry are all present and live-tested; the LRU key carries statement/parameter/result/server-version/schema fingerprints; bounds are documented and enforced; the matrix runner pins `postgres:13`–`:18` by digest.

**Follow-up findings:**

- (suggestion, in this diff) `crates/sifr_sql_runtime/src/statement_cache.rs:81-86` — `invalidate_schema` calls `retain(|key| key.schema_fingerprint == schema_fingerprint)`, i.e. it *keeps* the named schema and drops all others, the inverse of what the name states. It has no callers and no test, so nothing regresses today, but the first caller will get the opposite of the intended behaviour. Rename to `retain_schema` or invert the predicate.
- (suggestion, in this diff) `crates/sifr_sql_postgresql_runtime/src/transaction.rs:420` — `run_retry_safe` hardcodes `TransactionOptions::default()`, so the replay API always issues `BEGIN ISOLATION LEVEL read committed`, overriding the profile's verified `default_transaction_isolation` with no caller override. Since `40001` only arises at repeatable-read/serializable, `RetryPolicy::serialization` can effectively only fire on deadlocks (`40P01`) through this path. Not an M9 acceptance bullet; worth an options parameter.
- (pre-existing) `crates/sifr_sql_runtime/src/lib.rs:3` still says "Provider crates implement `ProviderRuntime`", but `sifr_sql_postgresql_runtime` implements neither `ProviderRuntime` nor uses `ProviderFuture`/`catch_codec_boundary`, and the `Pool`/`Connection`/`Transaction`/`RowStream` types in `handles.rs` are now parallel scaffolding to the real provider types. Worth reconciling when a second provider lands (M14/M15).
- (pre-existing) `ResourceLimitKind::StatementDeadline` and `::StatementCache` are declared but never produced; statement timeouts surface as `SqlErrorKind::Timeout` without resource-limit metadata, and cache bounds are enforced by eviction rather than an error.
- (suggestion) `crates/sifr_sql_runtime/src/pool.rs:76-93` — `open_new` is unconditionally `true`, so the `if !open_new` branch is unreachable; `crates/sifr_sql_postgresql_runtime/src/pool.rs:183` — the trailing `Err(provider_error())` after `finish_connection(..., Err(error))?` is likewise unreachable.
- (later-item work) the new `postgresql-live-runtime` suite is defined in `manifest.json` but not referenced by any profile. This matches the existing `postgresql-live-differential` precedent (operator-invoked), so it is consistent, but neither live suite is gated anywhere.

yaseralnajjar:
Verdict: SATISFIED

Original finding disposition:
- Closed. The recorded identity now matches the deterministic digest: `postgresql_component_inputs.guest_source_sha256(repo_root)` computes `aef6a2fb770687cf70b11af4ff313dd50352df254a91cc2e00445973b05e3e8c`, which is exactly what `crates/sifr_sql_postgresql/component-artifacts.json:6` records at `258f13a9a`. `check_postgresql_compiler.py` (ordinary, not just `--self-test`) exits 0, and the area runner over both suites reports all six variants green, including the previously failing ordinary case `sql-postgresql-compiler-qualification` (`target/verification/areas/sql-platform-results.json`: `total_failures: 0, total_variants: 6`). The blocker was real: the remediation changed no file in the digest input set other than `component-artifacts.json` itself (which is not an input — only `Cargo.toml`, `Cargo.lock`, `build.rs`, `component-sources.json`, `src/**/*.rs`, `wit/**`, `wasi_compat/**` are), so the true digest at `518244bc2` was already `aef6a2…` against a recorded `611a8a…`.

Blocking findings:
- none. The second remediation is correct and self-contained: `crates/sifr_sql_runtime/src/statement_cache.rs:83,85` now retain keys whose `schema_fingerprint != schema_fingerprint`, and `crates/sifr_sql_runtime/tests/runtime_policies.rs:94-105` pins that only the named schema is evicted. `invalidate_schema` has no production caller (grep across `crates/`), so the inversion fix cannot regress provider behavior, and the original live PostgreSQL 13–18 evidence remains valid. The refactor of `cache_key` into `cache_key_with_schema('d','c')` preserves the prior LRU test's inputs exactly.
- Re-validated independently on `258f13a9a`: `cargo test -p sifr_sql_runtime -p sifr_sql_postgresql_runtime --lib --tests` (4+1+0+3+9+7 pass, live suite ignored as designed); `cargo clippy … -D warnings`, `cargo fmt --all --check`, `git diff --check`, `check_hir_maintainability_guardrails.py`, `check_file_size_guardrails.py` (largest touched file 740 lines) all pass; `check_postgresql_runtime.py` and its mutation self-test pass.
- Full M9 criteria re-judged on the phase-base→remediated diff and hold: exact Milestone 0 pins with a raw-driver feature allowlist enforced by `check_postgresql_runtime.py`; `sifr_sql_runtime` owns pooling/leases/reset/cache/cancellation/accounting; evidence × strictness modes with `Compatible` comparing every recorded property including `None` absence facts (`verification.rs:91-98`); `Unverified` pools expose no execution surface (`pool.rs:57-125`); session apply+verify on connect, acquire and reset with transaction-pooler affinity rejected before execution (`session.rs:41`); the five-state machine with terminal transitions and consuming commit/rollback (`transaction.rs`); `ShareSafe` pool vs non-`Send` connection/transaction/stream asserted in `tests/runtime_types.rs`; no automatic retry in `finish_context`, fresh transaction per attempt only via `RetrySafeCallback` (`transaction.rs:405-436`); bounded shielded cancel/cleanup with secondary evidence (`control.rs:78-110`, `pool.rs:169-231`); bounded LRU cache with full semantic identity plus `warm`, `Statement` never public; structured limit errors; exact `ExecutionResult` contracts; and panic containment in codecs, manifest verifier and `ProviderFuture`.

Follow-up findings:
- test-coverage (carried from round one): `crates/sifr_sql_runtime/src/statement_cache.rs:81` — `invalidate_schema` is still exercised only by the new test; no provider path calls it, so schema-change invalidation in practice relies on connection lifecycle. Worth wiring or removing in a later item.
- maintainability (known, the alternative correction): `verification/areas/sql_platform/tools/postgresql_component_inputs.py:16` includes the root `Cargo.toml`/`Cargo.lock` in the guest-source identity, so any unrelated workspace dependency edit invalidates the PostgreSQL component identity even when the compiled guest is unchanged. Narrowing the input set would remove the recurring re-record.
- simplification (new): `crates/sifr_sql_runtime/src/pool.rs:76-93` — `open_new` is bound to a literal `true`, so the `if !open_new` branch is unreachable.
- correctness-minor (new): `crates/sifr_sql_postgresql_runtime/src/codec.rs:220` — a `BOOL` payload whose length is not 1 falls through to the catch-all `Encoded` arm instead of returning `SqlErrorKind::Decode` as every other fixed-width type does. No panic, but the malformed value surfaces later as a conversion failure rather than a decode error.

PR #3612, opened 2026-08-30
Close structured-loop witness state

yaseralnajjar:
Completes Emitted Rust Excellence Item 4B.

- restores sequence-guard state after async-for lowering, matching sync loop semantics
- makes loop-carried and while-condition witness refresh use loop-kind control flow instead of stale guard exits or body-skipping `if let`
- strengthens lowering regressions to assert the exact checked-place diagnostic
- adds native ordinary/closable async-for refresh coverage and terminating indirect-mutation while coverage

Validation before review:
- `cargo test -p sifr_codegen -p sifr_lowering` (1181 codegen; 1079 lowering passed, 1 intentional ignore)
- targeted codegen/lowering Clippy with `-D warnings`
- native checked-place collection fixture
- native ordinary and closable async-for fixtures
- generated demo freshness
- fmt, diff, HIR maintainability, and repository file-size guardrails

The generated-surface inventory remains at the same 716 paths as base; its pre-existing 705 record remains Item 8-owned.


yaseralnajjar:
I inspected the full base→candidate diff, the surrounding lowering/codegen mechanisms, and verified behavior on the candidate SHA (worktree clean at `0e8bdd33a`).

## Verdict: SATISFIED

### What I verified independently

**(1) async-for guard save/restore** — `async_for.rs:219/237` snapshots after the pre-loop `loop_invalidated_sequence_targets` clearing and restores after body lowering, exactly mirroring `statements/while_loop.rs:24/50` and `statements/control_flow.rs:795/853`. Pre-loop invalidations are therefore preserved and body-only proofs cannot escape. The new lowering regression is a true regression test: without the restore, `table[key] = value` in the body would leave a `SubscriptPresent`/`DictContains` guard live and the program would compile. `cargo test -p sifr_lowering guarded_index` → 44 passed. I also ran `sifr check` over every e2e pass fixture containing `async for` — all still type-check, so the restore does not reject previously-accepted programs.

**(2)/(3) loop-kind control flow** — callers pass `&RustStmt::Break` for `while` (`loops_try_finally.rs:51`, `stmt_block.rs:703`) and `&RustStmt::Continue` for `for` (`loops_try_finally.rs:132`, `stmt_block.rs:783`) and both async-for paths (`async_with_and_for.rs:201`, `native_async_for.rs:37`). The loop-carried branch in `checked_place/control_flow.rs:472-480` is now uniform, so branch-established witnesses (no missing action) and guard-exit witnesses both get the loop-kind exit; condition refreshes (`:87`) use `break` inside the `loop { … }` shape built by `checked_place_while_stmt_for_ir`. Confirmed on emitted Rust:
- `branch_witness_missing_breaks_while` → `let Some(…) = … else { break; };` at body top, ahead of `i = &i + 1`, with no body-skipping `if let`.
- `condition_witness_missing_breaks_while` → refresh `break` inside `loop`, no replay of `return -1`.

**(4) semantics** — `loop_else`, `while_else_borrowed_condition`, `nested_loops`, `two_pointer_guard_narrowing`, `guarded_index_narrowing`, `safe_list_index`, `negative_index_mutations`, `subscript_aug_assign`, `dict_missing_key_augassign_checked_error` all run green. In the closable async-for envelope the refresh's `continue` is correctly rewritten to `return Ok(Err(true))` (`python_context/sync.rs:669-677` recurses into `LetElse::else_body`), and in ordinary async-for it lands in the `Match` arm inside the real `loop`.

**(5)/(6)** — lowering regressions now assert `TYPE_UNSUPPORTED_OPERATOR`; the three native fixtures pass (`checked_place_collection_semantics`, `async_for_stream_result` → 12, `async_for_closable_iterator_cleanup` → 101), and the freshness asserts genuinely discriminate a stale cached read from a refreshed one.

**(7)** — bounded, no `unwrap`/`expect` added, all touched files under 900 lines (799/675/521). `cargo test -p sifr_codegen checked_place` → 22 passed; `cargo clippy -p sifr_codegen -p sifr_lowering -- -D warnings` clean; `cargo fmt --check` clean.

### Blocking findings

None.

### Non-blocking follow-ups

1. **`while … else` + compiler-inserted `break` skips the `_broke` marker** (`checked_place/control_flow.rs:476`, `:492`, `:87`). User `break` lowers to `{_broke = true; break;}` (`loops_try_finally.rs:11-20`), but the refresh/guard `break` is bare, so the `else` clause runs even though the loop exited via break — inconsistent with the break semantics 4B just adopted. This is pre-existing in kind (base already emitted a bare `Break` for loop guards at `:492` and as the condition-refresh default), and 4B widens its reach to loop-carried witnesses. If it is folded into the remediation batch, the precise fix is for the `while` callers to pass `RustStmt::Block([Assign(_broke, true), Break])` as `missing` when `loop_else_stack.last() == Some(true)`, reusing the shape in `lower_loop_control_stmt_for_ir`. I could not construct an executable repro without writing files, so treat reachability as code-evident rather than measured.
2. **Straight-line refresh paths still carry the old shape** — `try_lower_checked_place_mutation_tail_for_ir` (`:186-203`) still wraps the tail in a body-skipping `if let` when the witness has no missing action, and `refresh_checked_place_witnesses_after_emitted_stmt` (`:130-139`) still replays `witness.missing` even when emitted inside a loop body. Neither is a back-edge refresh, so neither can cause nontermination, and both sit outside criterion 2's "loop-carried" wording — worth recording in the owning checked-place item rather than expanding 4B.
3. **Test anchoring** — `loop_carried_branch_witness_uses_break_instead_of_skipping_while_progress` locates the loop with `generated.find("while ")` over the whole module; anchoring on the emitted function name would make it robust to prelude changes. Cosmetic.
4. **Gate evidence** — the summary reports targeted suites only; per AGENTS.md the merge gate (`scripts/run_all_tests.sh`) should still run once on this candidate. My spot-checks above cover the highest-risk fixtures but not the 716-fixture corpus.
5. The stale `705` expected surface-inventory record is pre-existing and Item 8-owned; not a defect of this change.

yaseralnajjar:
I inspected the full base→remediation diff plus the surrounding checked-place and loop-lowering machinery (`checked_place/control_flow.rs`, both structured while emitters, the for/async-for emitters, the simple-lowering while path, and the `sequence_guards` save/restore sites in lowering). No files were modified; no git or GitHub actions were taken.

## Verdict: SATISFIED

### Loop-else finding — resolved

- `lower_loop_break_for_ir` (`crates/sifr_codegen/src/stmt_support_emitter/loops_try_finally.rs:10`) is the single canonical break shape, derived from `loop_else_stack.last()`, and `lower_loop_control_stmt_for_ir` now delegates to it — so user `break` and compiler-inserted exits are the same construct by construction, not by parallel maintenance.
- Both maintained structured while emitters compute `missing` **after** pushing `has_else` and pass it to both refresh sites: `loops_try_finally.rs:45-58` and `stmt_block.rs:691-706`. I confirmed these are the only two while paths that participate in checked-place refresh — `lower_stmt/simple_dispatch_and_bindings.rs` (`try_lower_simple_while_stmt`) and `stmt_block_helpers.rs` have no checked-place involvement (grep for `checked_place` across `lower_stmt/` hits only a test file).
- `checked_place_loop_condition_refreshes_for_ir` (`control_flow.rs:88`) now uses the passed action rather than `witness.missing.unwrap_or(Break)`, closing the condition-refresh half.
- The natural-exit break inside the desugared `loop { refresh; if !cond { break }; body }` (`control_flow.rs:109`) correctly stays **bare**, so ordinary condition-false exit still runs `else`. This is the distinction the fix depends on and it is right.
- Nesting is sound: `_broke` is declared per while-else block, so an inner while-else shadows the outer, and a break inside an inner loop *without* else emits bare `break` because `loop_else_stack.last()` is that inner loop. For/async-for pass `RustStmt::Continue` (`stmt_block.rs:783`, `loops_try_finally.rs:138`, `async_with_and_for.rs:201`, `native_async_for.rs:37`), which correctly does **not** suppress `else` — a skipped iteration is not a break.

### Canonical helper — no regression

The `loop_carried` rewrite at `control_flow.rs:473-481` collapses the old `missing`/`IfLet` dual path into unconditional `LetElse`. For for/async-for the old `IfLet` wrapped the *entire* remaining body, which is behaviourally identical to `continue`, so this is a no-op there; for while it is the actual root-cause fix (the `IfLet` skip elided the progress update and could not terminate). `witness.missing` is still live for the mutation-tail path (`control_flow.rs:186-205`) and the `require_missing_body` filter, so nothing became dead.

### Complete item

- **async-for zero-iteration guards**: `async_for.rs:219/237` mirrors the sync for-loop (`statements/control_flow.rs:795/853`) exactly — snapshot after back-edge invalidation, restore after the const-integer merge. Back-edge invalidation correctly survives the restore; body-established guards correctly do not.
- **Regressions non-vacuous**: `loop_carried_branch_witness_uses_break_instead_of_skipping_while_progress` asserts ordering `_broke = true` < `break;` < `i = &i +` and explicitly forbids `if let Some(__sifr_checked_value_` in the loop body; a regression to bare break panics on the missing marker. Both native discriminators are real (7 vs 99 for body-refresh and condition-refresh). The extended async-for fixtures assert 101 and 12, which only hold if the body read is re-fetched per iteration.
- **Diagnostics strengthened**: the two `guarded_index.rs` tests moved from `is_err()` to a specific `TYPE_UNSUPPORTED_OPERATOR` assertion, and the new `test_async_loop_body_guard_does_not_escape_a_possibly_empty_loop` is a genuine new case.
- **Panic-free / bounded / guardrails**: no `unwrap`/`expect` on data in generated or emitter code; `control_flow.rs` is 676 lines. The nested `{ _broke = true; break; }` inside a let-else `else` is accepted by rustc — the native fixture compiles and runs, which is the direct evidence.

I did not re-run the gates (review-only, and this worktree is not mine to build in); the verdict relies on the reported validation for test/clippy/fmt outcomes and on static verification of the mechanism above.

### Blocking findings

None.

### Non-blocking follow-ups (candidates for later sequential items)

1. **Mis-scoped `missing` action in deeply nested mutation-tail refreshes (pre-existing shape, payload extended by this item).** `try_lower_checked_place_mutation_tail_for_ir` (`control_flow.rs:186`) emits `witness.missing` verbatim at whatever block depth the mutation occurs. The outer loop installs its guards with the *outer* loop's action (`control_flow.rs:427`), and that witness stays in `checked_place_read_witnesses` while the inner loop body is lowered. `checked_place_refresh_suppressed_depth` only suppresses at exactly `stmt_block_depth + 1`, so a mutation nested two blocks deep escapes it. Reproducer shape to investigate: outer `while … else`, containing an inner `for`, containing an `if` whose body mutates the guarded collection and is followed by a read of it in that same `if` body. At base this emitted a bare `break` targeting the *inner* loop (already wrong); after remediation it additionally assigns the *outer* `_broke`, so the outer loop's `else` would be suppressed even though the outer loop exits normally. Recording as a separate item because the mis-scoping is pre-existing and the fix is a scoping change to the suppression/witness-install mechanism, not to the canonical break shape.
2. **Second `_broke = true` construction site.** `lower_stmt/simple_dispatch_and_bindings.rs:299` builds the same shape independently for the simple-lowering path. The two currently agree, but "single canonical break shape" holds only within the structured emitter. Worth unifying or at least cross-referencing.
3. **Diagnostic precision for loop-invalidated proofs.** The two updated lowering tests now pin `TYPE_UNSUPPORTED_OPERATOR` — accurate, but the user-facing message for `total += table[key]` after a back-edge invalidation reports an operator failure rather than "the key-presence proof was invalidated on the loop back-edge." Pre-existing; a dedicated diagnostic would be a real UX improvement.
4. **Unit coverage gap for condition-refresh + else.** The `_broke`-before-`break` shape in the *condition*-refresh path is covered only by the native fixture; there is no codegen unit assertion equivalent to the new body-refresh test.

yaseralnajjar:
Final exact-SHA gate evidence for `68981d07cb6d088803d199e8924ecc9ab06d0a91`:

- `scripts/run_all_tests.sh --profile create-pr` — invoked exactly once; exited 2 in preflight because `create-pr` omits required `postgresql-live-differential`.
- `scripts/run_all_tests.sh` — invoked exactly once; exited 2 in preflight because `merge` omits required `postgresql-live-differential`.

Item validation is green: 1181 codegen tests; 1079 lowering tests with one intentional ignore; strict targeted Clippy; native checked-place, ordinary async-for, and closable async-for fixtures; generated-demo freshness; formatting, HIR, diff, and file-size guardrails. The sole remediation Opus review returned SATISFIED.

yaseralnajjar:
Latest-main integration note: after both allowed Opus reviews and both single gate attempts were exhausted on reviewed SHA `68981d07cb6d088803d199e8924ecc9ab06d0a91`, `main` advanced with PostgreSQL runtime work and profile repairs. The delta has no overlap with Item 4B and merged conflict-free as `5b1739b4853523b7a9b81bf1c8f1a6af28497a4c`.

Per the explicit no-second-gate rule, the newly repaired profiles were not rerun. Focused post-integration evidence passed: full `sifr_codegen`/`sifr_lowering` suites, strict targeted Clippy, fmt, diff hygiene, and the 3,488-file size guardrail.

PR #3615, opened 2026-08-30
Close mutation-tail witness continuations

yaseralnajjar:
## Summary
- remove stored witness missing-action continuations and renew straight-line witnesses without replaying outer control flow
- invalidate mutable-call sequence guards during lowering and require explicit re-guarding before checked reads
- share one canonical loop-break constructor and add nested/native/codegen regressions

## Validation
- `cargo test -p sifr_lowering -p sifr_codegen`
- `cargo test -p sifr_codegen checked_place_read_codegen_tests`
- `cargo run -q -p sifr -- run crates/sifr/tests/e2e/pass/checked_place_collection_semantics.sifr`
- `cargo fmt --check`
- touched source files remain below 900 lines

The non-E2E Sifr sweep reaches the pre-existing `numeric_sentinels.sifr` type diagnostic; exact base SHA `6862b4a21e` reproduces the same failure.

### DIFF
--- a/crates/sifr/tests/e2e/pass/checked_place_collection_semantics.sifr
+++ b/crates/sifr/tests/e2e/pass/checked_place_collection_semantics.sifr
@@ -51,7 +51,24 @@ def list_exit_guard_refreshes_after_mut_borrow_call() -> int:
     if len(values) == 0:
         return -1
     replace_first(values)
+    if len(values) == 0:
+        return -1
     return values[0]
+
+
+def nested_mutation_renews_outer_while_witness() -> int:
+    values: list[int] = [1]
+    if len(values) == 0:
+        return -1
+    total: int = 0
+    while values[0] < 3:
+        for index in range(1):
+            if index == 0:
+                values[0] += 1
+                total += values[0]
+    else:
+        total += 100
+    return total
 
 
 def list_witness_refreshes_on_for_back_edge() -> int:
@@ -124,6 +141,7 @@ def main():
     assert dict_exit_guard_refreshes_after_assignment() == 99
     assert positive_dict_guard_refreshes_after_assignment() == 99
     assert list_exit_guard_refreshes_after_mut_borrow_call() == 42
+    assert nested_mutation_renews_outer_while_witness() == 105
     assert list_witness_refreshes_on_for_back_edge() == 33
     assert dict_witness_refreshes_on_for_back_edge() == 101
     assert list_witness_refreshes_before_while_condition() == 3
--- a/crates/sifr_codegen/src/checked_place/control_flow.rs
+++ b/crates/sifr_codegen/src/checked_place/control_flow.rs
@@ -24,17 +24,60 @@ impl RustEmitter {
         used
     }
 
+    fn checked_place_witness_is_invalidated_by_stmt(
+        witness: &super::CheckedPlaceReadWitness,
+        stmt: &crate::HirStmt,
+    ) -> bool {
+        let mut invalidated = false;
+        crate::hir_analysis::traversal::walk_stmts(
+            std::slice::from_ref(stmt),
+            crate::hir_analysis::traversal::TraversalConfig::LOCAL_SCOPE_ONLY,
+            &mut |_| {},
+            &mut |expr| {
+                let (args, mutable_arg_places) = match expr {
+                    crate::HirExpr::Call {
+                        args,
+                        mutable_arg_places,
+                        ..
+                    }
+                    | crate::HirExpr::GenericCall {
+                        args,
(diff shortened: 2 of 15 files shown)


yaseralnajjar:
**Verdict: SATISFIED**

Blocking findings:
- None.

Verification notes (exact SHA `6a849e8d9`, tree clean):
- Criterion 1 — `CheckedPlaceReadWitness::missing` and `witness_with_missing` are deleted outright (`witnesses.rs:82-99`), so no witness can carry a stored break/continue/return/fallback payload into a nested region. The only control-flow-bearing markers left are built inside the region that owns them: `loops_try_finally.rs:37` computes `missing` via `lower_loop_break_for_ir()` *after* `loop_else_stack.push(has_else)`, and `lower_checked_sequence_loop_body_for_ir` (`control_flow.rs:553-575`) inserts those `LetElse` guards at the head of that loop's own body, never inside a nested tail.
- Criterion 2 — `refresh_checked_place_witnesses_after_emitted_stmt` and `try_lower_checked_place_mutation_tail_for_ir` now emit `let b = <option>.unwrap_or(b);` instead of `LetElse`/body-wrapping `IfLet`, so the tail is always emitted straight-line. `prepare_checked_place_witnesses_for_mutation` (`control_flow.rs:100-136`) materializes the borrowed witness as an owned clone *before* the mutation and rewrites `option` to `.cloned()`, keeping the `unwrap_or` well-typed; its `!witness.borrowed` guard makes the second call site idempotent, and the entrypoint's `Ok(false)` arm still emits `compile_error!`, so no silent fallback path. Lowering-side exposure is real: `invalidate_mutable_call_sequence_guards` (`sequence_guards.rs:284-298`) is wired into all five call-lowering paths (regular, `__call__`, method, super-method, class-method, attached API), and the new `mutable_call_sequence_guard_tests.rs` asserts the post-call read now surfaces the optional contract as a type diagnostic.
- Criterion 3 — `lower_loop_break_stmt` (`simple_dispatch_and_bindings.rs:35-46`) is the single constructor; `lower_loop_break_for_ir` delegates to it, and `_broke = true` now has exactly one construction site repo-wide.
- Criterion 4 — native fixture `nested_mutation_renews_outer_while_witness` (while-else + nested for/if mutation and read, 105) plus codegen tests `nested_mutation_renews_outer_witness_without_replaying_loop_exit`, `condition_refresh_uses_current_while_else_break_marker` (asserts one `_broke = true;` from the refresh guard, ordered before the natural `if !(...) { break; }` and the `if !(_broke)` dispatch, with no marker on the natural exit), and the tightened straight-line/positive-branch assertions (`== 2` `let Some(...)` + `.unwrap_or(` counts).
- Regression sweep of the accepted-program change: scanned all 717 `tests/e2e/pass` fixtures for a bare identifier passed to a call and later indexed; only `checked_place_collection_semantics.sifr` matches, and it is the fixture updated with the explicit re-guard. Touched files are all under 900 lines.

Follow-up findings:
- pre-existing issue: `checked_place_witness_is_invalidated_by_stmt` (`control_flow.rs:27-70`) only inspects `mutable_arg_places` on `Call`/`GenericCall`/`MethodCall` arguments; a receiver-mutating call (`obj.method()` with `mut self`, or a class-method path that builds `mutable_arg_places: Vec::new()`) is not treated as invalidation. Soundness there rests entirely on lowering clearing the sequence guard, and `invalidate_collection_flow_facts_for_method` (`mutating_methods.rs:209-227`) only clears for a fixed builtin shrinking-method list. Worth confirming under the guard-invalidation issue rather than here.
- suggestion: the `unwrap_or(previous_binding)` refresh silently retains a stale value if the place ever becomes absent. Today that arm looks unreachable (only guard-preserving mutations like `values[i] = …` / `d[k] = v` survive lowering's invalidation), and it matches Python value-binding semantics, but it is an untested defensive default; a debug assertion or a note in the item record would keep the reachability argument explicit.
- suggestion: `prepare_checked_place_witnesses_for_mutation` emits `(*b).clone()` and `option.cloned()` unconditionally for borrowed witnesses. Reads that only go through `checked_place_read_borrow_witness` (`checked_place.rs:629-649`) previously never required `Clone` on the element type; such a program would now fail in `rustc` rather than with a Sifr diagnostic. No fixture exercises it — a `list[NonCopyClass]` case would settle it.
- suggestion: the `let mut _broke = false; <loop>; if !(_broke) { … }` loop-else scaffold is still duplicated verbatim in `loops_try_finally.rs:65-83`, `loops_try_finally.rs:153-172`, and `stmt_block.rs:724-743`. Criterion 3's named debt (`_broke = true; break`) is closed, but the surrounding scaffold is the obvious next unification.

PR #3617, opened 2026-08-30
SQL M10: incremental compiler and editor

yaseralnajjar:
## Summary
- add complete embedded-analysis cache identities, bounded storage, dependency invalidation, and cancellation checkpoints
- add lossless virtual SQL documents and editor features through analysis and LSP
- add structured SQL fixes, budgets, qualification, and documentation

## Validation
- incremental-editor suite: 2/2
- SQL mutation suite: 9/9
- sifr_frontend: 134 tests
- sifr_analysis: 51 tests
- sifr_lsp: 79 tests
- strict Clippy, formatting, file-size, HIR maintainability, and diff checks pass

Draft pending exact-SHA Opus review and repository gates.


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_analysis/src/host/construction.rs:9-10` and `crates/sifr_analysis/src/host/overlay_updates.rs:14-15`, criterion "bounded frontend-owned cache / live analysis wiring must not degrade existing editor behavior". `load_sql_editor_profiles` is now called with `?` inside `open_project` / `open_project_with_overlays`, so any failure of SQL profile preparation aborts construction of the whole `AnalysisHost`. In the LSP, `LspProjectAnalysis::open` (`crates/sifr_lsp/src/analysis_workspace.rs:282-311`) maps that `Err` to `host: None` plus per-document load diagnostics, i.e. every Sifr language feature for the project is disabled. `load_sql_editor_profiles` (`crates/sifr_driver/src/build/sql_profiles.rs:90-105`) gates only on `sifr.toml` + `Cargo.toml` existing and then runs `load_package_graph_snapshot(..., CargoLockMode::Frozen, ...)`, which fails on a package project that has no `Cargo.lock` yet (the exact case `crates/sifr_lsp/src/python_declarations.rs:364-369` special-cases as `missing_lockfile_frozen_failure`) and on any `cargo metadata` failure; `prepare_sql_profiles` likewise returns `Err` for a malformed schema source or an unresolvable provider component. Failure scenario: open a package project that has `sifr.toml` and `Cargo.toml` but no `Cargo.lock` (or temporarily break `schema.sql`) — at base SHA the LSP gave full completion/hover/diagnostics with no SQL involvement; at candidate SHA every open document gets "cannot load the SQL editor package graph"/schema diagnostics and no analysis at all, including projects that declare no `sql.profiles`. Correction: make SQL profile loading non-fatal for host construction — carry the profile-load diagnostics inside `SqlEditorRuntime` (surfacing them through `sql_provider_diagnostics`, as `sql_editor_initialization_diagnostic` already does for component-host errors) and fall back to `PreparedSqlProfiles::default()` so non-SQL analysis keeps working; at minimum reuse the existing missing-lockfile-frozen degradation instead of propagating the error.

Follow-up findings:
- suggestion: no test constructs an `AnalysisHost` with non-empty `PreparedSqlProfiles` (`new_with_sql_profiles` is only reached with `default()` outside `open_project*`). Both new suites (`crates/sifr_analysis/src/host/tests/sql_editor_tests.rs`, `crates/sifr_lsp/src/session/tests/sql_editor_tests.rs`) use single-file/tempdir fixtures, so the corrected production path — `SqlEditorCatalog::from_schema` on a live profile, `request_for_document`, `provider_analysis_from_response` decoding, provider-diagnostic rendering, `slice_keys`/`EmbeddedAnalysisKey` identity, and `invalidate_dependencies` — is never executed under test. `verification/areas/sql_platform/tools/check_incremental_editor.py` checks token presence in the sources, not behavior, so the 2/2 qualification does not cover it. A package-fixture test through `AnalysisHost::open_project_with_overlays` would close this.
- suggestion: `LspAnalysisWorkspace::refresh_projects` rebuilds `LspProjectAnalysis` whenever the open-URI set changes, so each file open in a package project now re-runs `cargo metadata` plus WASM schema normalization and discards the per-host `SqlIncrementalAnalysisCache` and `slice_keys`. The bounded cache therefore rarely survives a session; consider hoisting profile preparation and the SQL runtime above the per-project host rebuild.
- suggestion: `SqlEditorRuntime::observed_dependencies` (`crates/sifr_analysis/src/sql_editor_runtime.rs:206-218`) derives observations from the same immutable `PreparedSqlProfiles` snapshot the cached entries were computed from, so `invalidate_dependencies` can never fire within a host lifetime; schema change is handled only by host re-creation. Relatedly, `SqlDependencyIndex::affected_by_observed` only inspects identities that a query already recorded, so a newly *added* schema object cannot invalidate a `SELECT *` entry. Neither is reachable today given the reconstruction behavior, but both make the invalidation mechanism weaker than it reads.
- pre-existing issue: `AnalysisHost::open_single_file` / `open_single_file_overlay` still build the runtime with empty profiles, so single-file editing has no SQL catalog. This matches the profile-in-package design rather than a defect of this item, but it means the analysis and LSP SQL tests above run without a catalog by construction.

yaseralnajjar:
Disposition: this is the single permitted remediation review. Its newly discovered SQL profile-load degradation and cache-lifetime follow-ups are deferred to the next owned phase item under the phase continuation rule; no third review will run. The create-PR and merge gates each stopped before tests at the already-recorded Milestone 18 contradiction between mandatory PostgreSQL live suites and offline profiles. Focused M10, mutation, provider, lint, formatting, and guardrail validation passed on the exact candidate.

PR #3619, opened 2026-08-30
SQL: add locked and confined host tool runner

yaseralnajjar:
Implements Milestone 11 host-tool graph and command runner. Adds a committed tool-graph lock, build-then-direct execution, fail-closed native sandbox capability enforcement, bounded output, application-graph isolation, structured SQL provisioning, behavioral evidence, and documentation.\n\nCandidate: f0790104720bfc943d420c484a1325e25a1d1f53\nBase: 7e8da6870b50e157b51b4f326a5b8ac2c42d11a3


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:

- **High / contract-not-enforced — capability grants are declarative only.** `crates/sifr/src/host_tool_cli.rs:78-88` (spawn), `crates/sifr_package/src/host_tools.rs:63-99` (plan).
  Violated criterion: "file, network, environment, credential-helper, and subprocess capabilities require explicit grants."
  The runner's entire enforcement is `.env("SIFR_TOOL_CAPABILITIES", …)`. `std::process::Command` inherits the full parent environment, filesystem, network and fork rights, so a tool declaring `capabilities = []` still reads `~/.aws/credentials` and `~/.pgpass`, inherits `PGPASSWORD`/`AWS_SECRET_ACCESS_KEY`, opens sockets, writes anywhere in the project, and spawns subprocesses — with no diagnostic and no failure. `environment` and `credentials` in particular are grants for access the child already has unconditionally. `internal_docs/sql_host_tools.md` ("Package review remains the trust boundary for native host code") concedes there is no enforcement, which directly contradicts the criterion checked `[x]` at `plans/issues/active/ad-hoc-schema-first-sql-platform.md:801`.
  Correction: enforce at the spawn boundary — `env_clear()` plus an explicit minimal allowlist (`PATH`, `HOME`, `CARGO_HOME`, `RUSTUP_HOME`, the `SIFR_TOOL_*` set) unless `environment` is granted; drop credential-bearing variables unless `credentials` is granted; apply an OS confinement layer (`sandbox-exec` / Landlock+seccomp / job object) for `project-read`/`project-write`/`network`/`subprocess`. Note that `cargo run` also executes the tool's build scripts and proc macros unconfined, so the confinement point must wrap the produced binary, not `cargo`. If enforcement is deliberately deferred, the criterion must be un-checked, not restated as an environment variable.

- **High / execution defect — the command plan cannot run the documented tool topology.** `crates/sifr_package/src/host_tools.rs:76-90` (`cargo run --package <name> --bin <entrypoint>`).
  Violated criterion: "`sifr <tool-namespace>` runs only the selected package entrypoint" and "Cargo resolves a dedicated tools workspace member separately from application and target packages."
  `cargo run -p` only accepts workspace members. Verified in this repo: `cargo run --offline --locked -p serde --bin whatever` → `error: package(s) `serde` not found in workspace`. `resolve_host_tool_graph` requires the tool package to be a *direct dependency* of the tools member (`host_tools.rs:554-564`), and `exact_package_checksum` explicitly supports registry-sourced tool packages (`checksum` branch) — yet any registry or git tool package, and any path dependency outside the workspace directory, resolves and verifies successfully and then fails at exec with a raw Cargo error. The documented example (`sifr-sql-postgresql-tools` in `internal_docs/sql_host_tools.md` and `internal_docs/sql_architecture.md:2023`) is exactly this unusable case. `crates/sifr/tests/host_tool_cli.rs:52` hides this because the fixture lists `provider` in `[workspace] members`, so the only executable topology is the one the test happens to use.
  Correction: either constrain resolution to reject non-member tool packages with an explicit diagnostic and correct both docs, or replace `cargo run -p` with a mechanism that works for external packages (e.g. `cargo install --locked --root <target>/sifr-tools` into a Sifr-owned prefix, then exec the resolved binary path directly, which also gives you a concrete artifact to hash).

- **High / integrity gap — the hash record is tautological and does not cover what actually executes.** `crates/sifr_package/src/host_tools.rs:222-262`, called from `crates/sifr/src/host_tool_cli.rs:54`.
  Violated criterion: "Cargo.lock, Cargo metadata, and tool config record exact packages, entrypoints, hashes, and capabilities" / "hash drift … fails closed."
  `verify_host_tool_graph` compares the lockfile and tools-manifest against fingerprints computed from the same files milliseconds earlier in the same process (`resolve_host_tool_graph`), and against path-package hashes computed in the same pass. Nothing is persisted between invocations, so there is no baseline to drift from: replacing a path tool package's sources between two `sifr sql …` runs passes verification unchanged. `host_tools_tests.rs:70-83` only passes because the test mutates `Cargo.lock` in the window between the two calls — a window that does not exist in `cmd_host_tool`. Registry integrity is real, but it comes from Cargo's `--locked`, not from this record.
  Correction: persist the resolved identity record (namespace → package id, source, checksum, entrypoint, sorted capabilities, tools-manifest fingerprint, lock fingerprint) as a project-owned, committed artifact, and make `verify_host_tool_graph` compare the freshly resolved graph against that stored record; keep the in-process re-read only as the TOCTOU narrowing it currently is.

- **High / bypass — `.cargo` config and `PATH` can substitute the executed program without touching any recorded hash.** `crates/sifr/src/host_tool_cli.rs:78-80`, `crates/sifr_package/src/host_tools.rs:76`.
  Violated criterion: "`sifr <tool-namespace>` runs only the selected package entrypoint"; review question 3.
  The plan runs `cargo run --target <host>` from the workspace root. Cargo honors `[target.<host-triple>] runner = "…"` from `<workspace>/.cargo/config.toml`, `$CARGO_HOME/config.toml` and parent directories, so the "selected entrypoint" is replaced by an arbitrary program while package id, checksum, lock fingerprint and manifest fingerprint all remain identical. `[build] rustflags`, `[env]`, `RUSTC_WRAPPER` and `target-dir` are equally unrecorded, and `Command::new("cargo")` resolves `cargo` through the inherited `PATH`. Nothing in `sifr_package` reads `.cargo/config.toml` (grep: no references).
  Correction: neutralize the injection surface on the command line (`cargo --config 'target.<triple>.runner=[]' --config 'build.rustflags=[]' …`, which outranks file config) and fold the effective config chain into the recorded fingerprint set, rejecting a `runner`/`rustc-wrapper` for the host triple; resolve the `cargo` program to an absolute path rather than a `PATH` lookup.

- **Medium / isolation gap — contamination fails closed only when a tool is invoked.** `crates/sifr_package/src/host_tools.rs:566-597`; sole caller `crates/sifr/src/host_tool_cli.rs:43`.
  Violated criterion: "Tool code and dependencies never enter target HIR, generated Rust, linker input, sysroot selection, or application artifacts" and "contamination fails closed."
  `target_contamination_diagnostics` runs inside `resolve_host_tool_graph`, which is reachable only from `cmd_host_tool`. `sifr build`, `check`, `emit`, `package`, and `publish` never call it. An application Cargo package that depends on the tools member or on a tool entry package therefore compiles and links that code into the application binary with no diagnostic; the `HostTools` classification (`crates/sifr_package/src/graph/derive.rs:86-89`) only removes the tools member from Sifr *source discovery*, it does not remove a Cargo dependency edge from an application. A project that never runs `sifr sql …` never gets the check at all.
  Correction: run the closure check during package-graph derivation (`derive_package_graph_from_normalized`) whenever `workspace.metadata.sifr.tools-package` is set, so every workspace/package operation fails closed, and keep the tool-invocation call as a redundant pre-exec check.

- **Medium / untruthful acceptance evidence.** `verification/areas/sql_platform/tools/check_host_tools.py:26-98`, `plans/issues/active/ad-hoc-schema-first-sql-platform.md:793-816`.
  Violated criterion: acceptance evidence must be truthful (review question 8).
  `validate()` asserts that `host_tool_qualification.json` equals constants hard-coded in the same script, and `validate_sources()` greps for identifier substrings; the "self-test 5/5" mutates that JSON and observes the same hard-coded comparison reject it. No mechanism is executed, so "host-tools qualification 3/3" is not evidence for any behavioral claim, and grep tokens like `"--locked" in graph` would still pass if the flag were moved into a comment. Concretely, these criteria are checked `[x]` with no executable test anywhere in the candidate: duplicate namespaces are hard errors (`negative_contracts` claims `duplicate-namespace`; the only mechanism is the `toml` crate's duplicate-key error, untested and not the built-in-shadowing case); `missing-entrypoint`; `non-direct-tool-package`; cross-compilation without graph leakage; non-entry into sysroot selection, HIR, generated Rust and linker input; capability grants (see finding 1); and `open_project` — `crates/sifr_analysis/src/host/tests/sql_editor_tests.rs:56` exercises only `open_project_with_overlays`, never `construction.rs:8`.
  Correction: replace the JSON-equality checker with executable negative cases (one `resolve_host_tool_graph` test per `negative_contracts` entry), add an artifact-level contamination assertion for a built application, add a direct `open_project` regression, and un-check any criterion that remains without executable evidence.

Follow-up findings:

- `crates/sifr/src/host_tool_cli.rs:88` — `.output()` buffers the child's stdout/stderr without bound and defers all output to process exit; a chatty or hostile tool exhausts CLI memory and no tool can stream progress. Owner: M11 CLI dispatch (stream with a documented cap), or explicitly deferred to M12 tooling UX.
- `crates/sifr/src/host_tool_cli.rs:118-137` — `provision_profile` shadow-parses the tool's own arguments: `--profile=app` is rejected before the tool runs, and a `--profile` token appearing as another flag's value is miscounted. Owner: M11; make the manifest-validation contract key off the tool's declared command contract rather than positional guessing.
- `crates/sifr/src/host_tool_cli.rs:118` — SQL-specific `sql`/`test`/`provision` semantics are hard-coded in the generic host-tool runner; a provisioning tool under any other namespace gets no manifest validation. Owner: M12 schema-lifecycle tools.
- `crates/sifr_package/src/host_tools.rs:19-23` — `RESERVED_TOOL_NAMESPACES` is hand-maintained and currently matches the clap `Commands` set by coincidence; nothing fails when a new built-in subcommand is added. Owner: M11; derive it from `Cli::command().get_subcommands()` or add an equality test.
- `crates/sifr_package/src/host_tools.rs:496` — `hash_path_package` skips every `.git` and `target` directory at any depth, while Cargo will happily compile `[[bin]] path = "target/…"`; content that is built is excluded from the identity hash. Also `relative.to_string_lossy()` (line 533) makes non-UTF-8 filenames collide. Owner: M11 integrity record (fold into the persisted-record fix).
- `crates/sifr_package/src/host_tools.rs:554` — `direct_dependencies` walks all resolve edges, so a dev- or build-dependency of the tools member qualifies as a tool package. Owner: M11.
- `crates/sifr_package/src/graph/derive.rs:86` — the tools member is identified by package *name* across all packages in metadata, not by workspace-member id; a registry dependency sharing that name is misclassified `HostTools`. Same name-based matching in `graph/workspace.rs:9-19`. Owner: M11.
- `crates/sifr_package/src/graph/workspace.rs:96` — explicit selection of the tools package reports `selected_rust_only`, which is a factually wrong diagnostic for a host-tools member. Owner: M11.
- `crates/sifr/src/host_tool_cli.rs:105` — a signal-terminated tool (`status.code() == None`) is reported as `EXIT_USER_DIAGNOSTIC`; forwarded child codes also collide with Sifr's own 1/2/3 contract. Owner: M11.
- `crates/sifr_analysis/src/sql_editor_runtime.rs:71-77` — the initialization diagnostic is prepended for every queried source, so one SQL init failure is reported once per open file. Owner: M11/M18 editor UX.
- `crates/sifr_sql_contract/src/provision.rs:38-42` — `ProvisionedCleanup` records only a namespace and a resource id; no cleanup command/argument contract is defined anywhere, so "usable for cleanup" is unproven. `ProvisionedCredential::Helper { command }` accepts an arbitrary command string validated only for non-emptiness and control characters. Owner: M12 (cleanup contract) / M11 (helper constraint).
- `crates/sifr/src/cli_model_and_entrypoint.rs:333-335` — `external_subcommand` removes clap's unknown-subcommand error and suggestions; `sifr biuld app.sifr` now loads a Cargo package graph and fails with a host-tool or package diagnostic. Owner: M11 CLI dispatch.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr/src/host_tool_sandbox.rs:306-312` (Linux `sandbox_command`) together with `crates/sifr/tests/host_tool_cli.rs:4`. Criterion: "File, network, environment, credential-helper, and subprocess capabilities require explicit grants," validated under the milestone's own focused/gate suites. Reason: grant enforcement is now the fail-closed native sandbox with no fallback, and the Linux branch hard-requires `/usr/bin/bwrap` or `/bin/bwrap` (plus `/usr/bin/prlimit` when `subprocess` is ungranted). Nothing in the repo provisions or declares that dependency: `verification/runner/sifr_verify/doctor.py` (unchanged here) checks no sandbox binary, and `.github/workflows/local-first-validation.yml` installs only Rust, uv, and CPython on `ubuntu-24.04`. The new integration test is unconditional (no target gate, no prerequisite skip) and asserts `output.status.success()` for the provision, probe, and flood cases; it is pulled into the blocking `sifr_cli_full` suite (`verification/profiles/merge.json`, `test -p sifr -- --skip test_e2e_pass`, `executed_in_merge: true`), which CI runs on Linux. Candidate evidence records only macOS denial coverage, so the entire Linux enforcement path — the only host OS the project's CI exercises — is unexecuted and unprovisioned; on a runner without bubblewrap every sandboxed assertion in that test fails through `run_sandboxed_tool` → `EXIT_INTERNAL_COMPILER_FAILURE`. Correction: declare the sandbox binaries as a validation prerequisite (doctor check plus the workflow install step) and record an executed Linux run of `crates/sifr/tests/host_tool_cli.rs` covering the same environment/project/network/subprocess denials; if a Linux lane cannot be provisioned in this item, make the prerequisite explicit and record the unexercised Linux enforcement path in the owning issue rather than leaving a blocking gate suite dependent on an undeclared host binary.

Follow-up findings:
- suggestion (new mechanism defect, later phase item): `crates/sifr/src/host_tool_cli.rs:39-45` runs `builtin_typo_suggestion` before namespace resolution, so any declared namespace within edit distance 1 (or one adjacent transposition) of a reserved name — `tests`, `runs`, `tree`-adjacent, etc. — is unreachable and reports a misleading "did you mean" error instead of executing its locked entry point. Only exact reserved names are rejected at resolve time, so these namespaces are legal to declare. Suggested correction: consult the resolved graph first and emit the typo suggestion only when the namespace is not declared.
- suggestion: `crates/sifr/src/host_tool_cli.rs:112-147` buffers all tool stdout/stderr until exit; long-running or interactive host tools produce no incremental output, and the 10 MiB combined bound discards the captured output entirely on overflow.
- suggestion: `crates/sifr/src/host_tool_sandbox.rs:255-262` grants unrestricted `mach-lookup` and `file-read*` on `/private/etc` to every confined tool regardless of grants; worth narrowing when the macOS profile is next revisited.
- suggestion: `crates/sifr/src/host_tool_cli.rs:207-232` confines only the tool process. `cargo build` still executes the tool package's build scripts and proc macros unconfined, so the grant set does not bound build-time code execution. Inherent to the Cargo build step, but not stated in `internal_docs/sql_host_tools.md`.
- suggestion: `crates/sifr/src/host_tool_cli.rs:286-298` scans the whole forwarded argument vector for `--profile`/`--profile=` without honoring a `--` separator, so a profile-shaped operand after `--` is counted.
- infrastructure issue: `verification/areas/sql_platform/tools/check_host_tools.py:74-80` now only asserts that four evidence files exist. The behavioral guarantees moved into real tests (correct direction), but the qualification tool no longer detects a file that is renamed into place with the coverage removed.

yaseralnajjar:
Gate evidence for exact candidate f0790104720bfc943d420c484a1325e25a1d1f53:\n\n- create-PR gate (single allowed run): stopped before tests because profile create-pr omits required SQL suites host-tools, postgresql-live-differential, and postgresql-live-runtime.\n- merge gate (single allowed run): stopped before tests because profile merge omits the same required SQL suites.\n\nThe phase record already assigns this repository profile contradiction to Milestone 18. Per the phase execution rule, neither gate was rerun on the unchanged SHA. Linux sandbox prerequisite/qualification and the legal declared-namespace typo collision from the second Opus review will also be recorded for the later closure item.

PR #3621, opened 2026-08-31
feat(sql): add schema lifecycle tools

yaseralnajjar:
## Summary
- add provider-neutral deterministic schema lifecycle artifacts and atomic writes
- add PostgreSQL schema pull, validate, and build host tool with live catalog normalization
- add PostgreSQL 13-18 live qualification and internal lifecycle documentation

## Validation
- cargo test -p sifr_sql_tool -p sifr_sql_postgresql_tools
- cargo clippy -p sifr_sql_tool -p sifr_sql_postgresql_tools --all-targets -- -D warnings
- cargo test -p sifr_sql_contract -p sifr_sql_postgresql
- PostgreSQL live schema-tool matrix 13-18
- SQL platform schema-tools suite
- formatting, HIR maintainability, and file-size guardrails

Draft pending exact-SHA Opus review.


yaseralnajjar:
## Review — Sifr Milestone 12 (`c560ae5d` → `bd8b04dd`, PR #3621)

I inspected the exact diff (2,294 added lines across the listed paths) plus the surrounding contracts the new crates depend on (`sifr_sql_contract::{normalization, generated, profile}`, `sifr_sql_postgresql::catalog`, `sifr_driver::build::sql_profiles`). I did not re-run the validation already recorded in the evidence.

**Verdict: NOT SATISFIED**

Blocking findings:

- **in-scope omission — `crates/sifr_sql_postgresql_tools/src/postgresql_catalog.sql:39` (also `:108`, `:130`, `:23-32`, `:164`); criterion "schema pull normalizes live provider catalogs" and "schema validate compares sources, canonical snapshots, migrations, and optional live state".** The live adapter emits a semantic vocabulary that is disjoint from the compiler's canonical `SchemaIr`, so the two graphs can never compare equal. Live columns carry `database-type-name` as text; the provider's canonical column objects carry `database-type` as a structured `DatabaseType` value (`crates/sifr_sql_postgresql/src/catalog.rs:130`, produced at `:644`). The same split applies to domains (`base-database-type-name` vs `base-database-type`, `catalog.rs:93`), ranges (`subtype-database-type-name` vs `subtype-database-type`, `catalog.rs:109`), casts (`source-type-name` vs `source-database-type`, `catalog.rs:154`), and tables (live emits `name`/`relation-kind`/`definition`; canonical emits `columns`/`primary-key`/`unique-sets`). Since `pull`/`validate` deliberately reuse the profile's own provider and dialect identity (`command.rs:177-182`) so that only object semantics differ, the consequence is unconditional: `schema validate --live` — and every `schema validate` under an `introspection` profile, which forces the live comparison at `command.rs:121` — reports every table, column, domain, range and cast as changed and can never exit 0; `schema pull` against a profile with no snapshot always returns status 2 and, once accepted, writes a canonical snapshot that `PostgresCatalog::from_schema` cannot load back (`catalog.rs:93,130` return "missing property" errors). Nothing in the evidence covers this: `tests/schema_lifecycle.rs` uses synthetic single-key objects, and the live matrix only asserts that the pulled object *kinds* are present and semantics are non-empty (`tests/live_catalog.rs:55-62`). Correction: emit the provider's canonical property names and structured `DatabaseType` values from the catalog adapter (reusing the same `database_value` encoding the DDL path uses), and add one test that pulls a live catalog for a schema also expressed as declarative DDL and asserts `semantic_diff` is empty.

- **in-scope omission — `crates/sifr_sql_postgresql_tools/src/postgresql_catalog.sql:110` and `:120`; criterion "incomplete provider metadata fail closed" / "preserves semantic provider objects".** The adapter hardcodes `'sifr_type', 'str'` for every domain and `jsonb_object_agg(a.attname, 'str')` for every composite attribute, independent of the real base or attribute type. These two properties are exactly what `generate_profile_module` consumes (`crates/sifr_sql_contract/src/generated.rs:182,187`), so an accepted pull of a domain over `integer`, or a composite with numeric fields, writes a `schema.sifr` declaring `value: str` / `field: str` — fabricated metadata in a shipped deterministic artifact rather than a closed failure. The live matrix does not catch it because its fixture uses only text-based types (`run_postgresql_schema_tool_matrix.py:19,20`), and no test builds artifacts from a pulled schema at all. Correction: derive the Sifr storage type from the actual base/attribute type (the provider type registry in `crates/sifr_sql_postgresql/src/types.rs` already maps database types), and fail closed when no mapping exists.

Follow-up findings:

- **suggestion:** `pull --accept` replaces `profile.schema` with live state but keeps the declarative `source_files`/`source_fingerprints` (`command.rs:207-214`), so the resulting `profile_fingerprint` asserts those sources produced a schema they did not.
- **suggestion:** no test exercises `run_schema_command` beyond the argument parser; `pull`, `validate`, and `build` have no end-to-end coverage against a real profile, which is why both blockers above pass the current gates.
- **suggestion:** unused manifest entries — `sifr_sql_postgresql`, `sha2` and `serde` are declared as dependencies of `sifr_sql_postgresql_tools` but unused in its `src/`, and `semver` in both new crates is used only by tests (belongs in `[dev-dependencies]`).
- **suggestion:** `parse_command` accepts `--profile --accept` by consuming the flag as the profile name (`command.rs:273-276`); rejecting a value that begins with `--` would keep the closed surface tight.
- **suggestion:** `read_snapshot` hardcodes `"schema.json"` (`command.rs:217`) instead of the exported `SNAPSHOT_PATH` constant it must stay in sync with.
- **suggestion:** `internal_docs/sql_schema_tools.md` states the diff is printed "before any write", but stdout is buffered in `CommandOutcome` and only printed by `main.rs:23` after the artifact replacement has already run.

yaseralnajjar:
I reviewed the full base→remediation diff (`c560ae5d..ba3629d6`), with focus on the two recorded blocking findings.

**What I verified**

- Live→DDL vocabulary parity: `crates/sifr_sql_postgresql_tools/src/normalization.rs` rebuilds every kind's semantics from the raw catalog rows, and the key/value shapes now match the compiler provider's DDL output exactly for namespace (`{}`), column (`name`/`database-type`/`nullable`/`has-default`/`generated`), table/view (`columns`/`primary-key`/`unique-sets`/`constraints`|`provider-query`), PK/unique/FK, index, sequence, enum, domain, composite, range, function, operator, cast. Type names go through `LibpgQueryParser` → `PostgresTypeRegistry` → structured `DatabaseType`, i.e. the same values `catalog.rs` emits; unresolvable names error (`resolve_type`, `incomplete`).
- Reload path: `PostgresCatalog::from_schema` reads `base-database-type`, `subtype-database-type`, `database-type`, `primary-key`, `unique-sets`, `arguments`/`result`, `source-/target-database-type` — all now produced by the live pass. Domain/range nominals are pre-registered with a fixpoint loop before object normalization, so dependent domains resolve.
- Fabricated `str` types are gone: `generated_sifr_type` (`crates/sifr_sql_postgresql/src/types.rs:175`) derives from `canonical_read_type`, is applied on both the DDL side (`catalog.rs` domain, `catalog_advanced.rs` composite `fields`) and the live side (`domain_semantics`, `composite_semantics`), and returns an error for types with no single static Sifr path. The residual `'sifr_type','str'` / `jsonb_object_agg(attname,'str')` literals in `postgresql_catalog.sql` are unconditionally overwritten by normalization.
- `enum_semantics` now supplies `variants` on both paths, which `generate_profile_module` requires — previously no DDL-sourced domain/composite/enum could produce a generated module at all, so the added strictness is a net widening, not a narrowing.
- Remediation hygiene items are present and correct: `emit(&stdout)` before any artifact write in `command.rs:80`, `--`-prefixed profile rejection at `command.rs:287`, `SNAPSHOT_PATH` reuse, PG13-safe multirange split (`postgresql_multirange.sql` only queried at major ≥ 14), and the removed empty-semantics guard is superseded by per-kind required-field checks (namespaces/sequences are now legitimately empty).
- Live test coverage matches the claim: reload + `build_schema_artifacts` on every major, `value: i32` / `unit_count: i32` / `latitude: sifr.sql.Numeric` assertions, and `semantic_diff(ddl, live).is_empty()` on 18.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr_sql_postgresql/src/catalog.rs:845` (`last_segment`) keys catalog functions by the final identity segment, but live function identities are `<ns>.<name>._overload_<md5>` (`postgresql_catalog.sql`). A pulled user function is therefore registered under `_overload_<md5>` and `PostgresCatalog::functions(["add_one"])` never finds it. Reload succeeds silently, so the live test does not catch it. Same identity-shape divergence applies to `_operator._overload_*` and `_cast._overload_*` versus the DDL's `<ns>.<name>`.
- new mechanism-level defect: `normalization.rs` `function_semantics` sets `result-nullable` to `!strict`, while the DDL path (`catalog.rs`) always sets `true`. Besides being a vocabulary divergence of exactly the finding‑1 class, the direction is unsound — a PostgreSQL `STRICT` function returns NULL for NULL input, so marking its result non-nullable is the unsafe reading, not the safe one.
- new mechanism-level defect: live validation still cannot reach parity for definition-valued objects. Views emit `provider-query` as `pg_get_viewdef` text while DDL emits the serialized query AST; check constraints emit `pg_get_constraintdef` text while DDL emits the serialized expression AST. Combined with server-assigned constraint names (`t_col_key`, `t_col_fkey`, `t_col_check`) versus the DDL's synthesized `t_unique_1`/`t_fkey_1`/`t_check_1`, and live-only `identity-column` objects, an `Introspection`-evidence profile validating an in-sync database still reports differences for any schema containing a view, check constraint, non-PK constraint, or identity column. The PG18 parity test only exercises namespace + table + columns + PK, so this is unmeasured.
- suggestion: `register_domains` runs before `register_ranges`, so a domain whose base is a user-defined range type fails closed with "missing required domain base type" even though the metadata is complete; registering ranges first (or folding ranges into the same fixpoint loop) removes the false failure.
- suggestion: a domain over an array (`CREATE DOMAIN d AS text[]`) or a composite with an array attribute now fails DDL normalization with "cannot be represented by one static Sifr type path", and aborts an entire live pull. This is consistent with the documented fail-closed policy and is not a regression (module generation already rejected these), but arrays are common enough in PostgreSQL that a representable generated form is worth scheduling.
- suggestion: the dead `'sifr_type','str'` and `jsonb_object_agg(a.attname,'str')` expressions in `postgresql_catalog.sql` should be dropped; they are the exact fabrication finding 2 removed, and they will silently resurface if the normalization pass is ever bypassed.

yaseralnajjar:
Exact-SHA gate evidence for final candidate `ba3629d634b15c192af46b66a4972c77683fd194`:

- `scripts/run_all_tests.sh --profile create-pr`: stopped before tests because the offline profile omits required SQL platform suites `host-tools`, `postgresql-live-differential`, `postgresql-live-runtime`, `postgresql-live-schema-tools`, and `schema-tools`. Per the item rule, it was not rerun.
- `scripts/run_all_tests.sh`: stopped before tests for the same known profile-model contradiction. Per the item rule, it was not rerun.

The inherited contradiction is already assigned to M18. M12 focused validation, strict linting, guardrails, and the PostgreSQL 13–18 live schema-tool matrix passed; the exact-SHA remediation review is satisfied.

PR #3622, opened 2026-08-31
Emit lazy resumable generators

yaseralnajjar:
## Summary
- replace eager Vec-backed sync and async generator lowering with demand-driven resumable futures
- make itertools generators lazy, keep count unbounded, and validate islice at construction
- add native suspension, side-effect, partial-consumption, infinite-source, async exhaustion, and differential coverage
- regenerate all affected authoritative demo companions

## Validation
- 1,183 sifr_codegen tests passed
- 1,084 sifr_lowering tests passed (1 ignored)
- workspace Clippy and cargo fmt passed
- full 91-project generated corpus, panic scan, determinism, demos, and freshness passed
- all 91 individual rustfmt and Clippy classifications passed; aggregate records remain Item 8-owned
- native generator, async-generator, CPython itertools, and consolidated itertools regressions passed
- file-size, HIR maintainability, audit inventory, companion freshness, and diff hygiene passed

## Deferred existing failures
- Item 6: deterministic JSON map-order assertions in parsers_and_encoders and structured_data_formats
- Item 8: sliding_window_narrowing optional-index build failure, stale E2E surface inventory, and aggregate generated rustfmt debt signature
- Item 12: qualification-profile composition/gate reconciliation


yaseralnajjar:
Exact-SHA Opus review for candidate `1029541fd69c9b1d6726f53331cc5319f17f3be3` (base `c560ae5da1e85f433df7e5bb6a034cd8e75c1f54`):

## Verdict: NOT SATISFIED

The lazy-generator mechanism itself is sound and the item's core criteria are met; one regression in the item's own return-lowering rewrite blocks acceptance.

**What I verified as correct (mechanism, not summary):**
- `crates/sifr_codegen/src/preamble/generator_runtime.rs` implements a genuine one-suspension-per-`next()` state machine: `__SifrYieldFuture::poll` stores the value and returns `Pending` on first poll, `Ready` on resume; `__SifrGenerator::next` polls the pinned producer exactly once with `Waker::noop()`, takes the slot, and latches `complete`/drops the producer only on `Ready`. No eager `_yields` vector remains anywhere (`generator_bodies.rs` eager path deleted; `!contains("_yields")` asserted in codegen tests). Criterion 1 met.
- `AsyncGenerator::anext` correctly distinguishes a yield-suspension (`Pending` + slot occupied → `Ready(Ok(Some))`) from a real await (`Pending` + empty slot → `Pending`, real waker propagated downward), closes on `Ok(None)`/`Err`, and `aclose` drops the producer. Lazy-start and close-before-start are covered by the pre-existing `async_generator_aclose_before_start.sifr`.
- Panic-freedom: no `unwrap`/`expect`; mutex poisoning handled via `into_inner()` on both store and take. The two `panic!`s in `resumable_generator_bodies.rs` are unreachable programmer invariants — `function_flow.rs:147-168` normalizes/rejects any generator whose annotation is not `Iterator[T]`/`AsyncGenerator[T, E]` before codegen.
- Criterion 2: `count()` is `while True: yield`, and `generator_resumption_and_lazy_itertools.sifr` drives 10,001 `next()` calls plus `islice(iter(count(50, 3)), 10000, 10005)` — an eager implementation could not pass either. Criterion 4's laziness/partial-consumption/side-effect ordering is covered by the same fixture's filesystem-marker generator.
- Criterion 3: `islice` validates negative bounds and `step <= 0` eagerly in the non-generator wrapper returning `Result[Iterator[T], ValueError]`, so error timing is at construction, not first pull; `_islice_impl` is the lazy generator. Adapter ports (`accumulate`, `compress`, `takewhile`, `cycle`, `permutations`, `combinations*`, `product`) are constant-state lazy loops, and the CPython-parity assertions were converted, not weakened.
- Audit anchors for ERQ-015/016/033 flip to `revision: "baseline"`, matching the convention already used by closed items 1/3/4.

## Blocking findings

**1. Generator `return <None-typed expression>` silently discards the expression and its side effects.**
- File/symbol: `crates/sifr_lowering/src/lower/return_lowering.rs:20-27` (`lower_return`, new `ctx.current_function_is_generator` branch).
- Mechanism: the branch lowers `val` via `lower_expr`, then, when `expr_ty.resolve_alias()` is `Type::None`, returns `HirStmt::Return { value: None }` — the lowered expression is dropped on the floor. It is never re-emitted as an `HirStmt::Expr`, so codegen produces a bare exhaustion (`return;` / `return Ok(())`) with the call erased.
- Failure scenario: `def note() -> None: ...` plus `def gen() -> Iterator[int]:\n    yield 1\n    return note()` compiles cleanly and never calls `note()`. Before this diff the equivalent program was a hard `SIFR-TYPE-0002` (async generators) or a return-type mismatch (sync generators), so this converts a rejected program into a silently miscompiled one. Any None-returning side-effecting call (`note()`, a mutating helper) in generator return position is lost.
- Violated criterion: 4 (side effects), and the item's own stated contract "bare return and return None exhaust; non-None returns rejected" — a None-*typed call* is neither.
- Correction: treat only `HirExpr::NoneLiteral` (and bare `return`) as exhaustion. For any other None-typed expression, emit the lowered expression as a preceding `HirStmt::Expr` before `HirStmt::Return { value: None }` (or reject it with a dedicated diagnostic). Add a native regression asserting the side effect is observed after the final `next()`.

## Follow-up findings

- **`crates/sifr_codegen/src/stmt_support_emitter/stmt_block.rs:634-637`**: `HirStmt::Yield` is lowered to `__sifr_yielder.suspend(..).await` unconditionally, with no `in_generator_closure` guard — unlike `with_yield_and_match.rs:151`, which now returns `None` outside a generator closure. Nested function lowering pins `in_generator_closure: false` (`simple_dispatch_and_bindings.rs:552`), so a nested generator that falls through to this path would emit an undefined `__sifr_yielder`. This mirrors the previous dangling-`_yields` behavior exactly, so it is not caused by this diff, but the new mechanism makes a guard or explicit unsupported-nested-generator diagnostic cheap and worthwhile. Owner: whoever next touches nested-function lowering.
- **`crates/sifr/tests/e2e/fail/islice_non_iterable_input.sifr`**: rewritten from `islice(42, 2)` to `islice(iter(42), 2)`. The fixture now exercises `iter()` on a non-iterable rather than a non-iterable reaching `islice`'s `own data: Iterator[T]` parameter. Consider keeping both forms so the parameter-level rejection stays pinned. Owner: stdlib-parity fixture maintenance.
- **CPython divergence worth recording in the parity notes** (not introduced defects, but newly relevant): itertools adapters now require `Iterator[T]` and reject bare lists/iterables, so every call site needs an explicit `iter(...)`; and `islice(it, start, None)` still means "stop = start_or_stop" rather than CPython's unbounded tail. Both are deliberate ownership/signature adaptations, but they are the kind of thing the audit inventory tracks.
- **`stdlib/sifr/itertools.sifr` `cycle`**: the `emitted >= n` check happens at the top of the source loop, so one extra element is pulled from the source before returning when `n` is reached mid-stream. Unobservable for owned sources; noted only for parity precision.
- **Cosmetic, from a bulk `iter(...)` rewrite**: two mangled comments — `crates/sifr/tests/e2e/pass/cpython_itertools.sifr` "`# accumulate (iter(supported signature: optional initial))`" and a matching one in the generic-accumulate section of the consolidated fixture.

yaseralnajjar:
Sole remediation Opus review for candidate `cc63e5d4e86725543ed111b3c194d2e89ab5e629` (phase base `c560ae5da1e85f433df7e5bb6a034cd8e75c1f54`):

I inspected the exact candidate `cc63e5d4e` and the remediation diff `1029541fd..cc63e5d4e` directly (no re-validation runs).

## Verdict: SATISFIED

### Blocking findings

None.

The initial blocker is fully resolved at the root. `crates/sifr_lowering/src/lower/return_lowering.rs:24-41` now splits the None-typed generator-return path: only `HirExpr::NoneLiteral` short-circuits to `HirStmt::Return { value: None }`; every other None-typed expression gets `DiagnosticCode::TYPE_MISMATCH` (SIFR-TYPE-0002) at `val.range()` before the value is dropped. `error_with_code_at` (`mod_context.rs:469`) pushes a hard error, so such a program cannot reach codegen — the discard is now unreachable rather than silent. `lower_return` is the single return-lowering path (`statements/statement_dispatch.rs:167`), and `current_function_is_generator` is set for free functions, methods, and async generators alike, so the guard covers generator methods too. `return await f()`, `return note()`, and `return x` where `x: None` are all rejected; a genuine `return None`/bare `return` is still exhaustion.

The cycle boundary is corrected: `stdlib/sifr/itertools.sifr:455-467` returns before any pull when `n <= 0` and returns immediately after the nth yield rather than at the top of the next iteration. Traced against every existing expectation — `cycle([1,2,3],10) == [1,2,3,1,2,3,1,2,3,1]`, `cycle(...,0) == []`, `cycle([5,6],5)`, `cycle([1,2,3,4],6)` — all unchanged. The new `cycle_source` fixture proves it: after `list(cycle(cycle_source(path), 2))` the marker is `cycle-second`, so the third source effect never runs.

Criteria coverage holds: laziness (`generator_resumption_and_lazy_itertools.sifr:36` asserts no side effect before the first `next`), infinite-source termination plus a memory proxy (10,001 pulls from `count()`, then `islice(count(50,3), 10000, 10005)`), eager `islice` validation with lazy consumption (`itertools.sifr:99-109` validates in a plain `Result`-returning function delegating to `_islice_impl`, pinned by the construction-site `ValueError` assertions in `cpython_itertools.sifr:80-92`), and the new compile-fail fixture at `col=12`, which is the correct column for `note()`.

The two mangled comments were introduced by `1029541fd` and are exactly reverted; a net `c560ae5da..cc63e5d4e` scan of `.sifr` comment lines shows no other residual corruption. The four regenerated companions contain only the cycle boundary movement.

### Follow-up findings (non-blocking)

- **Archived record references deleted fixtures** — `plans/reviews/archive/phase32_async_generator_return_deferred_fixtures.md` still describes `async_generator_bare_return_deferred.sifr` and `async_generator_return_none_deferred.sifr`, which this phase deleted when bare/`None` async-generator returns became legal exhaustion. Archives are historical snapshots, so no correction is required; owner is the phase records maintainer if a pointer to the superseding behavior is wanted.
- **`cycle` repeat loop advances `emitted` on a `None` slot** — `stdlib/sifr/itertools.sifr:468-474` increments `emitted` unconditionally while yielding only when `saved[index] is not None`, and the first loop likewise saves-but-does-not-count a `None` value. This is only reachable when `T` is itself instantiated as an optional type, and it is pre-existing (untouched by this phase in that respect). Assign to a later item — the stdlib itertools generic-optional semantics owner, not Item 5.
- **Raw rustfmt debt in the regenerated companions** is pre-existing and Item 8-owned, as stated.

yaseralnajjar:
Gate evidence on exact reviewed candidate `cc63e5d4e86725543ed111b3c194d2e89ab5e629`:

- Sole create-PR gate: stopped at preflight because profile `create-pr` omits required SQL suites `host-tools`, `postgresql-live-differential`, and `postgresql-live-runtime`.
- Sole merge gate: stopped at the identical preflight because profile `merge` omits the same required SQL suites.
- Neither gate was rerun. This profile-composition failure is assigned to Emitted Rust Excellence Item 12.

PR #3625, opened 2026-08-31
SQL M13: add migration compiler and engine

yaseralnajjar:
## Summary

- add checked migration DAG compilation with stable identities, schema fingerprints, reflected DDL, typed data steps, assertions, backfills, transactions, impact reports, and explicit rollback
- add affine nominal migration plan/state HIR plus scoped MigrationDb contracts
- add provider-neutral execution and recovery records with deterministic atomic migration artifacts
- add migration qualification, mutation coverage, focused tests, and internal architecture documentation

## Validation

- `python3 verification/areas/sql_platform/runner.py --suite migration-engine` (6/6 variants)
- `cargo test -p sifr_sql_contract -p sifr_sql_runtime -p sifr_sql_tool -p sifr_frontend`
- strict Clippy for affected libraries and all new test targets
- HIR maintainability guardrails
- first-party file-size guardrails
- SQL platform contracts and mutation self-test

Candidate: `8e51ecb0bd56be522e9b7530feae7566c0f8b8ff`


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_sql_runtime/Cargo.toml:11` (with `crates/sifr_sql_runtime/src/migration/mod.rs:5-7` and `src/migration/engine.rs:237-240`), criterion 11 / owned scope "compiler-only migration representation vs. application runtime dependencies". The application-runtime crate now takes a hard dependency on the compiler-only contract crate. `internal_docs/sql_common_contracts.md:16-17` states the invariant this change breaks: "The compiler emits closed runtime metadata; the runtime never links the compiler-only contract crate." Before this diff `sifr_sql_runtime`'s normal dependencies were exactly `sifr_runtime` + `tokio`; `cargo tree -p sifr_sql_runtime -e normal` now shows `sifr_sql_contract → sifr_compiler_component → wasmtime v48.0.1` inside the application runtime closure, and no other runtime crate (including `sifr_sql_postgresql_runtime`) does this. The engine consumes `CompiledMigrationGraph`/`CompiledMigration*`/`CompiledStepKind` directly rather than a closed emitted runtime record, so the compiler-side representation is the leak vector, not an incidental import. That documented invariant was also left untouched while the surrounding docs were updated, so the architecture record now contradicts the code. Correction: either emit a closed, runtime-owned migration plan record (matching the documented "compiler emits closed runtime metadata" contract) and keep `MigrationEngine` on that type, or relocate the engine to a host/tool crate that is allowed to link `sifr_sql_contract` (e.g. alongside `sifr_sql_tool`'s migration artifacts); if the boundary is intentionally being redefined, `sql_common_contracts.md` must be updated in the same change with the justification and the `wasmtime` closure consequence stated.

Follow-up findings:
- suggestion: the compiler compiles per-parent paths and multi-parent merges (`compiler.rs:84-136`), but `select_path` (`engine.rs:711-744`) treats any fork as `AmbiguousPath` — two siblings off a common head necessarily share `input_fingerprint` and both satisfy `parent_is_current`, so no branching graph the compiler accepts can ever be executed. Tests cover only linear graphs (`migration_contracts.rs:489` `topological_order_property_is_stable_for_linear_graphs`, `migration_engine.rs` single-migration graphs). Worth either a deterministic topological tie-break or an explicit compiler rejection of graphs the engine cannot run.
- suggestion: `MigrationDefinition` carries one shared `steps` list for all parents, and merges are rejected unless every parent path yields the same schema (`compiler.rs:121-133`), so a merge is only expressible when the parent schemas already agree. That is a narrower merge model than criterion 1's DAG wording implies; worth recording in `sql_migrations.md`.
- suggestion: explicit reverse plans are compiled and stored (`CompiledMigrationPath.rollback`), but `MigrationEngine` has no entry point that executes them — it only calls `runtime.rollback_transaction()`. Criterion 10 is met literally (nothing destructive is synthesized), but the compiled rollback is currently inert.
- suggestion: a crash while an explicit transaction is open leaves a ledger record with `transaction_open: true`, which both `validate_progress` (`engine.rs:751-761`) and its replay check (`engine.rs:802-807`) reject permanently. That is fail-closed rather than ambiguous, but it is an unrecoverable-without-operator-edit state worth documenting.
- suggestion: the whole surface (`MigrationCompiler`, `lower_migration_graph`, `MigrationEngine`, `build_migration_artifacts`) has no caller outside tests — nothing in `sifr_driver`/`sifr_codegen` constructs a `MigrationGraphDefinition`, and `DataCallbackContract.is_async`/`nonescaping` are trusted input booleans rather than facts derived from Sifr source. Criteria 3–5 hold at the contract/HIR level as the milestone doc scopes them, but the developer-facing path from `.sifr` source to a checked plan does not exist yet; if Milestone 14 does not own that wiring, it should be assigned explicitly.

yaseralnajjar:
Remediation candidate: `c6acf3da7518c2c67a844036c183f45e91440a15`

- replaced the runtime dependency on `sifr_sql_contract` with a closed runtime-owned `MigrationExecutionPlan`
- moved compiler-to-runtime lowering and artifact emission into `sifr_sql_tool`
- added a qualification guard that rejects compiler-contract linkage from the application runtime
- added stable topological branch/merge execution coverage
- documented equivalent-schema merge semantics, fail-closed transaction recovery, and later ownership for live reverse execution and source command wiring

Validation on this candidate: migration suite 6/6, all affected crate tests, workspace Clippy, strict affected all-target Clippy, formatting, HIR/file-size guards, SQL contracts and self-test, locked metadata, and runtime dependency-tree inspection all pass.

yaseralnajjar:
I inspected the diff between the initially reviewed SHA and the final candidate, plus the surrounding compiler/runtime/tool boundary.

**Prior blocker — corrected.** `crates/sifr_sql_runtime/Cargo.toml` no longer lists `sifr_sql_contract`; `cargo tree -p sifr_sql_runtime -e normal` resolves to `serde`, `sifr_runtime`, `tokio` only — no `sifr_compiler_component`, no Wasmtime. The closed types in `crates/sifr_sql_runtime/src/migration/plan.rs` are self-contained and serde round-trip, the dependency direction is now `sifr_sql_tool -> sifr_sql_runtime` via `lower_migration_execution_plan`, `graph.json` serializes the closed plan (asserted in `crates/sifr_sql_tool/tests/migration_artifacts.rs:59`), and the manifest guard in `verification/areas/sql_platform/tools/check_migration_engine.py:88` fails closed on reintroduction.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- mechanism-level defect (owned scope, present since the initial candidate, not introduced by remediation): `crates/sifr_sql_runtime/src/migration/engine.rs:194` removes every parent from `ledger.heads` without requiring that every parent was applied, and `select_path` at `engine.rs:481` matches a merge path as soon as one parent is the current head with a matching fingerprint. For a diamond whose branches change the schema (baseline `f0` -> `left`/`right` both producing `f_s` -> `merge`), applying `left` moves the live fingerprint to `f_s`; `right`'s only path still declares input `f0`, so it is never selectable, while `merge`'s path from `left` is. The engine applies `merge`, drops `right` from heads, and reports `Complete` with `right` absent from `applied`. The remediation's claim that the branching follow-up is resolved holds only for the schema-neutral branches used in `branching_plan_uses_stable_topological_order_and_reaches_merge_head` (`crates/sifr_sql_runtime/tests/migration_engine.rs:472`), where both branch paths keep `output_fingerprint == input_fingerprint`. The compiler accepts the schema-changing form (`crates/sifr_sql_contract/src/migration/compiler.rs:119` only requires equal merge outputs per parent), so an accepted graph is still not fully executable — the failure is now silent rather than an `AmbiguousPath` error. Record for a later item: either the compiler must reject sibling branches whose paths cannot be sequenced, or the engine must refuse a merge whose parents are not all applied.
- suggestion: the closed plan drops per-migration `provider` constraints (`minimum_server_version`, `required_capabilities`), `transaction_requirement`, `author`, and `created_at`, which `graph.json` previously carried. Family is still checked via `MigrationExecutionPlan::provider_family`, and the rest is compile-time-checked, so no Milestone 13 criterion is unmet — but Milestone 14's live server-major matrix will need these at execution time and will require a `format_version` bump to carry them.
- suggestion: `MigrationExecutionErrorKind::AmbiguousPath` is now only raised for the *no matching path* case (`engine.rs:501`); the multi-candidate meaning is gone. The name no longer describes the condition it reports.

yaseralnajjar:
Exact-SHA gate evidence for final candidate `c6acf3da7518c2c67a844036c183f45e91440a15`:

- One create-PR gate: `scripts/run_all_tests.sh --profile create-pr` — stopped before tests because the profile omits required SQL suites: `host-tools`, `migration-engine`, `postgresql-live-differential`, `postgresql-live-runtime`, `postgresql-live-schema-tools`, and `schema-tools`.
- One merge gate: `scripts/run_all_tests.sh` — stopped before tests on the same required-suite/profile contradiction.

Neither gate was rerun. The permanent `migration-engine` suite passed 6/6 variants directly on this SHA, all affected crate tests passed, and workspace Clippy plus all documented guards passed.

PR #3627, opened 2026-08-31
feat(sql): qualify PostgreSQL migrations

yaseralnajjar:
## Summary

- add PostgreSQL DDL reflection, closed transaction classification, and tamper-resistant operator plans
- add advisory-locked PostgreSQL migration ledger, truthful import, apply/resume, and explicit rollback
- qualify fresh, upgrade, merge, interruption, failure, drift, and rollback on PostgreSQL 13 through 18
- document crate ownership and exact locked Rust tooling

## Validation

- SQL platform migration-engine and postgresql-migrations suites: 9/9 passed
- PostgreSQL migration live matrix: majors 13-18 passed
- cargo fmt --check

## Phase

Milestone 14: PostgreSQL migration qualification


yaseralnajjar:
## Initial exact-SHA Opus review

Candidate: `e3d90614fe3149b1f49e24f1743089c1b9a3af55`

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr_sql_runtime/src/migration/engine.rs:89` — `execute_locked` returns `Complete` whenever `heads == {graph.head}` and the fingerprint matches, without inspecting `ledger.in_progress`. Rollback (new in this item) can now leave a `direction: Rollback` progress record with an unchanged fingerprint (a reverse path that failed on its first step), and a following `apply` reports `Complete` while that record is still live. State stays consistent and `rollback` resumes it, but the report omits the pending reversal.
- suggestion: `crates/sifr_sql_postgresql/src/migration.rs:268` — the "PostgreSQL DDL replaces an existing schema identity" guard is unreachable: `ddl_document` (`crates/sifr_sql_postgresql/src/catalog.rs:490`) filters out every identity already in `prior_objects` before returning, so the insert never collides. The documented rule (`internal_docs/sql_postgresql_migrations.md:67`) is actually enforced by falling through to `Opaque`, which then demands a declared effect. Behavior is fail-closed; the branch is dead code that reads as the enforcement point.
- suggestion: `crates/sifr_sql_postgresql/src/migration.rs:325` — `leading_sql_tokens` treats quotes as skippable punctuation, so string-literal contents become keywords. `CREATE TABLE t (c text DEFAULT 'CREATE INDEX CONCURRENTLY')` classifies as `RequiresAutocommit` within the 12-token window. Over-restrictive rather than unsafe, but it rejects a legal transactional statement at plan validation.
- suggestion: `crates/sifr_sql_postgresql/src/migration.rs:291` — the non-transactional inventory omits `ALTER TABLE … DETACH PARTITION CONCURRENTLY` (PostgreSQL 14+) and the `ALTER TYPE … ADD VALUE` in-transaction restriction. Those plan as transactional and fail at the server instead of at `validate_postgres_migration_plan`, so the operator learns late.
- suggestion: `verification/areas/sql_platform/tools/check_postgresql_migrations.py:45` — `reflected_object_kinds` and `declared_effect_object_kinds` are checked only by cardinality (16/12) and disjointness. Moving a kind between the two lists, or renaming one, still passes. Cross-checking against `SchemaObjectKind` or the capability matrix would make the criterion-1 inventory self-verifying rather than a hand-maintained count; the Rust test at `crates/sifr_sql_postgresql_tools/tests/migration_qualification.rs:74` is what actually pins coverage today.
- suggestion: `crates/sifr_sql_tool/src/migration_artifacts.rs:63` — lowering now stamps the constant `MIGRATION_EXECUTION_PLAN_FORMAT_VERSION` and drops `graph.format_version` entirely, so a compiled graph carrying an unexpected contract format is relabeled as 2 instead of rejected.
- suggestion: `crates/sifr_sql_postgresql_tools/src/postgresql_catalog.sql:5` — excluding `sifr_internal` applies to every catalog pull, not just migrations, so a user schema with that name silently disappears from the schema authority. The reservation is recorded only in `internal_docs/sql_postgresql_migrations.md:110`; `internal_docs/sql_schema_tools.md` does not mention it.
- infrastructure issue: `crates/sifr_sql_postgresql_tools/src/migration_runtime.rs:197` — ledger bootstrap runs `CREATE SCHEMA IF NOT EXISTS` / `CREATE TABLE IF NOT EXISTS` under a profile-scoped advisory key. Two different profiles bootstrapping the same database concurrently hold different keys, and PostgreSQL's `IF NOT EXISTS` is not race-free, so the first-run bootstrap can fail with a duplicate-object error. Retry recovers; the failure surfaces as "cannot initialize PostgreSQL migration ledger".

Criterion notes supporting the verdict: reflection reuses the schema-authority `ddl_document` and is pinned by an object-kind test (1); `validate_steps` enforces required/forbidden boundaries and a preceding named recovery point for autocommit DDL, and the runtime refuses non-transactional DDL inside an open transaction (2); lock, concurrent-start, checksum, head, provider-identity, incomplete-merge and schema drift all fail closed with distinct kinds, and the four Milestone 13 carry-overs at `plans/issues/active/ad-hoc-schema-first-sql-platform.md:1347-1350` are each addressed (3); `import_baseline` writes one head with an empty `applied` map and refuses a fingerprint mismatch or an existing row (4); `plan` emits per-step forward and reverse actions plus separated forward-only/reversible lists, and `rollback_last` refuses to synthesize a reverse path (5); the live contract covers create, upgrade, merge, bounded-backfill pause, resume, assertion failure with transaction rollback, and explicit rollback to recorded prior heads across majors 13–18 (6).

yaseralnajjar:
## Remediation exact-SHA Opus review

Candidate: `a8f02da6258f2938697cc696e316467c8a970902`

Verdict: SATISFIED

Blocking findings:
none

Verification notes (read-only, `d65ce11..a8f02da`):
- **Forward/rollback direction guard** — `crates/sifr_sql_runtime/src/migration/engine.rs:85` rejects a `Rollback` `in_progress` record before the head/fingerprint short-circuit, so the reported "reports Complete while rollback progress sits at the target head" path now returns `AmbiguousRecovery`. It mirrors the pre-existing symmetric guard in `rollback.rs:76`, and resume of an interrupted rollback still flows through `rollback_last`, so nothing legitimate is blocked. Regression test at `tests/migration_engine.rs:614` reproduces the exact original defect.
- **Quoted-literal scanning** — `skip_quoted`/`skip_dollar_quoted` (`crates/sifr_sql_postgresql/src/migration.rs:215,229`) run before whitespace/punctuation handling; `--` and `/* */` skipping is unaffected because neither starts with a quote byte. Unterminated literals terminate the scan safely. Tests cover both a single-quoted and a dollar-quoted `CREATE INDEX CONCURRENTLY` default.
- **New autocommit families** — `detach_partition_concurrently` slices `tokens[index + 2..]`, which is in-bounds whenever `tokens.get(index + 1)` succeeded; `add_enum_value` requires the adjacent `ADD VALUE` pair, so `ALTER TYPE ... RENAME VALUE` and `ADD ATTRIBUTE value` stay transactional. Both statements are pinned in `migration_qualification.rs:142`.
- **Lowering rejection** — `lower_migration_execution_plan` now returns `Result`; the only non-test caller is `build_once` (`migration_artifacts.rs:151`), and the re-export plus the `check_migration_engine.py` token requirement still resolve. No stale call site.
- **Bootstrap lock** — acquired only after the non-blocking profile lock, always released on both the success and setup-failure paths (`setup.err().or_else(bootstrap_release.err())`), and the profile lock is released on every new error path. Lock ordering is acyclic, so no deadlock is introduced. The live test starts two profiles behind a `Barrier`.
- **Dead reflection branch** — `ddl_document` filters `prior_identities` out of its result (`catalog.rs:501`), so the removed collision branch was genuinely unreachable and the replacement doc sentence ("a statement that does not add a new reflected object becomes opaque") is accurate.
- **Checker pinning** — I ran the contract sets against the JSON: reflected (16), declared-effect (12), and non-transactional (16) all compare equal, so the exact-set assertions are consistent with the shipped data.

Follow-up findings:
- pre-existing issue: `skip_quoted` (`crates/sifr_sql_postgresql/src/migration.rs:215`) only recognizes doubled-quote escaping. A backslash-escaped `E'it\'s'` literal ends the skip early and the remaining literal bytes re-enter keyword scanning, which can yield a false `RequiresAutocommit` (a statement run outside a transaction). The remediation narrows but does not close the original class.
- suggestion: the ledger bootstrap uses blocking `pg_advisory_lock` with no timeout while holding the profile lock (`migration_runtime.rs:197`). A peer stalled inside ledger creation blocks every other profile's migration start with no operator-visible diagnostic; a `lock_timeout`-bounded acquire with a distinct message would fail closed with better reporting.
- suggestion: the autocommit family list is maintained in three independent places — the classifier, the Rust qualification list (`migration_qualification.rs:139`), and `postgresql_migration_qualification.json`. The checker pins the JSON set but never cross-checks it against the classifier, so a family added to only two of the three still passes.
- suggestion: `live_scenarios` is still validated by cardinality (`check_postgresql_migrations.py:84`) while the object inventories are now exact; renaming a scenario would pass.
- suggestion: the shared `recovery_reason` string "PostgreSQL requires this DDL outside a transaction" is inaccurate for `ALTER TYPE ... ADD VALUE`, which PostgreSQL permits inside a transaction but whose value cannot be used until commit. A per-family reason would report the true operator constraint.

No files were modified.

yaseralnajjar:
## Final gate evidence

Exact candidate: `a8f02da6258f2938697cc696e316467c8a970902`

- Create-PR gate: invoked exactly once; stopped before tests because the `create-pr` profile omits required SQL platform suites: `host-tools`, `migration-engine`, `postgresql-live-differential`, `postgresql-live-migrations`, `postgresql-live-runtime`, `postgresql-live-schema-tools`, `postgresql-migrations`, and `schema-tools`.
- Merge gate: invoked exactly once; stopped before tests because the `merge` profile omits the same required SQL platform suites.
- Per the phase rule, neither gate was rerun.
- Candidate validation completed before gates: registered migration-engine and PostgreSQL-migrations suites (9/9), pinned live PostgreSQL 13–18 matrix including cross-profile bootstrap, workspace Clippy with warnings denied, strict Clippy for new migration test targets, formatting, HIR maintainability, 900-line source guard, qualification checks, and mutation checks.
- Remediation exact-SHA Opus verdict: SATISFIED with no blocking findings.

PR #3629, opened 2026-08-31
feat(codegen): close stdlib emitted semantics

yaseralnajjar:
Closes Item 6 of the Emitted Rust Excellence phase.\n\nImplements full stdlib emitted-semantics closure: checked string padding, signed-zero exact division, ordered JSON, sized/seekable I/O with error-kind preservation, iterator optional-value and islice parity, demand-driven generic bounds, and generated regressions/companions.\n\nValidation and exact-SHA Opus evidence will be added before readiness.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,7 +146,7 @@ rust_decimal = { version = "1.42.1", features = ["maths", "serde-with-str"] }
 rustc-hash = { version = "2.1.3" }
 schemars = { version = "1.2.2" }
 serde = { version = "1.0.229", features = ["derive"] }
-serde_json = "1.0.151"
+serde_json = { version = "1.0.151", features = ["preserve_order"] }
 semver = { version = "1.0.28", features = ["serde"] }
 serde_test = { version = "1.0.177" }
 sha1 = { version = "0.11.0" }
--- a/crates/sifr/tests/e2e/pass/cpython_itertools.sifr
+++ b/crates/sifr/tests/e2e/pass/cpython_itertools.sifr
@@ -17,9 +17,21 @@ def is_even(x: int) -> bool:
 def add2(a: int, b: int) -> int:
     return a + b
 
-def collect_islice(own data: Iterator[int], start_or_stop: int, stop: int | None = None, step: int = 1) -> list[int]:
+def optional_int(value: int) -> int | None:
+    return value
+
+def collect_islice(own data: Iterator[int], stop: int) -> list[int]:
     try:
-        sliced: Iterator[int] = islice(iter(data), start_or_stop, stop, step)
+        sliced: Iterator[int] = islice(iter(data), stop)
+        return list(sliced)
+    except ValueError as e:
+        _message: str = e.message
+        assert_true(False)
+        return []
+
+def collect_islice_range(own data: Iterator[int], start: int, stop: int | None, step: int | None = 1) -> list[int]:
+    try:
+        sliced: Iterator[int] = islice(iter(data), start, stop, step)
         return list(sliced)
     except ValueError as e:
         _message: str = e.message
@@ -74,9 +86,11 @@ def main():
 
     # islice validates invalid arguments at construction and consumes its source lazily.
     assert_true(str(collect_islice(iter([0, 1, 2, 3, 4]), 3)) == "[0, 1, 2]")
-    assert_true(str(collect_islice(iter([0, 1, 2, 3, 4]), 1, 4)) == "[1, 2, 3]")
-    assert_true(str(collect_islice(iter([0, 1, 2, 3, 4]), 1, 5, 2)) == "[1, 3]")
-    assert_true(str(collect_islice(iter([0, 1, 2, 3, 4]), 10, 110, 3)) == "[]")
+    assert_true(str(collect_islice_range(iter([0, 1, 2, 3, 4]), 1, 4)) == "[1, 2, 3]")
+    assert_true(str(collect_islice_range(iter([0, 1, 2, 3, 4]), 1, 5, 2)) == "[1, 3]")
+    assert_true(str(collect_islice_range(iter([0, 1, 2, 3, 4]), 10, 110, 3)) == "[]")
+    assert_true(str(collect_islice_range(iter([0, 1, 2, 3, 4]), 2, None)) == "[2, 3, 4]")
+    assert_true(str(collect_islice_range(iter([0, 1, 2, 3, 4]), 2, None, None)) == "[2, 3, 4]")
     invalid_zero_step: bool = False
     try:
         _zero_step: Iterator[int] = islice(iter([10, 20, 30]), 0, 3, 0)
@@ -143,6 +157,17 @@ def main():
     # cycle (intentional finite signature adaptation: cycle(data, n))
     empty_cycle: list[int] = []
     assert_true(str(list(cycle(iter([1, 2, 3]), 10))) == "[1, 2, 3, 1, 2, 3, 1, 2, 3, 1]")
+    optional_cycle: list[int | None] = []
+    optional_cycle.append(optional_int(1))
+    optional_cycle.append(None)
+    optional_cycle.append(optional_int(2))
+    optional_cycle_expected: list[int | None] = []
+    optional_cycle_expected.append(optional_int(1))
+    optional_cycle_expected.append(None)
+    optional_cycle_expected.append(optional_int(2))
+    optional_cycle_expected.append(optional_int(1))
+    optional_cycle_expected.append(None)
+    assert_true(list(cycle(iter(optional_cycle), 5)) == optional_cycle_expected)
     assert_true(str(list(cycle(iter(empty_cycle), 10))) == "[]")
     assert_true(str(list(cycle(iter([1, 2, 3]), 0))) == "[]")
 
--- a/crates/sifr/tests/e2e/pass/cpython_itertools_subset.sifr
+++ b/crates/sifr/tests/e2e/pass/cpython_itertools_subset.sifr
@@ -8,9 +8,19 @@ def lt3(x: int) -> bool:
 def add2(a: int, b: int) -> int:
     return a + b
 
-def collect_islice(own data: Iterator[int], start_or_stop: int, stop: int | None = None, step: int = 1) -> list[int]:
+def optional_int(value: int) -> int | None:
+    return value
+
+def collect_islice(own data: Iterator[int], stop: int) -> list[int]:
     try:
-        sliced: Iterator[int] = islice(iter(data), start_or_stop, stop, step)
+        sliced: Iterator[int] = islice(iter(data), stop)
+        return list(sliced)
+    except ValueError:
+        return []
+
+def collect_islice_range(own data: Iterator[int], start: int, stop: int | None, step: int | None = 1) -> list[int]:
+    try:
+        sliced: Iterator[int] = islice(iter(data), start, stop, step)
         return list(sliced)
     except ValueError:
         return []
@@ -32,6 +42,8 @@ def collect_core_actual() -> list[bool]:
     actual.append(batched_ok)
 
     actual.append(str(collect_islice(iter([10, 20, 30, 40]), 3)) == "[10, 20, 30]")
+    actual.append(str(collect_islice_range(iter([10, 20, 30, 40]), 1, None)) == "[20, 30, 40]")
+    actual.append(str(collect_islice_range(iter([10, 20, 30, 40]), 1, None, None)) == "[20, 30, 40]")
     actual.append(str(list(accumulate(iter([1, 2, 3, 4])))) == "[1, 3, 6, 10]")
     actual.append(str(list(accumulate(iter([1, 2, 3]), initial=0))) == "[0, 1, 3, 6]")
     actual.append(str(list(compress(iter([1, 2, 3]), iter([True, False, True])))) == "[1, 3]")
@@ -47,6 +59,17 @@ def collect_core_actual() -> list[bool]:
     c3: int | None = next(ctr)
     actual.append(c0 == 5 and c1 == 7 and c2 == 9 and c3 == 11)
     actual.append(str(list(cycle(iter([1, 2, 3]), 5))) == "[1, 2, 3, 1, 2]")
+    optional_values: list[int | None] = []
+    optional_values.append(optional_int(1))
+    optional_values.append(None)
+    optional_values.append(optional_int(2))
+    optional_expected: list[int | None] = []
+    optional_expected.append(optional_int(1))
+    optional_expected.append(None)
+    optional_expected.append(optional_int(2))
+    optional_expected.append(optional_int(1))
+    optional_expected.append(None)
+    actual.append(list(cycle(iter(optional_values), 5)) == optional_expected)
     actual.append(str(list(product([1, 2], [3, 4]))) == "[[1, 3], [1, 4], [2, 3], [2, 4]]")
     actual.append(str(list(product([1, 2], repeat=-1))) == "[]")
     actual.append(str(list(permutations(iter([1, 2, 3]), 2))) == "[[1, 2], [1, 3], [2, 1], [2, 3], [3, 1], [3, 2]]")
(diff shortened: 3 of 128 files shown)


yaseralnajjar:
Exact-SHA initial Opus review — base `868e86f27c77456d0c91a3678b2213d725c0836b`, candidate `1a11fcf55e578a57463148c0f53d7154f7accf9d`.

I reviewed the exact diff (`868e86f2` → `1a11fcf5`) and verified the generated companions directly with `rustc`.

**Verdict: NOT SATISFIED**

**Blocking findings:**

- **regression** — `crates/sifr_codegen/src/function_emitter/generic_bounds.rs:62-64` (with `crates/sifr_codegen/src/hir_analysis/queries/queries_impl.rs:598-670`), criterion "Keep generic Rust bounds demand-driven" and "Regenerate authoritative emitted demo companions". Dropping the unconditional `std::fmt::Display + PartialOrd` from the default type-parameter bound made bounds demand-driven **only from local-scope expressions** (`BinOp`/`Compare`/`UnaryOp`/`print`/`str`/f-string via `TraversalConfig::LOCAL_SCOPE_ONLY`). There is no propagation from a generic call site to its caller, so a function that merely forwards `T` to another generic function that does compare it now loses the required bound. Nine regenerated authoritative companions no longer type-check. Verified: `rustc --edition 2024 --extern sifr_runtime=... --emit=metadata demos/heapq/emitted.rs` → 3 × `E0277: can't compare T with T` at `demos/heapq/emitted.rs:206` (`heapify<T: Clone + 'static>` → `_sift_down<T: … + PartialOrd>`), `:214` (`heappush` → `_sift_up`), `:253` (`heappop` → `_sift_down`). Same failure in `demos/bisect/emitted.rs:107,116`, `demos/ordered_collections/emitted.rs:808,1038,1046,1085`, `demos/generic_stdlib/emitted.rs:920,961`, plus `demos/heap_option_drain`, `demos/max_heap`, `demos/python_regressions`, `demos/stdlib_expansion`, `demos/stdlib_ownership` (2–5 bound errors each). The same files at the exact base type-check with warnings only, so this is candidate-introduced and contradicts the recorded "corpus compilation passed" evidence — that run cannot have covered these emissions at this SHA. Correction: compute type-parameter bound requirements as a transitive closure over the generic call graph (a call whose callee constrains a type parameter must add that constraint to the caller's forwarded parameter, for `PartialOrd`/`Display`/`Hash + Eq` alike), regenerate every companion, and re-verify each of the nine files compiles.

**Follow-up findings:**

- **pre-existing issue** — `crates/sifr_runtime/src/string_padding.rs:69` (`Alignment::Center` → `left = padding / 2`): CPython's `str.center` uses `left = marg/2 + (marg & width & 1)`, so `"ab".center(5)` is `'  ab '` in CPython but `' ab  '` here. Base `lower_center` had the identical `_pad / 2` distribution, so the rewrite carried the divergence rather than introducing it, and the new fixture only asserts symmetric cases. Worth closing under the stdlib parity area with an odd-margin/odd-width case.
- **suggestion** — `crates/sifr_lowering/src/lower/expressions/call_shadowable_builtins.rs:59-68`: `register_compiler_open_handle_defaults` keys `function_defaults` by bare class name (`"FileHandle.seek"`, `"FileHandle.read_bytes"`), while the surrounding code deliberately guards identity (`FILE_HANDLE_IDENTITY`, "a local same-basename class is a distinct declaration"). A user-declared `FileHandle` with a differently-shaped `seek` in a module that also calls compiler-special `open()` could pick up a synthetic default. Key these by class identity.
- **suggestion** — `stdlib/sifr/itertools.sifr:99`: `islice(own data, start_or_stop: int, *slice_args)` still cannot express CPython's single-argument unbounded form `islice(it, None)`; the item's stated `islice(..., start, None)` case is covered, but `architecture.md:331-336` documents the contract without noting this remaining gap. Either accept the `None` first argument or record the divergence in the same paragraph.
- **suggestion** — emitted argument passing such as `file_read_bytes(&self._handle, (size.clone()).clone())` in `demos/io/emitted.rs:1418` shows redundant double clones on the new I/O paths; this is Item 7 (clone/borrow optimization) territory but the new surface adds fresh instances of it.

yaseralnajjar:
Exact-SHA sole remediation Opus review — base `868e86f27c77456d0c91a3678b2213d725c0836b`, candidate `511ec05e3ff21295fc0ba725f39abbe9900b1cdb`.

# Verdict

**NOT SATISFIED**

---

## 1. Blocking finding

### B1 — Propagated arithmetic bounds keep the *callee's* type-parameter spelling, emitting an out-of-scope type name (E0412)

The new closure copies callee bound **strings** verbatim onto caller type parameters. That is correct for name-independent traits (`PartialOrd`, `std::fmt::Display`, `PartialEq`, `std::hash::Hash`, `Eq`) — the three the initial review named — but every arithmetic requirement is *parameterized by the type-parameter name*, and nothing substitutes it during propagation.

Evidence chain:

- `crates/sifr_codegen/src/function_generic_bounds.rs:104` — declared `Addable` maps to `format!("std::ops::Add<Output = {type_param}>")` using the **callee's** parameter name.
- `crates/sifr_codegen/src/generic_bounds_helpers.rs:684-701` — inferred `Add/Sub/Mul/Div/Rem/Neg` bounds are likewise `…<Output = {tp}>` with the defining function's `tp`.
- `crates/sifr_codegen/src/function_generic_bounds.rs:76-88` — the fixed point does `target.extend(bounds.iter().cloned())`. There is no renaming/substitution step anywhere in the file; `forwarded_caller_type_params` computes *which* caller parameter receives the set, never rewrites the set's contents.
- `crates/sifr_codegen/src/function_emitter/generic_bounds.rs:89-110` — the closed strings are sorted and appended verbatim to the caller's rendered bound list. The dedupe filter compares whole strings, so `std::ops::Add<Output = T>` is *not* recognized as redundant with the caller's own `std::ops::Add<Output = U>`.

Source-reachable path (accepted by the front end):

```python
def add_same[T: Addable](left: T, right: T) -> T:
    return left + right

def relay_add[U: Addable](left: U, right: U) -> U:
    return add_same(left, right)

def main() -> None:
    print(relay_add(1, 2))
```

- `+` on a `TypeVar` requires a declared `Addable` bound: `crates/sifr_lowering/src/lower/expression_operators.rs:281-294` (negative test at `crates/sifr_lowering/src/lower/expressions_tests/basics_and_literals.rs:735-748`).
- Forwarding `U: Addable` into `T: Addable` type-checks by exact spec match: `crates/sifr_lowering/src/lower/type_bounds.rs:52-82`, `:548-551`. This is the same forwarding shape already governed as an authoritative pass fixture for `Comparable` (`crates/sifr/tests/e2e/pass/protocol_bound_forwarding_conforming_typevar.sifr`), and `demos/protocol_bounds/emitted.rs:8` shows caller/callee parameter names (`U`, `T`) are preserved distinctly in emission.

Emitted result: `fn relay_add<U: Clone + 'static + std::ops::Add<Output = T> + std::ops::Add<Output = U>>` → rustc **E0412 cannot find type `T` in this scope**.

**This is a candidate-introduced regression, not merely an unfixed case.** At the candidate's parent `1a11fcf55`, the shape below compiled:

```python
def double[T: Addable](a: T) -> T:
    return a + a

def combine[U: Addable](x: U, y: U) -> U:
    return double(x) + y
```

`combine`'s local `+` already yielded `Add<Output = U>`, satisfying `double::<U>`. Under the candidate, the closure additionally injects `Add<Output = T>` (not filtered, different string) and the function no longer compiles.

Why validation stayed green: no fixture or demo currently forwards an arithmetic protocol bound across differently-named parameters. `accumulate` is the only `Addable` function in the tree (`stdlib/sifr/itertools.sifr:394`), it is never called from another generic function, and its own parameter is `T`, so all five emitted occurrences render `Add<Output = T>` correctly (`demos/itertools/emitted.rs:192` and siblings). The defect is latent in the exact evidence set, and source-reachable.

Required fix: substitute the callee parameter for the corresponding caller parameter when propagating any bound whose rendering embeds a type-parameter name (reuse the same `formal ↔ actual` correspondence `forwarded_caller_type_params` already computes), or represent bounds structurally instead of as pre-rendered strings. Add an `Addable`-forwarding lowering regression plus a native/emitted companion so the case is governed.

Per the no-third-review rule I am not requesting another round. Concrete owner: a new **Item 6A** (following the Item 3A / 4A-4C precedent for a same-mechanism carry-forward), with Item 8 as the fallback owner if no 6A is opened.

---

## 2. Soundness of the transitive closure for the Item 6 authoritative surface

**Yes for the mechanism the initial review blocked; no as a general closure.**

- The original E0277 mechanism is resolved for `PartialOrd`, `Display`, and `Hash + Eq`. The least-fixed-point loop (`function_generic_bounds.rs:62-95`) genuinely closes multi-hop forwarding: `heappushpop → heappush → _sift_up` now yields `PartialOrd` on `heappushpop` (`demos/ordered_collections/emitted.rs:1117`), and the same holds for `heapify/heappop/heapreplace/nsmallest/nlargest/_heapify_max/_heappop_max/_heapreplace_max/insort_right`. Source protocol bounds survive with no local operator use (`demos/protocol_bounds/emitted.rs:4`), matching the new unit regression at `function_generic_bounds.rs:396-415`.
- Declared `Comparable`/`Hashable` mappings, hash-key-position inference (`function_generic_bounds.rs:110-134`), and deterministic sorted rendering are correct, and Eq's `PartialEq` supertrait is implied so no bound is under-specified.
- It is **not** sound in general: B1 above, plus the coverage gaps in §4.

---

## 3. Rest of the Item 6 implementation

Assessed against the Item 6 acceptance conditions and the deferred rows it owns:

- **Signed widths before allocation** — met. `ljust/rjust/center/zfill` no longer route through `to_usize_proven`; they call the new `crates/sifr_runtime/src/string_padding.rs` helpers, which clamp negative widths (`:69-71`), return a typed range error for out-of-platform widths, and use `try_reserve_exact` instead of an infallible allocation (`:98-106`). Character-count-based (not byte-based) width matches CPython; `zfill` sign handling is correct. Governed by `crates/sifr/tests/e2e/pass/string_padding_methods.sifr` including the `OverflowError` arm and `center(-100)`.
- **IO size/offset and error kinds** — met. `read_bytes(size)` now honors the limit through a chunked, allocation-checked reader (`crates/sifr_stdlib/src/fs.rs:428-467`); `seek/tell/flush` are real bridges with `InvalidInput` for a bad `whence`; the `raise IOError(e.message)` re-wrapping that flattened kinds to `"Other"` is replaced by `raise e` across `stdlib/sifr/io.sifr`, `pathlib.sifr`, `tempfile.sifr`, `_sifr/fs.sifr`, so `__io_err`'s classified `kind` survives (see `demos/decimal_verification/emitted.rs:3135-3155` vs `:3121-3127`).
- **Deferred rows owned by Item 6** — signed-zero exact division fixed at the single shared site (`crates/sifr_runtime/src/int.rs:604-609`); JSON map order reconciled via workspace `serde_json` `preserve_order`; `islice(it, start, None)` unbounded form implemented with explicit-`None` step handling (`stdlib/sifr/itertools.sifr:99-124`) and iterator-ownership contract documented (`internal_docs/architecture.md:331-336`); `cycle` no longer routes elements through an optional read, so `None`-valued optional elements are yielded and counted correctly (`stdlib/sifr/itertools.sifr:474-486`).
- **Inventory** — the two Item 6 rows correctly move their semantic anchors to `revision: "baseline"` (fixed in current), and the Item 11 row is re-anchored from the now-discharged `string.rs` site to the remaining `list.rs` site with owner/severity unchanged. No governed row was weakened or silently re-owned.
- **`Option`/`None` comparison lowering** (`crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_wrappers_and_compare.rs:604-654`) preserves Python chain semantics and removes the spurious `T: PartialEq` demand; chain continuation matches the surrounding handler.

---

## 4. Newly discovered defects deferred under the no-third-review rule

All non-blocking (no current source in the tree reaches them), each latent in the same mechanism:

| # | Defect | Evidence | Owner |
|---|---|---|---|
| D1 | Class-method type-param bounds and module-level function bounds are two disjoint closures; a generic class method forwarding a `T`-typed value into a generic free function gets no callee bounds, reproducing the original E0277 shape. Verified corpus-wide that this pattern has zero occurrences, so it is an untested specialization path. | `generic_bounds_helpers.rs:716-765` vs `function_generic_bounds.rs:18-98` | Item 8 |
| D2 | Call-site collection uses `TraversalConfig::LOCAL_SCOPE_ONLY`, so a generic callee invoked from a nested function/closure inside a generic body contributes no bounds. | `function_generic_bounds.rs:160-165` | Item 8 |
| D3 | The structural-mismatch fallback forwards the *full* callee bound set to every caller parameter mentioned anywhere in the actual type (e.g. formal `T` ↔ actual `tuple[A, B]` constrains both `A` and `B`), over-constraining callers of the caller. | `function_generic_bounds.rs:303-305` | Item 8 |
| D4 | `requirements` and the callee lookup are keyed by bare function name, so two same-named generic functions reachable in one emitted module would cross-contaminate bounds. Implicit invariant, unasserted. | `function_generic_bounds.rs:21-25`, `:70-75` | Item 8 |

---

## 5. Pre-existing / out-of-scope debt vs. candidate regressions

**Candidate regression:** B1 only.

**Pre-existing, not introduced here:**
- `str.center` odd-margin divergence from CPython (`padding/2` to the left; CPython uses `marg//2 + (marg & width & 1)`, so `"ab".center(5)` should be `"  ab "` but emits `" ab  "`). Identical at base — the rewrite preserved the old split (`string_padding.rs:87`). Does **not** invalidate an Item 6 acceptance condition (the governed inventory row is about signed-width validation, which is met), but the proposed **Item 12** ownership is a poor fit for a stdlib-semantics parity defect; recommend re-owning to Item 6A/Item 8.
- Redundant `__sifr_try_err.clone()` in the regenerated IO carriers (`demos/decimal_verification/emitted.rs:307-308`) — Item 7, as already recorded.
- `open()` default metadata keyed by bare class name — Item 8. Single-argument `islice(it, None)` — Item 12; the Item 6 acceptance form `islice(it, start, None)` is implemented.
- `exact_int_to_usize_expr` panic surface remaining in list methods — Item 11, correctly re-anchored rather than dropped.

**Worth one line of follow-up, not a finding:** enabling `serde_json`'s `preserve_order` at the workspace level also changes map ordering for the *compiler's own* JSON handling, not just generated programs; the gates passed, but any future governed JSON artifact now depends on insertion order rather than key sort.

yaseralnajjar:
Item 6 final implementation evidence for exact candidate `511ec05e3ff21295fc0ba725f39abbe9900b1cdb` (base `868e86f27c77456d0c91a3678b2213d725c0836b`):

- `cargo test -p sifr_codegen`: 1,189 passed.
- `cargo clippy -p sifr_codegen --lib -- -D warnings`: passed.
- All 10 changed authoritative emitted companions compiled with the required all-feature stdlib externs.
- Demo freshness, `cargo fmt --check`, HIR maintainability, 900-line file-size guardrail, emitted Rust audit inventory (33/29/4), and diff hygiene passed.
- Earlier exact-candidate Item 6 evidence covered focused E2E, codegen/lowering/runtime/exact-integer/stdlib API suites, seven native release fixtures, and generated corpus panic/determinism/freshness checks.

Review policy result:
- Initial exact-SHA review: NOT SATISFIED due missing transitive generic-bound propagation.
- Sole remediation review: NOT SATISFIED, but confirms the original PartialOrd/Display/Hash+Eq mechanism is fixed across the authoritative surface. Its newly discovered arithmetic bound parameter-substitution defect is rescaled to sequential Item 6A under the no-third-review rule. No further Item 6 review will run.

One-shot gates on this exact SHA:
- create-PR: stopped at preflight because the profile omits required SQL suites: host-tools, migration-engine, postgresql-live-differential, postgresql-live-migrations, postgresql-live-runtime, postgresql-live-schema-tools, postgresql-migrations, schema-tools.
- merge: stopped at the same preflight condition.
- Both profile-composition failures are owned by Item 12; neither gate will be repeated.

Since the reviewed base, main advanced only through SQL schema-polymorphism paths and records. Those changes do not overlap or invalidate the Item 6 compiler/runtime/stdlib evidence; the PR remains targeted at current main.

PR #3630, opened 2026-08-31
SQL M15: schema polymorphism and portable constraints

yaseralnajjar:
## Summary
- add provider-normalized structural schema requirements and deterministic subset proofs
- add compile-time schema witnesses, concrete-profile specialization, and verified execution binding
- add provider-owned capability accounting with PostgreSQL 13-18 artifacts and permanent qualification
- document the final portable SQL contract and manifest ergonomics

## Validation
- schema-polymorphism suite: 7/7
- adjacent SQL suites: schema profiles, query fragments, PostgreSQL compiler matrix and capability-free components
- rustfmt, HIR maintainability, file-size, and diff hygiene
- non-E2E Sifr sweep: all core/build-output tests passed; emitted-code safety reproduced the pre-existing Item 8 numeric_sentinels optional-index failure recorded in ad-hoc-emitted-rust-excellence.md

Draft pending exact-SHA Opus review and the one allowed create-PR/merge gates.


yaseralnajjar:
Reviewed the exact SHA range (`868e86f2..26cac063`, 53 files) with focus on the mechanism paths rather than re-running the supplied validation.

Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_sql_contract/src/requirement.rs:168` (with `crates/sifr_sql_contract/src/slice.rs:153-162`), criteria 1 and 3. `build_provider_schema_requirement` copies the *entire* normalized property map of every requirement object (`properties: object.semantic.clone()`), and `verify_compatible_slice` compares each property for exact equality. With the real PostgreSQL normalizer a table object carries membership lists — `columns`, `constraints`, `unique-sets`, `primary-key` (`crates/sifr_sql_postgresql/src/catalog.rs:687-720`) — so the "structural subset" proof degenerates into an exact table-shape match. A requirement built from `CREATE TABLE public.users (id bigint PRIMARY KEY, email text NOT NULL UNIQUE)` cannot be proven by any concrete profile whose `public.users` has one additional column, index, or check constraint: `prove` returns `IncompatibleSchema` on `public.users.columns`. This contradicts the milestone's own stated rule ("Extra objects in a concrete profile do not invalidate a structural requirement", `internal_docs/sql_schema_polymorphism.md`) and makes portable requirements unusable against real application schemas. No permanent fixture catches it: the contract test only adds an *unrelated* table (`public.audit`, `crates/sifr_sql_contract/tests/schema_polymorphism.rs:207-213`) to a synthetic `public.users` object that has no `columns` property, the frontend test builds the requirement from the *same* schema as the profile (`crates/sifr_frontend/tests/sql_schema_polymorphism.rs:141-172`), and the driver test points the requirement at the profile's own `db/schema.sql`. Correction: select only the structurally required properties per object (as `minimum_schema_slice` already does) and give membership-list properties containment rather than equality semantics; add a permanent case where a real normalized profile DDL adds a column and a constraint to a required table and the proof still succeeds.
- in-scope omission: `crates/sifr_frontend/src/sql_schema_polymorphism.rs:145-170`, criterion 7. The undeclared-object envelope is built only from relation-level effects (`effects.referenced_objects`, `effects.affected_objects`) plus result-field `source_object`. The PostgreSQL analyzer records only relation ids in `referenced`/`affected` (`crates/sifr_sql_postgresql/src/from_analysis.rs:48`, `crates/sifr_sql_postgresql/src/writes.rs:183,223,310`), so columns used only in a `WHERE`/`ON`/`GROUP BY` predicate or in an `UPDATE ... SET` assignment never enter the envelope. `SELECT id FROM public.users WHERE created_at > $1` and `UPDATE public.users SET last_login = $1` reach objects absent from the requirement without diagnosis. This is currently masked by the over-strict property comparison above, so fixing that finding alone opens the hole. Correction: have the provider account column-level object use in the effect contract (or add a distinct used-object set) and check it in `validate_query_envelope` alongside the relation set, with negative fixtures for a predicate-only and an assignment-only undeclared column.

Follow-up findings:
- infrastructure issue: `verification/areas/sql_platform/tools/check_schema_polymorphism.py` qualifies the milestone by asserting that specific token strings are present in Rust sources and that a hand-written JSON record matches hard-coded Python sets. The mutation self-test only mutates the JSON, so the checker cannot detect a behavioral regression in any of the mechanisms it names.
- infrastructure issue: `verification/areas/sql_platform/fixtures/schema_polymorphism/portable_by_email.sifr` is never parsed or compiled — only grepped for `has_users.Schema`, `SqlSchema[S]`, `by_email(app.schema`. There is no frontend surface for `from sifr.sql.requirements import ...`, `SqlSchema[...]`, or a requirement-constrained type parameter, and `sifr.sql.SqlSchema` (now emitted into every generated profile module, `crates/sifr_sql_contract/src/generated.rs:109`) has no definition anywhere. The witness use-site rules are enforced only over the hand-classified `SqlSchemaWitnessUse` enum, not over real resolution, so criterion 5 holds at harness level only.
- suggestion: the provider capability vocabulary is 16 coarse entries, while `ProviderAnalysis.required_capabilities` is documented as "every SQL capability used by the analyzed statement" (`crates/sifr_sql_contract/src/provider.rs:39-41`). Casts, array literals, `LATERAL`, `FILTER`, and arbitrary non-aggregate provider functions contribute no capability, so provider-specific behavior of those forms passes the envelope check. Either narrow the doc comment to "the qualified capability surface" or extend the vocabulary.
- suggestion: `normalized_schema_from_response` (`crates/sifr_sql_contract/src/component.rs:129`) is now a thin shim over `schema_normalization_from_response` with no production caller — only `tests/schema_profiles.rs` uses it. Removing it keeps the crate free of an unrequested compatibility path.
- pre-existing issue: the emitted-code safety tail `numeric_sentinels` optional-index failure reproduces on the base SHA and unrelated candidates and is already assigned to ad-hoc-emitted-rust-excellence Item 8; not attributable to this change.

yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- in-scope omission: `crates/sifr_sql_postgresql/src/ddl_constraints.rs:41` (with `crates/sifr_sql_contract/src/requirement.rs:158` and `crates/sifr_sql_contract/src/slice.rs:213`), criterion "normalized-DDL structural schema requirements — a valid concrete application schema with extra columns or constraints must not fail the structural subset proof" (prior blocking finding 1). Reason: the containment relaxation was applied only to the table-level `columns`/`constraints`/`unique-sets` lists, but the requirement builder still requests every normalized requirement object — including each constraint object — with all of its properties, and those objects are matched by *positional* generated identity (`{table}_unique_{i+1}`, `{table}_check_{n}`, `{table}_fkey_{n}`, numbered from `unique_sets` iteration order and running `identities.len()`). Adding an application-only constraint that sorts or occurs before a required one renames the required constraint's identity in the concrete profile, so `verify_compatible_slice` compares the requirement's constraint against a different constraint object and rejects a valid superset schema. Failure scenario: requirement DDL `CREATE TABLE public.users (id bigint PRIMARY KEY, email text NOT NULL UNIQUE);` yields `public.users_unique_1` with `columns = ["email"]`; application DDL `CREATE TABLE public.users (id bigint PRIMARY KEY, email text NOT NULL UNIQUE, display_name text UNIQUE);` yields `public.users_unique_1` with `columns = ["display_name"]` and `public.users_unique_2` with `["email"]`, so `SchemaRequirement::prove` fails with `IncompatibleSchema: required schema property 'public.users_unique_1.columns' changed`. The new PostgreSQL test at `crates/sifr_sql_postgresql/tests/postgresql_compiler.rs:311` only adds a trailing `CHECK`, whose generated index happens not to collide, so it does not cover this class. Correction: make constraint object identities content-derived (kind plus normalized column list / expression digest) instead of index-derived, or match required constraint objects structurally rather than by generated identity, and extend the normalized PostgreSQL test to an application-only unique constraint that sorts before a required one (and an application-only CHECK/FK declared before a required one).

Prior blocking finding 2 is corrected: `ProviderAnalysis.accessed_objects` (`crates/sifr_sql_contract/src/provider.rs:39`) is provider-owned, populated by the PostgreSQL analyzer at every column resolution (`analysis.rs:317`), JOIN USING (`from_analysis.rs:137`), insert targets, conflict targets, and update assignments (`writes.rs:36,108,271`), validated as a superset of the effect/result account (`provider.rs:107-125`), consumed by `validate_query_envelope` (`sql_schema_polymorphism.rs:148`), and covered by predicate-only and assignment-only negative specialization fixtures plus PostgreSQL assertions on join/predicate/assignment columns.

Follow-up findings:
- suggestion: `accessed_objects` completeness is enforced only as a superset of the previously derived effect/result-field account; a provider that under-reports predicate-only columns would still validate. Consider a provider-side property test asserting that every catalog column named in a statement appears in `accessed_objects`.
- suggestion: `analyze_insert` records every relation column when the INSERT omits a column list (`writes.rs:31-39`), so a column-list-free INSERT under a wider application schema will report application-only columns as accessed and be rejected as undeclared. That is defensible strictness, but it deserves an explicit note in `internal_docs/sql_schema_polymorphism.md`.
- pre-existing issue: the negative specialization fixtures in `crates/sifr_frontend/tests/sql_schema_polymorphism.rs` hand-construct `ProviderAnalysis` values rather than deriving them from real SQL, so they verify the envelope check rather than provider accounting — this overlaps the already-deferred "behavioral qualification beyond token checks" follow-up.

yaseralnajjar:
Final exact candidate: `31eb20ff2e634bd299bc620b76705a141f702244`

One-shot gate evidence:
- `scripts/run_all_tests.sh --profile create-pr`: stopped before tests because the profile omits nine required SQL platform suites.
- `scripts/run_all_tests.sh`: stopped before tests for the same merge-profile composition defect.

Per the phase rule, neither gate is rerun. The second review’s newly discovered positional constraint-identity mechanism defect is deferred to M18; no third review round is permitted.

PR #3633, opened 2026-08-31
fix(codegen): preserve generic bound parameter identity

yaseralnajjar:
## Summary

- represent module-level generic operator bounds structurally as ordinary traits or self-output traits
- render self-output bounds with the receiving function's type-parameter identity during fixed-point propagation
- govern differently named `Addable` forwarding in the authoritative protocol demo and E2E fixture
- match CPython odd-margin placement in `str.center`

## Validation

Exact candidate: `aff53a422ee8c55185d0110c51483be0d600375d`

- `cargo test -p sifr_codegen`: 1,190 passed
- `cargo test -p sifr_runtime`: 89 unit + 8 exact-integer tests passed
- strict codegen/runtime Clippy passed
- both focused Sifr fixtures compiled and ran as release-native binaries
- authoritative protocol companion emitted `Add<Output = U>`, compiled as Rust metadata, and is fresh
- formatting, HIR maintainability, file-size, audit inventory, and diff checks passed

### DIFF
--- a/crates/sifr/tests/e2e/pass/protocol_bound_forwarding_conforming_typevar.sifr
+++ b/crates/sifr/tests/e2e/pass/protocol_bound_forwarding_conforming_typevar.sifr
@@ -6,8 +6,16 @@ def keep_comparable[T: Comparable](x: T) -> T:
 def relay_comparable[U: Comparable](x: U) -> U:
     return keep_comparable(x)
 
+def add_same[T: Addable](left: T, right: T) -> T:
+    return left + right
+
+def relay_add[U: Addable](left: U, right: U) -> U:
+    return add_same(left, right)
+
 def main():
     n: int = relay_comparable(7)
     s: str = relay_comparable("ok")
+    total: int = relay_add(4, 5)
     assert str(n) == '7'
     assert str(s) == 'ok'
+    assert total == 9
--- a/crates/sifr/tests/e2e/pass/string_padding_methods.sifr
+++ b/crates/sifr/tests/e2e/pass/string_padding_methods.sifr
@@ -3,6 +3,7 @@ def main():
     try:
         s: str = "x"
         centered: str = s.center(5)
+        odd_margin: str = "ab".center(5)
         left: str = s.ljust(4)
         right: str = s.rjust(4)
         zeroed: str = "7".zfill(4)
@@ -10,6 +11,7 @@ def main():
         unicode: str = "é".ljust(3)
         unchanged: str = s.center(-100)
         assert str("center = [" + centered + "]") == 'center = [  x  ]'
+        assert odd_margin == "  ab "
         assert str("ljust = [" + left + "]") == 'ljust = [x   ]'
         assert str("rjust = [" + right + "]") == 'rjust = [   x]'
         assert str("zfill = [" + zeroed + "]") == 'zfill = [0007]'
--- a/crates/sifr_codegen/src/function_emitter/generic_bounds.rs
+++ b/crates/sifr_codegen/src/function_emitter/generic_bounds.rs
@@ -92,8 +92,8 @@ impl RustEmitter {
                     .and_then(|by_param| by_param.get(type_param))
                     .into_iter()
                     .flatten()
+                    .map(|bound| bound.render_for(type_param))
                     .filter(|bound| !extra_bounds.contains(bound))
-                    .cloned()
                     .collect::<Vec<_>>();
                 closed_bounds.sort();
                 extra_bounds.extend(closed_bounds);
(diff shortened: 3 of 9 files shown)


yaseralnajjar:
Exact-SHA Opus review — base `2bd6f373fb62ecba70bff7e95cffd4a8d52eb97c`, candidate `aff53a422ee8c55185d0110c51483be0d600375d`.

Read-only review complete. I independently rebuilt the compiler at `aff53a422` and re-derived the emission rather than relying only on the stated evidence.

**What I verified directly**
- `demos/protocol_bounds/emitted.rs` is byte-fresh: emitting `demos/protocol_bounds/main.sifr` with the candidate compiler reproduces it exactly, including `add_same<T: … ::std::ops::Add<Output = T>>` and `relay_add<U: … ::std::ops::Add<Output = U>>`. The `::` prefix comes from the `stdlib_filter` path-absolutization pass (`crates/sifr_codegen/src/stdlib_filter/implementation.rs:90`), not hand editing.
- Both touched E2E fixtures compile and run as release-native binaries with all asserts passing.
- Fresh source-reachable probes: `[T: Comparable]` forwarding through `-`, `*`, `/`, `%` emits `Sub/Mul/Div/Rem<Output = U>` on the *caller's* `U`; `Comparable` → `PartialOrd` and `Hashable` → `Hash + Eq` closure still propagates.
- Only consumer of the closed map is `crates/sifr_codegen/src/function_emitter/generic_bounds.rs:95`, which renders with the receiving `type_param`; `Trait(String)` payloads are all constant strings containing no parameter name, and `OutputSelf` is the sole variant that writes one.
- Exhaustive differential of the new `center` rule (`padding / 2 + (padding & width & 1)`) against CPython `str.center` over lengths 0–7 × widths −3..11: zero mismatches. Negative width short-circuits through `normalized_width` → `padding == 0`, so the bit rule is never reached with a clamped width; oversized width still returns the typed range error.

Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- pre-existing issue: `crates/sifr_codegen/src/function_generic_bounds.rs:161` maps `Addable` to `std::ops::Add<Output = Self>`, but the frontend admits `str` for that bound (`crates/sifr_lowering/src/lower/type_bounds.rs:551`). `add_same[T: Addable]("a", "b")` type-checks and then fails `rustc` E0277 (`String: Add<String>` does not exist; only `Add<&str>`). I reproduced this on the candidate; base emits the identical bound for the declaring function, so it is not an Item 6A regression and is outside the "differently named forwarding" criterion. Concrete owner: **Item 7 (Ownership, borrowing, and clone quality)** — the fix is choosing the borrowed operand shape for generic `+` so the `Addable` protocol's admitted `str` member has a satisfiable Rust bound; add a `str` instantiation fixture alongside the existing `int` one.
- suggestion: `demos/protocol_bounds/idiomatic.rs` still models only `keep_comparable`/`relay_comparable` and prints three lines while `main.sifr` now prints four. No gate compares them (`scripts/check_demo_emitted_freshness.py` only diffs `emitted.rs` against `main.sifr`), and rewriting hand-authored `idiomatic.rs` files is explicitly phase-out-of-scope, so this is not blocking. Concrete owner: **Item 8**, which already holds the deferred row for reconciling non-authoritative demo companions against their sibling sources.

**Prior E0412 mechanism:** closed for every self-output arithmetic trait this candidate represents. `Add`, `Sub`, `Mul`, `Div`, `Rem`, and `Neg` are all constructed as `FunctionTypeParamBound::OutputSelf(trait_path)` in the single canonical `direct_type_param_bounds` constructor (plus `Addable` in `rust_bounds_for_typevar_spec`), carrying no parameter spelling. A callee-local name can no longer survive propagation because the only place a type-parameter identifier enters a bound string is `render_for(type_param)`, and its sole call sites pass the receiving parameter. I confirmed Add/Sub/Mul/Div/Rem end-to-end through source-reachable forwarding; `Neg` is structurally identical but is not currently reachable through a type variable (the frontend rejects unary `-` on `T` with `SIFR-TYPE-0005`), so it is closed by construction rather than by fixture.

yaseralnajjar:
Final exact-candidate evidence for `aff53a422ee8c55185d0110c51483be0d600375d`:

- 1,190 codegen tests passed.
- 89 runtime unit tests and 8 exact-integer architecture tests passed.
- Strict codegen/runtime Clippy passed.
- Differently named `Addable` forwarding and odd-center fixtures compiled and ran as release-native binaries.
- The authoritative protocol companion is fresh, emits `Add<Output = U>` on `relay_add<U>`, and passed standalone Rust metadata compilation.
- Formatting, HIR maintainability, 900-line file-size guardrail, emitted audit inventory (33/29/4), and diff hygiene passed.
- Exact-SHA Opus review: SATISFIED with no blockers.

One-shot gates on this exact SHA:
- create-PR: stopped at profile preflight because required SQL suites are omitted: host-tools, migration-engine, postgresql-live-differential, postgresql-live-migrations, postgresql-live-runtime, postgresql-live-schema-tools, postgresql-migrations, schema-polymorphism, schema-tools.
- merge: stopped at the identical preflight.
- Profile composition remains Item 12-owned; neither gate will be repeated.

PR #3635, opened 2026-08-31
feat(sql): complete MySQL provider

yaseralnajjar:
## Summary
- add the complete provider-owned MySQL compiler, runtime, host tools, migrations, and editor integration
- qualify exact MySQL 8.4, 9.7, and 26.7 behavior with checked-in WebAssembly components and live matrices
- extend shared SQL contracts for safe provisioning and cancellation discard semantics

## Validation
- SQL platform verification: 62 variants, 0 failures
- live MySQL all-surface matrix: 8.4, 9.7, and 26.7 passed
- strict Clippy for all M16-owned crates and tests
- formatting, HIR maintainability, and file-size guardrails passed

Draft pending exact-SHA Opus review and the one-shot create-PR/merge gates.


yaseralnajjar:
Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_sql_contract/src/profile.rs:46` (`dialect_modes_for_session`), consumed at `crates/sifr_package/src/sql_profiles.rs:60` and `crates/sifr_package/src/sql_requirements.rs:51`, criterion 6. The new mode encoding folds `session.character-set` and `session.collation` into the expected `dialect.modes` for every family, but the PostgreSQL component always emits `modes: BTreeSet::new()` (`crates/sifr_sql_postgresql/src/component.rs:555`, `crates/sifr_sql_postgresql/src/catalog.rs:532`), and `[sql.profiles.<name>.session] collation`/`character-set` are accepted for any family (`crates/sifr_package/src/manifest/sql_profiles.rs:604-630`). A PostgreSQL or SQLite profile that declares either key validated before this change and now fails `build_authority` with "normalized schema dialect inputs do not match the profile configuration". No fixture sets those keys on a non-MySQL profile, so the suites do not catch it. Correction: gate the charset/collation mode encoding on the MySQL dialect family (or reject those two session keys at manifest parse for non-MySQL families) so non-MySQL profiles compare against `session.sql_modes` as before.
- in-scope omission: `crates/sifr_sql_mysql_runtime/src/control.rs:44-53` (`arm_cancellation`), criterion 4. The carrier-driven cancellation path spawns `kill_query(spawned_opts, target)` with no time bound and no cleanup evidence, while only the timeout path (`cancel_after_timeout`, `control.rs:81-85`) applies `cleanup_timeout`. `MysqlProfile::new` does not require any connect/read timeout on `control_opts` (`config.rs:51-58`), so a wedged control connection leaves an unbounded detached task. This contradicts the crate metadata `cancellation = "bounded-kill-query-then-discard"` (`crates/sifr_sql_mysql_runtime/Cargo.toml:15`) and `internal_docs/sql_mysql_provider.md:203-204` ("within the cleanup budget"). Correction: wrap the spawned `kill_query` in `tokio::time::timeout(control.cleanup_timeout, ...)` (already carried on `ControlHandle`) and record `AsyncCleanupEvidence` on the carrier for failure and timeout, matching `cancel_after_timeout`; the discard half of the contract is already correct.

Follow-up findings:
- suggestion: `crates/sifr_sql_mysql/src/analyzer.rs:41-56` — the mode check re-reads the `character-set:` entry from the schema it is validating and echoes it into the expected set, so the character set is never actually verified against parser-owned state (SQL modes and collation are). An owned expected character set on `MysqlParser` would make the gate exact.
- suggestion: `crates/sifr_sql_mysql/src/syntax_lowering.rs:314-317` — a table with `DEFAULT CHARACTER SET <cs>` and no `COLLATE` inherits the profile default collation rather than the declared character set's default collation, which can diverge from the server's own resolution.
- suggestion: `crates/sifr_sql_mysql_runtime/tests/live_runtime.rs:122-135` — the only live cancellation case passes `cancellation: None`, exercising just the timeout path; the `CancellationCarrier` claim path (the one flagged above) has no live coverage.
- suggestion: `crates/sifr_sql_mysql_runtime/src/transaction.rs:126-162` — `MysqlSavepoint::finished` is written but never read and the type has no `Drop`, so an abandoned savepoint neither poisons nor rolls back; either drop the field or give it release semantics.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Evidence reviewed (candidate `5d585f4b0` vs base `7bcd08839`, remediation commit `5d585f4b0`):

Finding 1 (dialect mode regression) — corrected. `crates/sifr_sql_contract/src/profile.rs:47-66` now takes `dialect_family` and appends `character-set:`/`collation:` modes only when the family is `mysql`; base sql_modes pass through for every dialect. Both authority callers pass a dialect family: `crates/sifr_package/src/sql_profiles.rs:61` uses `schema.dialect.family`, and `crates/sifr_package/src/sql_requirements.rs:48-53` compares `normalized.dialect.family` against the manifest-declared family *before* deriving modes, so the family is authoritative there. These are the only production call sites (grep confirms). Producer side stays consistent: the MySQL component emits `sql_modes + character-set:… + collation:…` (`crates/sifr_sql_mysql/src/schema.rs:155-168`) while the PostgreSQL component emits an empty mode set (`crates/sifr_sql_postgresql/src/component.rs:555`), so PostgreSQL/SQLite authority comparison no longer sees session charset keys. Regression tests cover both directions (`crates/sifr_sql_contract/tests/schema_profiles.rs:23-42`, `crates/sifr_package/src/sql_profile_tests.rs:93-183` with legacy `collation`/`character-set` manifest keys present).

Finding 2 (unbounded carrier kill) — corrected. `crates/sifr_sql_mysql_runtime/src/control.rs:33-66` captures `Handle::try_current()` at arm time, and the claimed hook now spawns the shared `bounded_kill_query` (`control.rs:102-124`) under `control.cleanup_timeout`, recording `AsyncCleanupEvidence::cleanup_failed`/`cleanup_timed_out` on a carrier clone. `cancel_after_timeout` reuses the same helper, so deadline and carrier paths share one bounded implementation. Poison/discard is untouched (`connection.rs:130,173`, `pool.rs:187`).

No regression introduced by the remediation:
- `arm_cancellation`'s new `Handle::try_current()` failure path cannot regress callers: every call flows through `run_controlled`, which already requires a Tokio timer context for `tokio::time::timeout` (`control.rs:69-80`; callers in `connection.rs:117,149` and `stream.rs:73,93,199,219`).
- The hook holds a `CancellationCarrier` clone while the carrier holds the hook, but `CancellationClaimLease::drop` clears `state.exact` (`crates/sifr_runtime/src/cancellation.rs`), so the Arc cycle is broken when `run_controlled` returns — no leak, including on repeated per-chunk stream claims.
- Hooks are invoked after the state mutex guard is released, and `Handle::spawn` after runtime shutdown drops the task rather than panicking, so panic-safety is unchanged.
- Component/artifact churn is recompilation only: digests and sizes in `component-artifacts.json` match the regenerated wasm, and no component calls `dialect_modes_for_session`.

Follow-up findings:
- suggestion: `crates/sifr_sql_mysql_runtime/tests/live_runtime.rs:176-180` asserts `take_async_cleanup_evidence().is_empty()` immediately after the cancelled future resolves, but the carrier kill task is detached; the assertion can pass before the task records anything, so it only weakly proves "no evidence on successful KILL QUERY". Joining or polling with a bounded wait would make it exact.
- suggestion: unlike the requirement path, `ResolvedSqlProfile::build_authority` (`crates/sifr_package/src/sql_profiles.rs:61`) derives the mode gate from the provider-reported `schema.dialect.family` with no manifest-declared family to compare against (`SqlProfileConfig` has no `family`). Provider identity is verified, so this is not exploitable today, but pinning the expected family in the profile config would make the gate authoritative on both paths.
- suggestion: the carrier path's kill remains detached, so its evidence is recorded on the carrier only and is never attached as secondary evidence to the returned `Cancelled` error, and the kill can in principle land after the target connection id is retired — narrower than the deadline path's awaited behavior, and unchanged from base in that respect.
- pre-existing issue: `sifr_package` all-target lint debt remains outside this item; its library target passes strict Clippy.
- infrastructure issue: the final one-shot create-PR and merge gates have not run yet; approval here is on the reviewed content, not on those gates.

yaseralnajjar:
Exact-SHA gate evidence for `5d585f4b0b05d3572a1cededcc81c0fc1823b40f`:

- `scripts/run_all_tests.sh --profile create-pr` was invoked exactly once and stopped before tests because the profile omits the required SQL-platform suites: `host-tools`, `migration-engine`, `mysql-live`, `mysql-provider`, `postgresql-live-differential`, `postgresql-live-migrations`, `postgresql-live-runtime`, `postgresql-live-schema-tools`, `postgresql-migrations`, `schema-polymorphism`, and `schema-tools`.
- `scripts/run_all_tests.sh` was invoked exactly once and stopped before tests for the same merge-profile omission.
- Per the phase rule, neither gate was rerun.
- Independent validation on this SHA includes the complete live MySQL 8.4/9.7/26.7 matrix, component qualifications, focused shared-contract/package tests, strict M16 Clippy, and all guardrails. The earlier full SQL-platform adapter passed all 62 variants with zero failures before the base-only serialization feature sync; every affected component and MySQL surface was rebuilt and revalidated after that sync.

PR #3637, opened 2026-08-31
refactor(codegen): make emitted ownership explicit

yaseralnajjar:
## Summary
- introduce one explicit ownership/materialization plan for emitted values
- prefer unsized string, byte, and collection views throughout generated APIs
- remove redundant clone chains and enforce representative clone/complexity budgets
- close deferred ownership defects for callback effects, context targets, receiver mutation, checked witnesses, IO, and generic string addition
- regenerate all authoritative demo companions

## Validation
- cargo test -p sifr_codegen: 1,213 passed
- cargo test -p sifr_lowering: 1,094 passed; 1 intentional ignore
- cargo fmt --check
- direct full E2E: 717 passed; only exact-base numeric_sentinels failed (Item 8-owned)
- generated demo freshness and emitted Rust audit inventory
- sole create-PR and merge gates both stopped at the same Item 12-owned SQL profile preflight omission and were not repeated

## Review
- initial exact-SHA review found two blockers
- sole remediation review verified both blockers fixed
- its new receiver-effect distinction defect and latent non-registry setdefault ownership contract are deferred to immediate Item 7A under the no-third-review rule


yaseralnajjar:
## Independent verification performed (read-only, HEAD binary at `27840aa67`)

I reviewed the exact diff and reproduced emitted/native behavior with the already-built `target/debug/sifr` in `/tmp` (no repo files touched). Four minimal programs were compiled; two reveal defects introduced by this candidate.

Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_codegen/src/module_body.rs:22` (gating) with `crates/sifr_codegen/src/function_generic_bounds.rs:161` and `:32` — criterion 5 (`Addable` operand ownership agrees with its Rust trait impl) and criterion 6 (generated native compilation). `Addable`/`needs_add` now renders the bound `__SifrAdd`, but `ownership_plan::addable_support_items()` is emitted only when `self.function_type_param_bounds` (free functions) demands it. A generic **class** type parameter or class method renders `impl<T: Clone + __SifrAdd>` with no trait definition. Repro (`sifr build`): `class Accumulator[T: Addable]: total: T; def add(self, extra: T) -> T: return self.total + extra` → `error[E0405]: cannot find trait '__SifrAdd' in this scope` and `error[E0433]` on `__SifrAdd::__sifr_add`. The base spelling `std::ops::Add<Output = T>` needed no support item, so this is newly broken output, not the Item 8-owned bound-demand gap. Correction: make support-item emission demand-driven from every rendered bound source (class type params and class-method bounds included, e.g. a shared "bound requires support item" registry consulted after all bounds are computed), and add a generic-class `Addable` shape test plus a native regression.
- regression: `crates/sifr_codegen/src/intrinsic_method_emitters/collection_methods.rs:287-303` (list `insert`) and the `Type::Set(..) , "add"` element path reached from `collection_methods.rs:252-266` — criterion 2 (clone insertion driven by the explicit ownership plan) and criterion 6. These sites still construct a raw `.clone()` instead of `ownership_plan::materialize_owned_value`, and with the new unsized-view ABI `.clone()` on `&str`/`&[T]` reproduces the reference, exactly the hazard documented in `ownership_plan.rs:110-114`. Repros: (a) `def add_key(mut seen: set[str], key: str) -> None: seen.add(key)` → `seen.insert((key).clone());` → `error[E0308]: expected 'String', found '&str'`; (b) `out.insert(0, text)` with `text: str`, `mut out: list[str]` → `out.insert(..., text.clone())` → same E0308. Both compiled at base (`&String.clone()` → `String`). Correction: route every owned collection-element boundary (list `append`/`insert`, set `add`, dict key/value insert, `setdefault`) through `materialize_owned_value`, and extend the new ownership-quality budgets with these shapes, since the current suite asserts only call-argument and local-binding boundaries.

Follow-up findings:
- pre-existing issue: unguarded `dict[key]` in a value position emits `mapping.get(&key).map_or("None".to_string(), ...)` — both a silent `"None"` fallback and `error[E0277]: String: Borrow<&str>`; `set.remove/discard(key)` and `key in mut_set` likewise emit `&key`. I verified with a direct `rustc` check that the base `&String` spelling fails identically (`String: Borrow<&String>`), so these are not introduced here; they belong to the deferred optional-place/`&Q` normalization family and need an explicit owner row.
- suggestion: `crates/sifr_lowering/src/lower/nested_function_inference/call_effects.rs:12-15` keeps `_captures`/`_params` after removing the callable-alias effect; drop the parameters so the narrowed contract is stated by the signature.
- suggestion: the with/async-with mechanism was already structurally covered (`state_collection.rs:577` collects `optional_vars`), so this closure is regression-only; `nested_async_context_target_is_local_instead_of_an_outer_capture` binds `group` and never reads it, so it does not exercise the misclassification it guards.
- suggestion: receiver-effect "unification" adds a second, convention-driven invalidation path while the older fixed list `mutating_methods.rs:185 method_invalidates_collection_flow_facts` remains and is now subsumed by it; collapsing to one summary would match the stated required action.
- suggestion: `ownership_plan::string_view_expr` drops the previous explicit `Paren`; the renderer parenthesizes `BinOp`/`If`/`Match`/… but not `Ref`/`Deref` receivers, so a future `&`-prefixed lowering at any of the 18 call sites would silently render `&x.as_str()`.

yaseralnajjar:
Verification performed read-only at the exact SHAs. I built the candidate compiler (`778e13268`) and, for differential evidence, the exact base compiler (`02842e49`) from `git archive` in `/tmp` (no repo files touched, separate `CARGO_TARGET_DIR`).

**Prior blocker 1 — verified fixed.** `module_requires_addable_support` (`crates/sifr_codegen/src/function_generic_bounds.rs:26-54`) spans class `methods` + `operator_impls` × (`class.type_params` ∪ `method.type_params`), which is a superset of every rendered source (`class_method_type_param_bounds` computes bounds from the same `direct_type_param_bounds`, and its transitive closure only propagates bounds already present in some method body). Native repro of the prior failure now emits `pub trait __SifrAdd`, `impl<T: Clone + __SifrAdd> Accumulator<T>`, `__SifrAdd::__sifr_add(...)` and runs; inheritance and self-call variants also build and run. Duplicate-trait risk across modules is handled by `stdlib_filter::dedup_rust_items`.

**Prior blocker 2 — verified fixed.** All four boundaries route through `ownership_plan::materialize_owned_value`. Emitted for `set[str]`/`list[str]`/`dict[str,str]` with a borrowed `value: str`: `seen.insert(key.to_owned())`, `out.insert(..., text.to_owned())`, `defaults.entry(value.to_owned()).or_insert(value.to_owned()).clone()`; a local owned key (`table.setdefault(k, v)` with later use of `k`/`v`) emits `entry(k.to_owned()).or_insert(v.to_owned())` and compiles/runs. No raw `.clone()` on a `&str` remains on these paths; the `setdefault` clone removal in `methods/dict.rs` is compensated on the registry path. Demos show a net −1,172 `.clone()` occurrences.

However, the remediation's collapse of the receiver-effect mechanism introduced a new, verified behavioral regression against the exact base.

Verdict: NOT SATISFIED

Blocking findings:
- regression: `crates/sifr_lowering/src/lower/expressions/method_calls.rs:171-189` (unconditional `invalidate_mutable_receiver_sequence_guards` with the `receiver_convention_for_non_class_method` fallback) together with the deletion of `method_invalidates_collection_flow_facts` / `invalidate_collection_flow_facts_for_method` (`crates/sifr_lowering/src/lower/mutating_methods.rs`, −45 lines) and `crates/sifr_lowering/src/lower/mutating_methods.rs:26-30`. Criteria 5 (mutating receiver/argument summaries) and 6 (preserve semantic correctness). For builtin receivers `method_function_type` returns `None` (it resolves only `Class`/`Protocol`), so before this commit builtins were summarized by the deleted shrinking-only set (list `clear|pop|popleft|remove`, dict `clear|pop`, set `remove|discard|clear|*_update`). The remediation replaced it with `is_collection_mutating_method`, which reports `MutableBorrow` for growth-only and order-only mutators too, so `sequence_guards.rs:301-323` now clears length/key guards and narrowing for `append`/`appendleft`/`insert`/`extend`/`sort`/`reverse`/`set.add`/`set.update`/`dict.update`/`dict.setdefault` — none of which can falsify a `len(x) > 0` or `k in m` guard. Verified regression (same file, both binaries): `def head(mut xs: list[int]) -> int:` / `if len(xs) > 0:` / `xs.append(5)` / `return xs[0]` → base `02842e49`: "no errors found"; candidate `778e1326`: `error[SIFR-TYPE-0002]: return type mismatch: expected 'int', got 'None | int'`. Same divergence for `xs.sort()`, `xs.extend(ys)`, and `if "a" in m: m.setdefault("b", 2); return m["a"]`. Correction: keep the single convention-driven entry point, but have it consult a builtin receiver-effect summary that preserves the shrinking/reordering distinction the deleted function encoded — clear length/key guards only for methods that can shrink or remove (and clear element-position narrowing for `sort`/`reverse`) — and add regression tests asserting a length/key guard survives `append`/`insert`/`extend`/`setdefault` while `pop`/`clear`/`remove` still invalidate (alongside the existing `builtin_shrinking_receiver_invalidates_sequence_guard`, whose name already states the intended scope).

Follow-up findings:
- pre-existing issue: the dict/set lookup-key `&Q`/`Borrow` normalization defect and the silent `"None"` fallback in unguarded `dict[key]` value positions reproduce identically on the exact base; they belong to Item 8's canonical optional-place and Rust IR cleanup work.
- new mechanism defect: `methods/dict.rs::lower_setdefault` no longer materializes key/default ownership itself and now relies entirely on its caller. The registry path in `collection_methods.rs:249-278` does this correctly, but `lower_stmt/candidate_and_validation.rs:72` (`try_lower_expr_stmt_with_bindings`, `Any`/`Unknown` receiver resolved from `local_binding_types`) reaches `lower_method` with raw leaf args and no ownership plan. I could not construct a reaching source program, so this is a latent contract gap rather than a demonstrated defect; record it as a later item (e.g. assert the ownership plan inside the boundary or route that leaf path through the registry arg preparation).
- suggestion: the new registry-literal and `entry` key boundaries materialize for every non-`Copy` `Name`, not only borrowed ones, so owned locals that are dead afterwards now pay an avoidable `to_owned()`/`clone()` (e.g. `table.entry(k.to_owned()).or_insert(v.to_owned())` where `k`/`v` are never reused). Within the clone budgets, but a candidate for last-use move promotion in `ir_optimize`.
- infrastructure issue: nested generic functions with a `TypeVar` bound (`def inner[T: Addable](...)` inside a function) are rejected by the frontend with SIFR-TYPE-0005/0002; unrelated to this diff and cleanly diagnosed, but it means the shared support-demand query has no nested-function bound source to cover today.

PR #3640, opened 2026-08-31
feat(sql): complete SQLite provider

yaseralnajjar:
Closes Milestone 17 of the schema-first SQL platform phase.\n\nThis completes the SQLite compiler component, dedicated-worker bundled runtime, catalog and lifecycle tools, common migration-engine integration, editor behavior, portable requirement specialization, and locked qualification evidence.\n\nValidation and exact-SHA review evidence will be added before this draft is marked ready.


yaseralnajjar:
Exact-SHA Opus remediation review for candidate `22a87c70d2b41bdb76ec4dc2ebc51ae13f17e25e`:

I reviewed the exact commit range (`917a4e89` → `22a87c70`) across the SQLite compiler, runtime, tools, contract/package changes, docs, and verification inputs, focusing on A1–A18 and B1–B5.

What I confirmed as corrected (not re-reported): live catalog now reads `<schema>.sqlite_schema` DDL and feeds the same normalizer (`catalog.rs:40`), with attached-scope validation and a namespace test; rowid alias/AUTOINCREMENT rule (`schema.rs:302`); `LIMIT off,cnt` inversion (`syntax_lowering.rs:806`); SQLite parameter slots incl. `?N`, repeated names, gap filling (`analyzer.rs:41`); RETURNING/conflict/DROP/untyped/STRICT/keyword-column handling with named assertions in `tests/sqlite_compiler.rs`; `ProviderMetadata` as a separate document with the strict builder intact (`component.rs:195`, `requirement.rs:11`); `spawn_blocking` open plus interrupt-and-join `Drop` (`worker.rs:97`, `worker.rs:333`); two-row cardinality probe (`pool.rs:262`); common `MigrationRuntime`/`MigrationEngine` with baseline import, ledger, `BEGIN IMMEDIATE`, drift/checksum/rollback exercised end-to-end (`tests/migration_qualification.rs`); token-based scope guard (`migration_plan.rs:191`); real bundled probe recording version number and sorted compile options (`sqlite-runtime-probe.rs`, `run_sqlite_library_matrix.py`); `force = true` grammar pin plus checker assertion; artifact digest/size match the checked-in wasm (verified locally).

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_sql_sqlite_tools/src/migration_plan.rs:137` — the unowned-scope/one-statement guard runs only for `Ddl` steps; `SqlData`/`Backfill`/`Assertion` statements are only checked non-empty and are executed with `execute_batch` (`migration_runtime.rs:121`), so a hand-edited `graph.json` could carry `ATTACH`/`DETACH` there. Compiler-produced plans cannot, since the analyzer rejects those forms.
- suggestion: `crates/sifr_sql_sqlite_runtime/src/pool.rs:328` — the two-row probe is capped by `limits.max_collected_rows`, which `RuntimeLimits::validate` only requires to be non-zero; a profile with `max_collected_rows == 1` would return the first row instead of `ExpectedExactlyOneFoundMany`. PostgreSQL avoids this with an explicit `OverflowPolicy::Cardinality`.
- suggestion: `crates/sifr_sql_sqlite/src/lexer.rs:436` — `==` is not folded into a single operator token (SQLite's equality alias). Harmless today because `normalized_statement` is a fingerprint input, not executed SQL, but it makes two spellings of the same query normalize differently.
- suggestion: `crates/sifr_sql_sqlite/src/syntax_lowering.rs:225` — conflict-form detection scans keyword sequences over the whole statement, so `OR replace(...)`/`OR ignore(...)` inside an `INSERT ... SELECT` predicate can add a spurious `sql.sqlite.write.conflict` capability.
- suggestion: dead code retained behind `#[allow(dead_code)]` — `component.rs:597` (`schema_response_payload`) and `editor.rs:105` (`collect_tokens`).
- suggestion: `crates/sifr_sql_sqlite_runtime/tests/runtime_types.rs:360` — the corruption evidence writes a non-SQLite byte blob rather than a truncated or page-corrupted database, while the qualification surface is labelled `corruption-locking-performance`. Locking, recovery, and the performance budget are genuinely exercised.
- pre-existing issue: `crates/sifr_sql_contract/src/migration/compiler.rs:453` feeds provider `normalized_statement` into executable `SqlData`/`Assertion` steps, but provider normalizers mask string literals as `?literal` (SQLite `lexer.rs:189`, MySQL `lexer.rs:163`). Not introduced here and identical in the MySQL provider.

yaseralnajjar:
Final gate evidence for candidate `22a87c70d2b41bdb76ec4dc2ebc51ae13f17e25e`:

- Create-PR gate (`scripts/run_all_tests.sh --profile create-pr`): stopped before tests because the profile omits 12 required SQL-platform suites. It was not rerun.
- Merge gate (`scripts/run_all_tests.sh`): stopped before tests on the same profile-configuration error. It was not rerun.
- Ownership: Milestone 18 explicitly owns integrated qualification and profile wiring. This is the same known repository-profile contradiction recorded by Milestone 16, now including the completed SQLite suite.
- M17 focused SQL-provider validation passed 6/6 variants; contract/package tests, strict Clippy, formatting, diff hygiene, HIR maintainability, and file-size guards passed.
- Exact-SHA Opus remediation review: SATISFIED, no blockers.

PR #3643, opened 2026-08-31
fix(lowering): close receiver fact ownership gaps

yaseralnajjar:
## Summary

- classify subscript non-None facts by whether sequence growth can change their referent
- invalidate negative and otherwise potentially end-relative facts on append/extend while preserving absolute facts
- make affine dict.setdefault ownership rejection explicit at lowering and defensive at codegen
- encode the no-relevant-sequence-facts domain for mutable Python buffers and join sets

## Validation

- 1,103 lowering tests passed; 1 ignored
- 1,218 codegen tests passed
- 569 E2E fail fixtures passed
- focused release-native receiver/setdefault fixture passed
- full E2E: 718 passed; only unchanged Item 8 numeric_sentinels failed
- workspace Clippy, rustfmt, diff, HIR, file-size, audit inventory, and exact demo freshness passed

Exact candidate: `5b6c68d0508c5e79b0ddfc8d598480314ef8ef14`.

### DIFF
--- a/crates/sifr/tests/e2e/fail/python_buffer_setdefault_owner_conflict.sifr
+++ b/crates/sifr/tests/e2e/fail/python_buffer_setdefault_owner_conflict.sifr
@@ -1,0 +1,6 @@
+# expect-error: SIFR-PYZC-0001
+def select(
+    mut values: dict[str, python.Buffer[uint8]],
+    own fallback: python.Buffer[uint8],
+) -> python.Buffer[uint8]:
+    return values.setdefault("key", fallback)
--- a/crates/sifr/tests/e2e/pass/receiver_effect_setdefault_ownership.sifr
+++ b/crates/sifr/tests/e2e/pass/receiver_effect_setdefault_ownership.sifr
@@ -34,6 +34,10 @@ def copy_default(mut values: dict[str, bool], key: str, default: bool) -> bool:
     return values.setdefault(key, default)
 
 
+def existing_default(mut values: dict[str, str], default: str) -> str:
+    return values.setdefault("key", default)
+
+
 def main():
     numbers: list[int] = [0]
     assert read_after_growth_and_reorder(numbers) == 4
@@ -49,3 +53,6 @@ def main():
 
     flags: dict[str, bool] = {}
     assert copy_default(flags, "enabled", True)
+
+    existing: dict[str, str] = {"key": "stored"}
+    assert existing_default(existing, "unused") == "stored"
--- a/crates/sifr_codegen/src/methods/dict.rs
+++ b/crates/sifr_codegen/src/methods/dict.rs
@@ -27,7 +27,7 @@ fn render_key_arg_expr(arg: &RustExpr) -> RustExpr {
 }
 
 fn materialize_setdefault_storage_arg(ty: &Type, arg: &RustExpr) -> RustExpr {
-    if crate::helpers::is_copy_type_for_codegen(ty) || ty.contains_affine_resource() {
+    if crate::helpers::is_copy_type_for_codegen(ty) {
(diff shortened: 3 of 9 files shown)


yaseralnajjar:
Reviewed the exact base→candidate diff (`e4f5cff` → `5b6c68d`) read-only, plus surrounding guard/summary/codegen code.

What I verified against each criterion:
- **1 & 2:** `subscript_reference_stability` (`sequence_guard_detection.rs:274`) classifies only literal nonnegative indices as `StableAcrossGrowth`; `literal_int` resolves `USub`, so `-1` lands in `MayChangeOnGrowth`, and non-literal indices default to the conservative class. `clear_growth_sensitive_subscript_presence_guards_for_target` (`sequence_guards.rs:194`) retains stable facts and drops end-relative ones; `mutating_methods.rs:88` routes only list `append`/`extend` there, while `insert`/`appendleft`/`reverse`/`sort` still clear all presence facts and removals still clear everything. Tests cover both directions (`mutable_call_sequence_guard_tests.rs:124` preservation at index 0, `:137` rejection at `-1` for both methods).
- **3:** `method_type_collections.rs:321` is the single reachable checked contract, with a message naming both insertion transfer and the second-owner return conflict; codegen `lower_setdefault` (`dict.rs:240`) now rejects affine key/value defensively instead of bypassing materialization, and the removed `contains_affine_resource()` short-circuit in `materialize_setdefault_storage_arg` is no longer reachable for affine types. Fail fixture + pass fixture (`existing_default` covering the existing-value return path) back it.
- **4:** `receiver_fact_domain` is consulted first in the summary match; for `PythonBuffer` the only mutable-borrow method is `write` (`mutating_methods.rs:162`, `:279`), and every mutable-borrow `JoinSet` method is growth-only, so no falsifiable fact path exists today. Buffer behavior is identical to base; `fact_domain` is asserted in the summary test.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_lowering/src/lower/mutating_methods.rs:112` adds `JoinSet` to `NoRelevantSequenceFacts`, which disables the conservative `_ => Removal` catch-all for `__sifr_spawn_blocking`/`__sifr_spawn_cpu` and for any future mutable-borrow JoinSet method. Sound today because all such methods only grow the set, but this loosening is beyond what criterion 4 required, is untested, and would silently preserve facts if a removing JoinSet mutator is added later.
- suggestion: `crates/sifr_lowering/src/lower/mutating_methods.rs:110` carries no in-code rationale; the "no relevant sequence fact can target this receiver type" proof currently exists only as an enum name plus a test assertion, so a future reader cannot check the argument (buffer `write` only appends, buffer elements are non-optional, JoinSet is not subscriptable) without re-deriving it.
- suggestion: `crates/sifr_lowering/src/lower/sequence_guard_detection.rs:274` classifies stability purely from index-literal sign, so dict string-key presence facts and variable-index list facts are all labeled `MayChangeOnGrowth`. Harmless now (dict growth never reaches the growth-sensitive clearing), but it over-invalidates variable-index list facts on `append`/`extend` and would mislabel dict keys if the growth-sensitive path is ever extended to dicts.
- suggestion: `crates/sifr_lowering/src/lower/expressions/method_type_collections.rs:342` still includes `setdefault` in `requires_reusable_values`, so that affine branch is now unreachable for `setdefault`; dropping `setdefault` from that list would leave exactly one owner of the message.
- pre-existing issue: `crates/sifr_codegen/src/methods/dict.rs:240` returns `None` for affine key/value, which propagates as "method not lowered" through the emitter chain rather than an internal diagnostic. Unreachable given the lowering rejection, but the defensive gate is silent if the lowering check is ever weakened.

Issue #3667, opened 2026-09-02
Repair stale TypeVar diagnostic message assertions on main


yaseralnajjar:
Exact `origin/main` SHA `74bbb636744adaacb8c3eca09108b6fff9725357` fails two existing `sifr_lowering` unit tests after `066300ff185f38b425a884a2225b72990a194e58` changed TypeVar diagnostic wording without updating their exact assertions.

Failing tests:
- `lower::expressions_tests::control_flow_and_strings::test_typevar_invalid_bound_shape_has_primary_range` expects `TypeVar bound must be a simple type name`; the producer emits `TypeVar bound must be a type name`.
- `lower::expressions_tests::control_flow_and_strings::test_pep695_typevar_constraint_shape_has_primary_range` expects `TypeVar constraints must be simple type names`; the producer emits `TypeVar constraints must be type names`.

The code, diagnostic code, and primary ranges are otherwise unchanged. This is an upstream test/provenance regression found while validating emitted-Rust excellence Item 8; it is outside that item and reproduces on its exact base.

yaseralnajjar:
Observed the same two stale TypeVar assertions while qualifying Item12J in [PR #3699](https://github.com/sifr-lang/sifr/pull/3699), exact candidate `f720a342edd87004975355b478948f7eb5c8b406`, base `4ce05473f58716a611ac190581bf0737ba15331e`.

`cargo test -p sifr_lowering` reports 1114 passed, 2 failed, 1 existing ignored. The failures remain `test_typevar_invalid_bound_shape_has_primary_range` and `test_pep695_typevar_constraint_shape_has_primary_range`: assertions expect “simple type name(s)” while the producer emits “type name(s)”. Both test and producer files are unchanged from the exact base. Their sources contain no classes/error channels and do not exercise12J's repair. This is candidate execution plus unchanged-input provenance, not an independently repeated base run.

The failure remains owned here and was not repaired or waived by12J. Raw log: `/private/tmp/sifr-item12j.pT6Xkk/tests-sifr_lowering-f720a342e.log`; SHA-keyed evidence is preserved with PR #3699.

yaseralnajjar:
[...]

PR #3668, opened 2026-09-02
Complete canonical emitted Rust cleanup

yaseralnajjar:
Closes Item 8 of the Emitted Rust Excellence phase.

This change completes structured canonical emission cleanup, removes dead and identity constructs, makes demand and name handling deterministic, regenerates authoritative demos, and strengthens exact generated-code debt governance.

Qualification includes 723/723 E2E fixtures, package/codegen/runtime/lowering coverage, all 262 authoritative companions, selected full-corpus Clippy, workspace Clippy with warnings denied, rustfmt, determinism, panic scans, inventory, file-size/HIR guardrails, and exact-binary demo freshness. Two stale TypeVar assertion failures reproduced on exact main remain owned by #3667.

The required exact-SHA Opus review and the one-shot create-PR/merge gates will be recorded before merge.


yaseralnajjar:
# Verdict: **NOT SATISFIED**

Review record written to `/Users/yaseralnajjar/.claude/plans/perform-the-one-allowed-dreamy-river.md`. Working tree untouched (`git status --porcelain` empty before and after); no Git, PR, or external state changed.

## Blockers (in Item 8 scope, reproduced from the tree)

**B1 — the checked-in debt file fails its own self-test; the inventory gate fails in every profile.**
`verification/areas/generated_code_quality/data/generated_quality_debt.json` keeps `clippy::manual_let_else` and `clippy::unnecessary_semicolon` in `clippy.lint_owners` with no matching `clippy.entries` row. `quality_policy.py:313-321` computes `stale = set(lint_owners) - governed_lints` and is exercised unconditionally by the self-test at `quality_policy.py:588-595`, called from `inventory_gates.py:131`. I ran it read-only: `RuntimeError: Clippy debt contains stale lint owners: ['clippy::manual_let_else', 'clippy::unnecessary_semicolon']`. `inventory` is in smoke, representative and full (`runner.py:29,38,49`); create-PR→smoke, merge→representative. The record's claim that the generated inventory gate passed (record lines 364, 418) is false at this SHA.

**B2 — I8-30 is unimplemented.** Required: *"Key the closure by canonical function identity and prove same-basename functions remain distinct"* (record:568). `function_generic_bounds.rs:131-175` still keys `HashMap<String, …>` on bare `function.name`, and none of the six tests in `function_generic_bounds_tests.rs` concerns same-basename callables. The ledger marks it Implemented.

**B3 — I8-25's mechanism and its mandated regression are absent.** Required: *"Normalize the checked optional index/place before method-argument emission and add an isolated native regression"* (record:559). The only change is the fixture guard in `sliding_window_narrowing.sifr:26`, whose own header now says "keep generated code on the non-optional path for set.remove/add". No producer, no test for `Option<String>` into `HashSet::remove`. (I8-22's parallel fixture edit *is* legitimate — it supplies exactly the nonnegative-index fact `subscript_guards.rs` consumes.)

**B4 — Item-8-owned companion debt was re-owned, not removed.** 19 lint families / 124 diagnostics that the base file attributed to Item 8 are relabeled to Items 2/5/7/9/10/11 — `dead_code` 12, `unused_assignments` 12, `semicolon_if_nothing_returned` 22, `useless_conversion` 8, `unused_unit` 6, `missing_const_for_fn` 6, `branches_sharing_code` 4, `non_camel_case_types` 2, `unused_mut` 1 — i.e. literally Item 8's dead-construct, identity-transform and generated-name categories. Nothing validates an owner label (`quality_policy.py:384-396` only checks membership in `owner_items`), so Item 8 debt can be discharged by editing an integer. This contradicts the rule this commit itself added ("Item 12 must remove the remaining owned debt rather than rebase it"). The candidate's evidence statement ("removed only disappearing lint entries and signature-only drift with unchanged counts") is contradicted by the file: 25 new lint families and 19 owner reassignments.

Genuine credit: the retained corpus selection fell 35 982 → 4 196 diagnostics and 99 → 32 lints with zero Item-8-owned lints remaining; `rustfmt` and `safety` debt are truly empty; the companion selection is newly *measured*, so B4 is rebasing, not concealment.

## Ledger closure

**No — not every row is closed.** I8-25 and I8-30 fail outright. I8-03 and I8-17's "Qualified in the candidate" wording is falsified by B1. I8-13 (real producer is `render_expr_and_blocks.rs:693`, not the identifier policy), I8-18 (cites a pre-existing assertion), I8-06 (proves injectivity, asserts nothing about warnings) and I8-21 (no test comparing sync/async/statement-block renderings) have pointer or coverage weaknesses. I8-33 narrows the language beyond its quoted authorization (all nested generics, not only bounded ones), though no in-tree program regresses. Everything else I checked holds, including all five named `.sifr` fixtures and every other named test fn.

## New mechanism defects → later items, not a third review round

Canonicalizer rewrites lacking guards, none reproducible on today's corpus and none tested: `make_constant_unwrap_default_eager` drops `unwrap_or_else`'s closure without checking parameter usage (`idiom_cleanup.rs:707-724`, `expression_is_pure_initializer` accepts `Expr::Path` at `:285`); unused-binding wildcarding rewrites `let x = <init>` → `let _ = <init>` with no `Drop` check (`syntax_cleanup.rs:440-470`); `expression_is_discardable` deletes `Expr::Binary` initializers (`:634-646`). Separately, the GCQ producer fingerprint omits `stdlib/` and `crates/sifr_stdlib` while this commit changes two `stdlib/*.sifr` files — with a shared cache root, Clippy can run on stale materialization; that belongs to the verification area's owner.

## Gate readiness

**Not ready.** Running the one-shot create-PR gate now would consume it on a failure I can already reproduce locally. Fix B1, then B2/B3, adjudicate B4, and only then run create-PR followed by merge.

yaseralnajjar:
## Sole remediation review — exact SHA `a77acce704ccab8bf568ea4156ff05dd706c66c1`

I reviewed the exact diff `74bbb636…a77acce7` (582 paths, 4 commits), concentrating on the four blocker corrections and the named remediation mechanisms in `49f375e1`.

**What I verified**

- **Blocker 1/4 (Item 8 lint debt):** `generated_quality_debt.json` is now schema 2 with `clippy.owner_items = [2,5,7,9,10,11]` — no Item 8 owner remains; all 51 `lint_owners` values are later items and every owner is governed by at least one of the two selection entries (`selection-0902166e369de295`, `selection-16ae5ff4b2d3462f`); `rustfmt`, `safety`, and `freshness` entries are empty. `validate_debt_entry` now requires per-lint `count`/`signature_sha256` with an exact owner, and `run_debt_self_test` fails closed on ownerless lints, malformed signatures, stale owners, and companion-identity-insensitive merges. Elimination is real, not suppressed: **zero** `#[allow(` or `#[expect(` attributes exist across all 262 authoritative `emitted.rs` files.
- **Blocker 2 (callable identity):** `closed_function_type_param_bounds` keys requirements by `ModuleFunctionIdentity(index)` and re-projects names only at return; nested callables resolve through `LexicalScope` with `root/foo/bar` path identities, `shadowed_names`, and a visited set. The residual bare-name lookup map is safe by construction — `module_function_registry.rs:24` rejects duplicate module-level function names.
- **Blocker 3 (optional `set.remove`):** `resolve_set_method_type` now type-checks `remove`/`discard` arguments, accepting a compatible `optional_member_type` and emitting `TYPE_CONTAINER_ELEMENT_CONFLICT` otherwise, with three focused lowering regressions plus the new `set_remove_optional_index.sifr` fixture (native pass per evidence).
- **Remediation mechanisms:** branch-local shadowing (`references_after_statements` walks backwards through `let` shadows; `factor_shared_if_*` check condition/branch-binding disjointness); format captures extended to `assert`/`assert_eq`/`assert_ne`; duplicate enum owner names fail closed (`owner_counts … == Some(1)`, with a same-named nested-enum test); `format_generated_binary_project` iterates const discovery to a fixed point with an explicit non-monotonicity diagnostic and no cap; drop scopes preserved by replacing `local.pat = wildcard_pattern()` with `_name` renaming and narrowing `expression_is_discardable` to literals (with direct `Guard`/`clone`/division regressions); producer fingerprints add `sifr_stdlib` and `stdlib`; `materialize_binary_project_sources` still runs `try_generate_sysroot_dependency_plan` and writes the full Cargo project, skipping only compilation/linking; `run_strict_clippy` does a per-package `cargo clean` before collecting diagnostics for determinism.

Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- suggestion: `syntax_cleanup/idiom_cleanup/structured_control_cleanup.rs:98` (`factor_shared_if_suffix`) hoists token-identical trailing statements out of both branches after checking only name disjointness, so a shared statement now runs *after* branch-local bindings are dropped instead of before. Unreachable in the current corpus and narrower than the drop-point preservation the same commit added for unused bindings, but the guard should be effect-aware, not name-aware.
- suggestion: `ir_optimize/dead_bindings.rs:189` treats every `RustExpr::BinOp` with pure operands as pure, while the syn layer deliberately removed its `Expr::Binary` arm from `expression_is_discardable` for exactly the "unknown binary effect" reason. Deletion is confined to unread non-tail `__sifr_`-prefixed temporaries, so nothing fails today, but the two cleanup layers now disagree on the same rule.
- suggestion: `member_demand.rs:517` (`prune_unused_private_fields`) deletes struct-literal field initializers for undemanded private fields with no purity check on the initializer, and collects demand from everything except `Item::Mod`, so a side-effecting initializer for a never-read field would be dropped silently and a `super::`-referenced parent-scope field would look undemanded. Not observed in the 262-companion corpus.
- suggestion: several new rewrites are name-based rather than type-based — `remove_redundant_iterator_into_iter` strips `.into_iter()` after adapter *names* that `Option`/`Result` also define, `rewrite_typed_redundant_len_closures` substitutes `::std::vec::Vec::len` even behind `as_deref` (which yields `&[T]`), and `rewrite_none_comparison` assumes `== None` implies `Option`. Failures surface as non-compiling generated Rust rather than wrong behavior, but only for shapes the corpus exercises.
- suggestion: `collect_format_capture_names` is duplicated three times (`item_demand.rs`, `syntax_cleanup/identifier_collection.rs`, `syntax_cleanup/liveness.rs`) and none of the copies recognizes width/precision captures (`{:width$}`), which would make such a binding look unused. No `$` specs exist in emitted output today; consolidating to one implementation would keep the three liveness consumers from drifting.
- infrastructure issue: `source_quality_checks.py::run_strict_clippy` runs `cargo clean --package` inside the shared generated target before each of the 262 companions. It buys determinism but repeatedly invalidates cached units and is unsafe if two gate runs ever share that target concurrently.
- pre-existing issue: exact main retains the two stale TypeVar message assertions (#3667) and the stale attached-API fixture lock (#3669); both are recorded in the deferred-findings table with concrete owners and are correctly not absorbed here.

yaseralnajjar:
## Item 8 final qualification evidence

- Opus remediation review: SATISFIED on exact SHA `a77acce704ccab8bf568ea4156ff05dd706c66c1` with no blockers: https://github.com/sifr-lang/sifr/pull/3668#issuecomment-5523601034
- Sole create-PR gate on that SHA: every reached check passed until the stdlib native-intrinsic allowlist rejected the new `sifr_runtime::count_byte` root. Per explicit user adjudication, documentation-only commit `fa661c6eccd4c1fa3eb0092e3106ac4d44dddeda` records that root; the targeted guard then passed with `direct_runtime_roots=14`. The create-PR gate was not rerun.
- Sole merge gate on final SHA `fa661c6eccd4c1fa3eb0092e3106ac4d44dddeda`: all Item 8 guardrails passed, including demo freshness and the corrected intrinsic allowlist. It stopped only in the coverage/taxonomy matrix on unclassified SQL packages/targets and existing SQL milestone taxonomy. None of the reported failure paths changed in Item 8; this is the already-recorded Item 12 qualification owner. The merge gate will not be rerun.
- Opus's non-blocking second-review mechanism suggestions are assigned to later Item 8A issue #3670 under the no-third-review rule.

PR #3672, opened 2026-09-03
Harden emitted Rust cleanup proofs

### LINKED ISSUE #3670: Item 8A: Canonical cleanup effect and identity hardening
## Source

Sole remediation review for emitted-Rust excellence Item 8 on exact SHA `a77acce704ccab8bf568ea4156ff05dd706c66c1`:
https://github.com/sifr-lang/sifr/pull/3668#issuecomment-5523601034

The review verdict was SATISFIED with no blockers. Under the phase's no-third-review rule, these newly identified mechanism suggestions require a later sequential owner before Item 9.

## Required full solution

- Make shared `if` suffix factoring effect- and drop-order-aware, not only name-disjoint.
- Unify IR and syn-layer discardability so unknown binary effects cannot be deleted.
- Preserve side effects in pruned private-field initializers and include nested-module demand.
- Replace name-based iterator/len/None rewrites with type- or structure-proven transforms.
- Consolidate format-capture analysis and support width/precision captures such as `{:width$}`.
- Isolate or synchronize generated Clippy cleaning so concurrent gates cannot invalidate a shared target.

## Acceptance

Add focused regressions for every mechanism, preserve semantics and generated-code determinism, obtain the normal exact-SHA Opus review, run applicable gates under the phase rules, merge, and update the phase record before Item 9 starts.

yaseralnajjar:
Closes #3670.

## Scope

- preserve branch-local drop order during shared suffix factoring
- share a conservative discardability policy across Rust IR and syntax cleanup
- preserve private-field initializer effects and nested-module demand
- require typed or compiler-owned structural proof for iterator, length, and `None` rewrites
- consolidate format-capture parsing, including dynamic width and precision captures
- isolate generated Clippy runs in invocation-owned Cargo targets

## Validation

- `cargo test -p sifr_codegen --lib` (1,359 passed)
- `cargo test -p sifr -- --skip test_e2e_pass` (all non-ignored tests passed)
- demo companion regeneration and freshness check
- HIR maintainability and 900-line file-size guardrails
- generated-code inventory, full panic scan, full rustfmt scan, freshness, required-demo corpus, and required-demo determinism checks
- two concurrent strict Clippy processes used distinct Cargo targets and produced identical diagnostic counts/signatures
- full E2E attempt reached 720/724; its three shared regex build failures identified the `Option<String>` length regression fixed in this SHA, and the unchanged timeout cleanup case was timing-sensitive; a post-fix targeted run of all four affected fixtures passed 4/4

The phase-specific exact-SHA Opus review and the one-shot create-PR/merge gates will follow on the frozen candidate per the phase record.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7596,6 +7596,7 @@ dependencies = [
  "sifr_type_system",
  "static_assertions",
  "syn 3.0.4",
+ "unicode-ident",
 ]
 
 [[package]]
--- a/crates/sifr_codegen/Cargo.toml
+++ b/crates/sifr_codegen/Cargo.toml
@@ -18,6 +18,7 @@ syn = { workspace = true, features = ["printing", "visit-mut"] }
 prettyplease = "0.3.0"
 proc-macro2 = { version = "1.0.107", features = ["span-locations"] }
 quote = "1.0.45"
+unicode-ident = { workspace = true }
 bigdecimal = { workspace = true }
 rust_decimal = { workspace = true }
 
--- a/crates/sifr_codegen/src/discardability.rs
+++ b/crates/sifr_codegen/src/discardability.rs
@@ -1,0 +1,64 @@
+//! Shared, conservative policy for deleting an evaluated Rust expression.
+//!
+//! Both structured-IR optimization and post-render syntax cleanup use this
+//! module. An expression is discardable only when evaluating and immediately
+//! discarding it cannot call user code, panic, allocate, or observe `Drop`.
+
+use crate::{RustExpr, RustLiteral};
+
+pub(crate) fn rust_ir_expression_is_discardable(expression: &RustExpr) -> bool {
+    match expression {
+        RustExpr::Literal(
+            RustLiteral::Int(_)
+            | RustLiteral::Float(_)
+            | RustLiteral::Bool(_)
+            | RustLiteral::StaticStr(_)
+            | RustLiteral::Char(_)
+            | RustLiteral::Unit
+            | RustLiteral::None,
+        ) => true,
+        RustExpr::Paren(inner) => rust_ir_expression_is_discardable(inner),
+        RustExpr::Tuple(elements) | RustExpr::Array(elements) => {
+            elements.iter().all(rust_ir_expression_is_discardable)
+        }
+        RustExpr::Literal(RustLiteral::Str(_))
+        | RustExpr::Verbatim(_)
+        | RustExpr::Ident(_)
+        | RustExpr::Path(_)
+        | RustExpr::MethodCall { .. }
+        | RustExpr::FnCall { .. }
+        | RustExpr::MacroCall { .. }
+        | RustExpr::FormatMacro { .. }
+        | RustExpr::BinOp { .. }
+        | RustExpr::UnaryOp { .. }
+        | RustExpr::Field { .. }
+        | RustExpr::Index { .. }
+        | RustExpr::Slice { .. }
+        | RustExpr::Ref { .. }
+        | RustExpr::Deref(_)
+        | RustExpr::Clone(_)
+        | RustExpr::Cast { .. }
+        | RustExpr::Block { .. }
+        | RustExpr::If { .. }
+        | RustExpr::Match { .. }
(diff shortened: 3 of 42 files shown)


yaseralnajjar:
Reviewed the diff between `8ecc16d9` (base) and `46c95c86` (candidate) at the exact SHAs; no files were modified.

What I checked in depth: the shared-suffix hoist guard (`structured_control_cleanup.rs:128-160`) against branch-local `let`s, statement macros, and let-chain condition bindings; the unified discardability contract (`discardability.rs`) and both consumers (`dead_bindings.rs:44`, `syntax_cleanup.rs:710`); private-field pruning demand/effect/apply symmetry (`member_demand.rs:519-593`, `private_field_effects.rs`) including nested-module demand, `..rest`, `Self` owner resolution, and the new `visit_item_mod_mut` skip against the recursive `prune_unused_members_in_scope` call; the typed `Option` rewrites (`registry_option_none_query`, `lower_option_len`, `registry_iterable_to_owned_into_iter_arg_expr`) including receiver parenthesization in `render_expr_and_blocks.rs:591` and the `&Option<Vec<_>>` → `as_ref()` → `Vec::len` shape confirmed byte-identical in the unchanged `demos/union_ops/emitted.rs`; the single format-capture parser and all six liveness/rename consumers, including width/precision `$` scanning, `{{` escapes, positional args, letter fill chars, and rename ordering after `rename_tokens`; and every generated-Clippy call site receiving `cargo_target_dir` (`generated_code_quality.py:631,648,801`) with `run_id` = mode+time+pid and success cleanup.

Verdict: SATISFIED

Blocking findings:
- None.

Follow-up findings:
- pre-existing issue: `syntax_cleanup/idiom_cleanup/assignment_cleanup.rs:144` still discards a `let` initializer under a name-based purity rule (`X::new()`, `.to_string()`, `.len()`, `SifrInt::from`), i.e. a second deletion-adjacent contract outside `discardability.rs`. Not one of the two criterion-2 sites and untouched here; belongs in Item 12 residual debt.
- pre-existing issue: `format_capture::format_argument_index` covers the `format`/`print`/`write`/`assert` families only. A binding referenced solely by an implicit capture in `panic!`/`unreachable!`/`todo!` stays invisible to liveness. Latent only because generated code avoids panic macros; the old parsers had the same gap.
- pre-existing issue: struct literals inside macro token streams (e.g. `vec![Foo { field: … }]`) are invisible to `FieldDemandCollector`, `retain_effectful_initializers`, and `PrivateFieldPruner` alike, so definition and literal could diverge. This diff only narrows candidates (trivial-drop plus effect retention), so it is strictly safer than the base; no demo currently emits that shape.
- pre-existing issue: `lower_option_len`'s closure branch yields `String::len` (bytes) for `str | None`, while `Type::Str` uses `chars().count()`. The new test `optional_string_length_keeps_a_payload_compatible_callback` locks in the callback shape, not the char semantics. Outside the six criteria.
- pre-existing issue: `stmt_support_emitter/iterator_lowering.rs:631` (`is_iterator_like_expr_for_ir`) still classifies iterators by method name; the criterion-4 hardening landed only in the post-parse and length/None paths.
- infrastructure issue: each generated-Clippy gate invocation now owns `target/sifr_generated_code_quality/<run>/cargo-target`, so dependencies rebuild per invocation and the tree is preserved on failure. Worth watching against the AGENTS.md Cargo storage guidance; the shared-target path remains for non-Clippy commands.

yaseralnajjar:
One-shot gate evidence for exact candidate `46c95c86582761c9a1f4003577f97ae8fb723ead`:

- `scripts/run_all_tests.sh --profile create-pr` was run exactly once. All reached Item 8A guardrails and Rust interop passed; it stopped only in coverage-matrix readiness on the existing unclassified SQL packages/targets and stale SQL milestone taxonomy owned by Item 12.
- `scripts/run_all_tests.sh` was run exactly once. It produced the same bounded result: all reached Item 8A guardrails and Rust interop passed, then the unchanged Item 12 SQL coverage/taxonomy debt stopped the lane.
- Neither named gate was repeated, and neither reported an Item 8A candidate defect.

PR #3678, opened 2026-09-04
Disambiguate emitted character comparison states

### LINKED ISSUE #3676: Item 9A: Character comparison state disambiguation
## Source

Second and final exact-SHA remediation review for Item 9 in #3674 / PR #3675.

## Defect

The allocation-free character/string comparison path collapses two distinct states to Option::None:

- an out-of-range string index has no character;
- a present string comparand is empty or contains multiple characters.

As a result, out-of-range indexed character equality against an empty or multi-character string incorrectly evaluates to true, and inequality is inverted.

Candidate reproducer on 6ab6adc08f3ad253bcb4d1d080d5f2c5554cae70:

- probe("xyz", 10, "ab") prints eq; required result is other.
- probe("xyz", 10, "") prints eq; required result is other.

## Required full solution

- Preserve the allocation-free character comparison path.
- Keep index absence distinct from present-but-not-single-character strings.
- Cover both operand orders, equality and inequality, literal and variable comparands, optional strings, valid Unicode scalar comparisons, empty strings, multi-scalar strings, and out-of-range positive/negative indices.
- Preserve the existing optional-absence equality contract where both operands are genuinely absent.
- Add focused emitted-shape and native semantic regressions plus generated-quality budgets.
- Reconcile the unrelated compiler_safety demo source behavior drift noted by the Item 9 remediation review: restore the prior demo contract or move that coverage under an explicitly justified owner, then regenerate its companion.

## Process

This is the immediate later item created under the phase no-third-review rule. It receives its own exact-SHA review and normal one-shot gates after Item 9 is merged and recorded.

yaseralnajjar:
## Summary
- distinguish absent indexed characters from present empty or multi-scalar strings without allocation
- specialize literal comparisons to compile-time character states
- add emitted-shape and native semantic regressions, refresh affected demo companions, and restore the compiler-safety output contract

Closes #3676

## Validation
- 1,380 sifr_codegen tests passed
- all non-pass-fixture sifr tests passed
- E2E pass suite: 726/726, signature 11427061fe6b7498
- workspace Clippy, rustfmt, HIR maintainability, file-size, diff hygiene, demo freshness passed
- authoritative 262-companion checks passed individually; corrected compiler_safety companion passed strict recheck
- exact affected generated corpus, panic, rustfmt, and determinism checks passed
- full fresh corpus retains the Item 11-owned tinyvec 1.13.0 / Rust 1.98 dependency-resolution failure
- existing e2e-060 signed_zero_key missing_const_for_fn debt assigned to Item 12

### DIFF
--- a/crates/sifr/tests/e2e/pass/emitted_rust_item9a_character_comparison.sifr
+++ b/crates/sifr/tests/e2e/pass/emitted_rust_item9a_character_comparison.sifr
@@ -1,0 +1,50 @@
+def main():
+    text: str = "a🦀z"
+    crab: str = "🦀"
+    empty: str = ""
+    multiple: str = "ab"
+
+    # Valid Unicode scalar, literal and variable comparands, both operand orders.
+    assert text[1] == "🦀"
+    assert "🦀" == text[1]
+    assert text[1] == crab
+    assert crab == text[1]
+    assert not (text[1] != "🦀")
+    assert not (crab != text[1])
+
+    # Present empty and multi-scalar strings never equal a present character.
+    assert text[1] != empty
+    assert empty != text[1]
+    assert text[1] != multiple
+    assert multiple != text[1]
+    assert not (text[1] == "")
+    assert not ("ab" == text[1])
+
+    # Missing positive and negative indices remain distinct from present strings.
+    assert text[99] != empty
+    assert empty != text[99]
+    assert text[99] != multiple
+    assert multiple != text[99]
+    assert text[-99] != ""
+    assert "ab" != text[-99]
+    assert not (text[99] == "")
+    assert not ("ab" == text[-99])
+
+    # Checked string lookups exercise present and absent optional string states.
+    present_empty: list[str] = [""]
+    present_multiple: list[str] = ["ab"]
+    present_crab: list[str] = ["🦀"]
+    missing: list[str] = []
+
+    assert text[1] == present_crab[0]
+    assert present_crab[0] == text[1]
+    assert text[99] != present_empty[0]
+    assert present_empty[0] != text[99]
+    assert text[-99] != present_multiple[0]
+    assert present_multiple[0] != text[-99]
+
+    # Two genuinely absent optional values retain their equality contract.
+    assert text[99] == missing[0]
+    assert missing[0] == text[-99]
+    assert not (text[99] != missing[0])
+    assert not (missing[0] != text[-99])
--- a/crates/sifr_codegen/src/borrowed_string_compare.rs
+++ b/crates/sifr_codegen/src/borrowed_string_compare.rs
@@ -7,7 +7,11 @@ use crate::{
 
 enum StringCompareValue {
     Char(RustExpr),
-    Str { direct: RustExpr, view: RustExpr },
+    Str {
+        direct: RustExpr,
+        view: RustExpr,
+        comparison_state: Option<RustExpr>,
+    },
     OptionalStr(RustExpr),
 }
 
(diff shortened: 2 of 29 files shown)


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Notes on what I verified (read-only, exact SHAs `ea8112a25` → `9f311def5`):
- `crates/sifr_codegen/src/borrowed_string_compare.rs:382-449` — the three-state encoding is sound: absent indexed char / absent optional string → outer `None`; present empty or multi-scalar string → `Some(None)`; present single scalar → `Some(Some(c))`. `char_option_to_comparison_state` (`.map(Some)`) and `string_option_to_comparison_state` (`.map(closure)` over `chars()`/`next()`) produce matching `Option<Option<char>>` on both operand orders, so OOB-vs-empty and OOB-vs-multi are unequal, and `!=` is the exact inverse because the same `BinOp` carries `op` through (`:81-85`).
- Criterion 3 preserved: `Char`×`OptionalStr` now uses `.map` instead of `.and_then` on both sides, so two genuinely absent values still compare `None == None` → true, with no `String` allocation (`:64-71`, `:401-448`).
- No one-character allocation anywhere in the new path: literals are compile-time specialized to `Some(Some('c'))` / `Some(None)` (`:390-399`, escaped via `escape_default` at `render/render_expr_and_blocks.rs:480`), and runtime strings use borrowed `chars()`. Confirmed in regenerated companions, e.g. `demos/safe_edge_cases/emitted.rs` now emits `.map(Some) != Some(Some('-'))`.
- No panic-capable operations introduced: `chars()`, `next()`, `normalize_index_or_len`, `get(..).copied()` are all total.
- `Str`×`Str` and `Str`×`OptionalStr` arms are untouched (`:32-39`, `:72-79`), so no behavior drift for pure string equality; `comparison_state` is only consulted on the mixed char/string arms.
- Demo diff hygiene: the union of all demo changes is exactly rustfmt re-indentation, the new comparison shape, and the `compiler_safety` restoration — nothing unrelated leaked in.
- Criterion 4: `demos/compiler_safety/main.sifr:99-101` restores the output contract to `21` (matching pre-Item-9 `accf50125` behavior) while keeping `c.callback(c.value)` as an explicit discarded call so the callback field/`double` stay live; `demos/compiler_safety/emitted.rs:193-224` matches that source.
- Coverage: `crates/sifr/tests/e2e/pass/emitted_rust_item9a_character_comparison.sifr` covers valid Unicode scalar match, present empty/multi, positive and negative OOB, literal and variable comparands, optional lookups, both operand orders, `==`/`!=`, and two absent optionals; the shape test at `item9a_character_comparison_codegen_tests.rs:1-61` pins six `.map(Some)` sites, four runtime `chars()` states, literal specialization, and absence of `to_string`/`to_owned`/`chars().collect` allocation. Its `__sifr_cmp_*` assertions are in the raw-`generate_rust` namespace, so they are non-vacuous.
- `surface_inventory.json` e2e count/hash correctly moved 725 → 726 for the added fixture.

Follow-up findings:
- suggestion: a `str` variable typed `Type::LiteralStr(_)` still takes the runtime `chars()` state path (`borrowed_string_compare.rs:105-111` sets `comparison_state: None`), so a compile-time-known single-scalar binding misses the `Some(Some('c'))` specialization. Correctness is unaffected; only a missed optimization.
- suggestion: `demos/compiler_safety/main.sifr:99` now contains a value-discarding `c.callback(c.value)` purely to keep the callback field live after Item 9's dead-code tightening. The output contract is restored, but the demo source is not byte-identical to its pre-Item-9 form; worth a one-line comment in the demo explaining why the call exists so a future reader does not delete it.
- pre-existing issue: `emitted_rust_item9_complexity` triggering `clippy::missing_const_for_fn` on `signed_zero_key`, SQL coverage/taxonomy readiness debt, and tinyvec 1.13.0 under Rust 1.98 — all confirmed unrelated to this diff and owned by Items 11/12.
- infrastructure issue: an untracked/ignored stray file `crates/sifr/tests/e2e/pass/Untitled` (3 bytes, dated 2026-08-29) sits in the fixture directory. It does not affect the `*.sifr` inventory pattern, but it is clutter in a governed fixture root.

yaseralnajjar:
Final candidate `9f311def58ee809d55f8f12517775c6faedb082d` received the sole merge-gate run. The merge lane completed in 165.02s; all reached Item 9A-relevant guardrails passed, including demo emitted freshness, HIR maintainability, file-size (3,741 files), dependency direction, stdlib guards, verification runner self-tests, and Rust interop. It stopped only on the unchanged SQL coverage classification and verification-taxonomy readiness failures already owned by Item 12. Per the phase closure rules, the gate was not rerun and unrelated SQL work was not absorbed.

PR #3684, opened 2026-09-04
Preserve module-scoped builtin error identities

### LINKED ISSUE #3682: Item 10A: preserve module-scoped builtin error shadow identities
## Context

The sole remediation review for emitted-Rust excellence Item 10 found a new mechanism defect on exact SHA `0bb73783b2daf2d0f20b63cbe16407493d4d217a`. Per the phase rule, this is a later item and receives no third Item 10 review round.

Parent phase: #3680
Item 10 PR: #3681

## Defect

`ModuleSupportDemand::merge` unions user-defined builtin-error suppressions across project modules. Shared support rendering then suppresses a builtin globally when only one module shadows that name. Another module that genuinely references the builtin gets dangling generated Rust paths.

The same scenario exposes an older root defect: project nominal planning keys builtin error classes by bare name, so a module-local user error class that shadows a builtin can be silently replaced by the builtin nominal identity.

## Required full solution

- Preserve each user-defined error class under its exact module identity.
- Preserve the builtin error identity for other modules that reference it.
- Make late stdlib/runtime/file-derived builtin-error demand module-aware; never union per-module suppression into a crate-wide veto.
- Keep correct single-file behavior.
- Cover normal projects and generated test projects.
- Compile and run a fixture where one module defines `ValueError` while another uses builtin `ValueError`.
- Remove or wire the currently production-unused `referenced_error_classes_with_source` helper.
- Document or enforce the flat generated-support trait invariant noted by the bounded review.

## Workflow

This is immediate Item 10A. It requires one exact-SHA Opus review and at most one remediation review. Compiler changes require one create-PR and one merge gate on the final SHA.

yaseralnajjar:
Closes #3682

## Summary
- keep user-defined error shadows local while aggregating builtin support demand per project
- key shared builtin nominals by canonical sifr.builtin.* identity without replacing module-qualified user paths
- preserve local shadow definitions during project nominal relocation
- fail closed when generated support traits violate the flat top-level ownership invariant
- cover single-file, normal project, and generated test-project paths with a distinct local field shape

## Exact candidate
- base: 5017ddb9f73089febc461affe597350ba6d695ae
- candidate: c9d0fb34331c32fb90342debf1eea28a0c6ee7e1

## Validation
- cargo test -p sifr_codegen item10a_ -- --nocapture (5 passed)
- cargo test -p sifr_driver item10a_ -- --nocapture (2 passed, including project and test-project native compile/run)
- python3 scripts/check_file_size_guardrails.py (3,751 files passed)
- cargo fmt --all -- --check (passed)
- sole merge-profile gate reached and passed generated-demo freshness, HIR/file-size/ownership/dependency/resource/stdlib/driver/verification guardrails, and the complete Rust-interop area; it then stopped only on unchanged SQL coverage/taxonomy readiness debt already owned by Item 12. The gate was not repeated.

## Review
- Exact-SHA Opus review: SATISFIED, no blocking findings: https://github.com/sifr-lang/sifr/pull/3684#issuecomment-5538828920
- No remediation review required.
- Pre-existing fixture lock drift found by review is recorded separately in #3685.

### DIFF
--- a/crates/sifr_codegen/src/builtin_errors.rs
+++ b/crates/sifr_codegen/src/builtin_errors.rs
@@ -37,3 +37,9 @@ pub(crate) const BUILTIN_ERROR_CLASSES: &[&str] = &[
     "WorkerRuntimeError",
     "WorkerError",
 ];
+
+pub(crate) fn builtin_error_identity(name: &str) -> Option<String> {
+    BUILTIN_ERROR_CLASSES
+        .contains(&name)
+        .then(|| format!("sifr.builtin.{name}"))
+}
--- a/crates/sifr_codegen/src/generated_rust_canonicalizer/project_support_pruning.rs
+++ b/crates/sifr_codegen/src/generated_rust_canonicalizer/project_support_pruning.rs
@@ -50,7 +50,7 @@ pub(crate) fn prune_generated_support_for_consumers(
         consumer_roots.extend(crate::stdlib_filter::rust_source_required_trait_names(
             source,
             support_source,
-        ));
+        )?);
         let file = syn::parse_file(source)
             .map_err(|error| format!("failed to parse generated support consumer: {error}"))?;
         consumer_items.extend(file.items);
--- a/crates/sifr_codegen/src/lib.rs
+++ b/crates/sifr_codegen/src/lib.rs
@@ -5,7 +5,7 @@
 mod lib_modules_and_codegen;
 pub use lib_modules_and_codegen::*;
 mod builtin_errors;
-pub(crate) use builtin_errors::BUILTIN_ERROR_CLASSES;
+pub(crate) use builtin_errors::{BUILTIN_ERROR_CLASSES, builtin_error_identity};
 mod discardability;
 mod generated_dependency_metadata;
 mod generated_rust_canonicalizer;
--- a/crates/sifr_codegen/src/lib_codegen_tests.rs
+++ b/crates/sifr_codegen/src/lib_codegen_tests.rs
@@ -75,6 +75,8 @@ mod generic_inheritance_codegen_tests;
 #[cfg(test)]
 mod item10_support_assembly_codegen_tests;
 #[cfg(test)]
+mod item10a_error_identity_codegen_tests;
+#[cfg(test)]
 mod item8_canonical_codegen_tests;
 #[cfg(test)]
 mod item9_emitted_rust_quality_codegen_tests;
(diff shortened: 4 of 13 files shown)


yaseralnajjar:
Reviewed the exact diff at `c9d0fb34` against base `5017ddb9`, plus surrounding code (`project_stdlib_nominals.rs`, `stdlib_filter/relocation.rs`, `generated_visibility.rs`, `union_type_helpers.rs`, `lib_project_codegen.rs`). Ran two suites not covered by the supplied evidence: `cargo test -p sifr_codegen` (1407 passed, 0 failed) and `cargo test -p sifr_driver` (573 passed, 1 failed — pre-existing, see follow-ups).

Disclosure: while diagnosing that failure I ran `cargo metadata` inside the fixture directory, which rewrote `verification/.../non_string_leaf_rejected/Cargo.lock`. I restored it with `git checkout --` on that single path; `git status` is clean again and matches `c9d0fb34`.

Key checks: builtin identities now key on `sifr.builtin.<Name>` on both the registration side (`register_builtin`) and the lookup side (`project_nominal_path`), with no bare-name registration left; `merge_project_module` drops shadow propagation only, and the shadow set is still honored in the single-file `render_support` path; `relocate_project_stdlib_nominals_owned_by` excludes local class names from both stripping and import emission; the nested-trait rejection matches the existing flat-support assumption already baked into `crate_visible_generated_support_source`; `referenced_error_classes_with_source` has no remaining callers and its imports (`collect_source_builtin_error_classes`, `BUILTIN_ERROR_CLASSES`) are still used elsewhere in the file.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `tests::attached_api_codegen::non_string_leaf_negative_is_package_compilable` fails because `verification/areas/core_language/fixtures/static_class_adapter/negative/non_string_leaf_rejected/Cargo.lock` is missing the `memchr` entry, so `cargo metadata --locked` refuses. The `memchr` dependency was introduced in `84ebe95b9`, an ancestor of the base SHA; nothing in this diff touches manifests or fixtures. Owns its own issue, not Item 10A.
- infrastructure issue: the supplied evidence covers only `item10a_` tests, file-size guardrails, and `fmt`. `cargo clippy --workspace -- -D warnings` and the E2E pass suite are not represented, and this change alters two broadly used paths (project-level support suppression and `rust_source_required_trait_names`). The PR/merge gate should run on this candidate before merge.
- suggestion: `rust_source_required_trait_names` failure handling is inconsistent — `project_support_pruning.rs:53` propagates with `?`, while the four call sites in `lib_project_codegen.rs`/`lib_test_project_codegen.rs` convert to `panic!`. Uniform propagation would keep the invariant enforcement without a compiler abort.
- suggestion: `union_type_helpers.rs:14-19` treats any identity-less class whose name is in `BUILTIN_ERROR_CLASSES` as the builtin. A user error class that reached a project union without an identity would resolve to the shared builtin path. Behavior is unchanged from before (bare-name keying had the same effect), and the new explicit `sifr.builtin.` key makes the assumption visible; a debug assertion there would close it.

PR #3687, opened 2026-09-04
Close emitted Rust excellence item 11

### LINKED ISSUE #3685: Refresh static class-adapter negative fixture lockfile
## Defect

The exact-SHA Item 10A review found a pre-existing fixture failure outside Item 10A: tests::attached_api_codegen::non_string_leaf_negative_is_package_compilable fails because verification/areas/core_language/fixtures/static_class_adapter/negative/non_string_leaf_rejected/Cargo.lock lacks the memchr entry required by the existing dependency graph. As a result, cargo metadata --locked refuses the fixture.

## Provenance

Observed while reviewing candidate c9d0fb34331c32fb90342debf1eea28a0c6ee7e1 against base 5017ddb9f73089febc461affe597350ba6d695ae. The dependency was introduced in ancestor 84ebe95b9; Item 10A does not change compiler manifests, lockfiles, or that fixture.

## Required outcome

Refresh and commit the fixture lockfile through its owning workflow, then prove the locked metadata/package-compilability test passes. Do not broaden Item 10A.

yaseralnajjar:
## Summary
- separate ephemeral local Cargo resolution from portable source-only artifacts
- pin emitted manifests and locks to exact Git/registry sources and resolve normal projects against the compatible sysroot vendor inventory
- preserve direct process argv boundaries, reject generated path escapes, and remove panic-on-proof conversions from user-controlled insert/split/replace/repeat paths
- refresh the static class-adapter negative fixture lock (closes #3685)

## Focused evidence
- cargo test -p sifr_codegen item11_ -- --nocapture
- cargo test -p sifr_driver item11_ -- --nocapture
- cargo test -p sifr_stdlib --features process item11_ -- --nocapture
- cargo test -p sifr_driver tests::attached_api_codegen::non_string_leaf_negative_is_package_compilable -- --exact --nocapture
- cargo test -p sifr --test e2e item11_materialized_project_is_locked_portable_and_relocatable -- --nocapture
- cargo fmt --check
- python3 scripts/check_sifr_driver_maintainability_guardrails.py

Exact candidate: 4e24b61bb0c06315afb51d92b42f3ae19e364c68

### DIFF
--- a/crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs
+++ b/crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs
@@ -242,6 +242,11 @@ def repeat_text(text: str, count: int) -> str:
 
 def main():
     assert repeat_text("ab", 3) == "ababab"
+    text = "a,b,c"
+    repeated = text * 2
+    assert text == "a,b,c"
+    assert repeated == "a,b,ca,b,c"
+    assert text.split(",", 1) == ["a", "b,c"]
 "#;
     let first = tempfile::tempdir().expect("first materialization root");
     let report = sifr_driver::materialize_single_file(
--- a/crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs
+++ b/crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs
@@ -235,6 +235,73 @@ fn missing_metadata_is_not_repaired_from_generated_rust() {
 }
 
 #[test]
+fn item11_materialized_project_is_locked_portable_and_relocatable() {
+    let source = r#"
+def repeat_text(text: str, count: int) -> str:
+    return text * count
+
+def main():
+    assert repeat_text("ab", 3) == "ababab"
+"#;
+    let first = tempfile::tempdir().expect("first materialization root");
+    let report = sifr_driver::materialize_single_file(
+        source,
+        Path::new("item11_portable.sifr"),
+        first.path(),
+    )
+    .unwrap_or_else(|errors| {
+        panic!(
+            "portable materialization should succeed: {}",
+            errors
+                .iter()
+                .map(|error| error.message.as_str())
+                .collect::<Vec<_>>()
+                .join("; ")
+        )
+    });
+    let project = report.project_path();
+    let manifest = std::fs::read_to_string(project.join("Cargo.toml"))
+        .expect("portable manifest should be readable");
+    let lock = std::fs::read_to_string(project.join("Cargo.lock"))
+        .expect("portable lock should be readable");
+    let worktree = env!("CARGO_MANIFEST_DIR");
+    assert!(!manifest.contains(worktree), "{manifest}");
+    assert!(!lock.contains(worktree), "{lock}");
+    assert!(manifest.contains("git = \"https://github.com/sifr-lang/sifr.git\""));
+
+    let relocated = tempfile::tempdir().expect("relocation root");
+    let relocated_project = relocated.path().join("artifact");
+    copy_dir_recursive(project, &relocated_project);
+    let metadata = Command::new("cargo")
+        .args(["check", "--locked"])
+        .current_dir(&relocated_project)
+        .output()
+        .expect("relocated Cargo check should execute");
+    assert!(
+        metadata.status.success(),
+        "relocated Cargo check should remain locked: {}",
+        String::from_utf8_lossy(&metadata.stderr)
+    );
+}
+
+fn copy_dir_recursive(source: &Path, destination: &Path) {
+    std::fs::create_dir_all(destination).expect("destination should be created");
+    for entry in std::fs::read_dir(source).expect("source directory should be readable") {
+        let entry = entry.expect("source entry should be readable");
+        let target = destination.join(entry.file_name());
+        if entry
+            .file_type()
+            .expect("entry type should be readable")
+            .is_dir()
+        {
+            copy_dir_recursive(&entry.path(), &target);
+        } else {
+            std::fs::copy(entry.path(), target).expect("artifact file should relocate");
+        }
+    }
+}
+
+#[test]
 fn fixtures_with_different_production_plans_cannot_repair_each_other() {
     let missing = case_with_plan(
         "missing_metadata",
--- a/crates/sifr_codegen/src/borrowed_string_compare.rs
+++ b/crates/sifr_codegen/src/borrowed_string_compare.rs
@@ -321,10 +321,7 @@ impl RustEmitter {
             expr: Some(Box::new(Self::as_str_option(RustExpr::MethodCall {
                 receiver: Box::new(RustExpr::Ident("__sifr_cmp_list".to_string())),
                 method: "get".to_string(),
-                args: vec![RustExpr::Cast {
-                    expr: Box::new(RustExpr::Ident("__sifr_cmp_norm".to_string())),
-                    ty: RustType::Named("usize".to_string()),
-                }],
+                args: vec![RustExpr::Ident("__sifr_cmp_norm".to_string())],
             }))),
         }))
     }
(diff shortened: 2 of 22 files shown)


yaseralnajjar:
**Verdict: NOT SATISFIED**

Reviewed worktree HEAD `4e24b61bb74ee768a09e168b91bb058bbafb7260` (the request's candidate SHA `4e24b61bb0c0…` does not exist; the short prefix `4e24b61bb` matches HEAD, base `cea27c0d9` confirmed).

Blocking findings:

- **regression**: `crates/sifr_codegen/src/methods/string.rs:277-289` and `:333-345` (`lower_split` via `exact_int_to_bound_expr`), criterion 4. Every two-argument `str.split(sep, maxsplit)` now emits `(<count> + 1).clamp_slice_bound(<len>+1)`, whose receiver type has no inference anchor, so the generated crate fails to compile. Reproduced on HEAD with `text.split(",", 1)`: `error[E0282]: type annotations needed --> src/main.rs:11:13 (SifrInt::from_i64(1) + 1).clamp_slice_bound(text.len().saturating_add(1usize))`; base emitted `::sifr_runtime::to_usize_proven(&(SifrInt::from_i64(1) + 1))` and compiled (existing e2e fixture output in `crates/sifr/target/sifr_e2e_cache/.../main.rs:2974,825`). Correction: anchor the sum's type before clamping — e.g. bind `let __sifr_maxsplit: SifrInt = <count> + SifrInt::from_i64(1);` (or add an explicit `SifrInt` RHS/annotation) and call `clamp_slice_bound` on that binding.

- **regression**: `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_binop.rs:71-77`, criterion 4. The new `let __sifr_repeat_src = <string_expr>;` binds by value, so `s * n` moves an owned `String` operand that the previous `(s).repeat(n)` only borrowed. Reproduced on HEAD with `s: str = "ab"; t: str = s * 2; print(s + t)`: `error[E0382]: borrow of moved value: 's' … value moved here: let sifr_generated_repeat_src = s;`. Correction: bind a borrow with an explicit target type, e.g. `let __sifr_repeat_src: &str = &(<string_expr>);` and `push_str(__sifr_repeat_src)`, so both `String` and `&str` operands coerce without moving.

Follow-up findings:

- **infrastructure issue**: the listed evidence contains no E2E pass suite or merge-gate run; both regressions above are covered by fixtures already present in the e2e cache (`splitn`, string-multiply cases), so the focused `item11_` tests were not a sufficient gate for a change to shared codegen lowering. Re-run `verification/runner/e2e/run_e2e_pass.sh` after the fixes.
- **suggestion**: `crates/sifr_driver/src/build/entrypoint_resolution.rs:39-48` — with `entrypoint: None` the policy is now `Locked` unconditionally, and `prepare_cargo_resolution` hard-errors ("locked Rust interop Cargo resolution has no authoritative package or sysroot lockfile") when `stdlib.interop.sysroot` is `None`. Every single-file/project build depends on that lock existing; worth an explicit diagnostic or a covered assertion for the sysroot-less configuration.
- **suggestion**: `crates/sifr_codegen/src/methods/string.rs:8-18` — `replacement_or_split_limit(object)` re-emits the receiver expression a second time. Safe today because receivers are rendered idents, but it is a latent double-evaluation hazard if receiver lowering ever yields a call.
- **suggestion**: the new repeat loop in `stmt_expr_binop.rs` builds the output without `String::with_capacity`/`Vec::with_capacity`, losing the allocation pre-sizing that `.repeat`/`iter::repeat().take()` had; a bounded reserve based on the clamped count would keep the previous cost profile.
- **suggestion**: `crates/sifr_driver/src/build/portable_project.rs:22-60` — `publish_portable_project` replaces only `src/`, `Cargo.toml`, and `Cargo.lock`; unrelated files already present at a reused `project_path` survive republication and are not covered by the portability assertion.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

What I verified on the exact remediation diff (`4e24b61bb..78c28c1e4`):

**Blocker 1 (E0282 on `maxsplit + 1`)** — `crates/sifr_codegen/src/methods/string.rs:284` and `:346` now emit `SifrInt::from_i64(1)` as the addition RHS in both arms of the 2-arg `split` lowering (whitespace/`None` pattern and explicit pattern). With `impl Add for SifrInt` / `Add<SifrInt> for &SifrInt` (`crates/sifr_runtime/src/int_ops.rs:83,105`) the operand type is now exact, so the previously ambiguous untyped-literal candidate set (`Add<i8..usize>`, `int_ops.rs:437`) no longer applies; the result feeds `clamp_slice_bound(&self, len: usize) -> usize` (`crates/sifr_runtime/src/int.rs:434`), so the bound path stays checked and criterion 4 holds. The renderer's SifrInt coercions leave an `FnCall` RHS untouched, so no double wrapping.

**Blocker 2 (moved owned `String` in repetition)** — `crates/sifr_codegen/src/stmt_support_emitter/stmt_expr_binop.rs:72-79,148-152` binds `let __sifr_repeat_src: &str = &(<src>);` and passes the ident directly to `push_str`. The source is borrowed, not moved, so a live owned `String` survives the expression; the borrow's region ends at the block, so even `text = text * n` shapes stay borrowck-clean, and `&str`/`&String`/temporary sources all coerce (temporary lifetime extension covers rvalue sources). `RustType::Named("&str")` renders verbatim (`render_paths.rs:8` returns names without `::` unchanged) and passes `ir_validate.rs:459`.

**Regression scan** — the only other assertion on this emitted shape (`crates/sifr_codegen/src/python_buffer_codegen_tests.rs:93`) matches on the identifier name only, and the list/bytes repeat branch (still `Clone`-based) is untouched. Item 11's other mechanisms (ERQ-030 portable manifest/lock, ERQ-032 argv boundaries, #3685 fixture lock) are not touched by this diff.

Follow-up findings:
- suggestion: `stmt_expr_binop.rs:75` models the annotation as `RustType::Named("&str")` although the IR has a structured `RustType::Ref { mutable, inner }` (`rust_ir.rs:317`); the structured variant would be the consistent representation. No behavioral difference — do not fix in this item.
- suggestion: `crates/sifr/tests/e2e_support/dependency_plan_authority_tests.rs:238` only `cargo check`s the relocated artifact, so the added `assert` lines (owned-String reuse, `split(",", 1)` result) are compile-only there; their runtime semantics rest on the existing E2E pass fixtures (`crates/sifr/tests/e2e/pass/collection_methods.sifr:54`, `cpython_core_object_model_subset.sifr:64`), which the single merge-profile gate will exercise.
- suggestion: `item11_portable_secure_codegen_tests.rs:38` asserts the loose substring `"+ SifrInt::from_i64(1)"`, which is weaker than anchoring on the `splitn(...)` bound expression; it would still pass if the exact-int anchor moved elsewhere in the file.

No new mechanism defect identified, so nothing needs recording as a later item on this account.

yaseralnajjar:
Merge-profile gate on exact candidate `78c28c1e4c42bd85d685d3a3cffdf132fcdfcc40` stopped at the generated-demo freshness guardrail after Cargo cache setup, HIR maintainability, and the 3,753-file size guardrail passed. The guardrail reported 15 stale companions: additional_modules, bisect, config_json_csv, container_methods, ergonomics, file_streams, glob, io, ordered_collections, stdlib, stdlib_ownership, structured_parsing_serialization, subscript_assignment, tempfiles_and_zip, and text_and_patterns. This is an in-scope candidate defect. Per phase rule 10 and the user's one-gate/no-second-gate instruction, the gate was not rerun and the PR remains unmerged pending explicit adjudication.

yaseralnajjar:
Superseded by Item 11A integration PR #3689. The preserved reviewed Item 11 candidate was integrated, all 15 stale companions were regenerated by that compiler, exact-SHA Opus review was SATISFIED, and Item 11A’s sole merge-profile gate was consumed.

PR #3695, opened 2026-09-05
fix(verification): close Item 12G DLPack dependency path

yaseralnajjar:
The Python dependency audit still constructed the removed DLPack demo path and failed with FileNotFoundError before checking dependency versions. Item 12G selects `demos/python_dlpack` and adds four regressions for computed project paths, authoritative manifest/lock reads, missing inputs, and the original stale-path mutation. Exact version and artifact-hash enforcement is unchanged.

Carries the authorized orchestration amendment and dependency validation mapping, preserving both retained Item12B checkpoints and main's naming-cleanup findings. No Item12H–12K implementation is included.

Candidate: `1cb24bdd088bddf42077f6e42112e53bba7c3562`; base: `b475ebdcd37081aa2860d9c348ace4100b546eff`.

Validation: four focused unittests passed; named `python_interop:dependency-versions` suite passed (one variant, seven negative mutations, two projects, 19 packages, two locks, two images); canonical file-size guardrail passed (3,754 files). Commands were recorded in the owner issue before testing. Explicit user rules omit Sifr gates for these runner/test/docs-only changes. The [single exact-SHA Opus review](https://github.com/sifr-lang/sifr/pull/3695#issuecomment-5554835685) returned SATISFIED with no blockers; no remediation review was needed.

### DIFF
--- a/plans/issues/active/ad-hoc-emitted-rust-excellence.md
+++ b/plans/issues/active/ad-hoc-emitted-rust-excellence.md
@@ -188,7 +188,15 @@ It does not broaden the active item.
 | 10A | merged | Module-scoped builtin error shadow identities | Project support demand preserves user-defined and builtin error identities per module, without crate-wide suppression or dangling generated paths. |
 | 11 | merged | Portable and secure generated projects | Reviewed candidate `78c28c1e4c42bd85d685d3a3cffdf132fcdfcc40` is preserved and merged through Item 11A after its consumed gate's stale companions were regenerated. |
 | 11A | merged | Generated-companion freshness and Item 11 integration | The reviewed Item 11 candidate and all 15 compiler-regenerated companions are merged through a separately bounded review and gate without rerunning Item 11's consumed gate. |
-| 12 | pending | Residual semantic completion and full-corpus qualification | Finish remaining semantic/profile work, remove all governed generated-code debt, regenerate every owned surface, and pass the uncompromising final qualification and applicable one-shot gates. |
+| 12 | blocked: external algorithmic corpus | Residual semantic completion and full-corpus qualification | Finish remaining semantic/profile work, remove all governed generated-code debt, regenerate every owned surface, and pass the uncompromising final qualification and applicable one-shot gates. |
+| 12B | blocked: Python qualification dependencies | Bounded algorithmic dependency repair | Both reviews passed. Preserve the approved candidate and both failed gates until Items 12G–12J and integration Item 12K resolve qualification. |
+| 12C | incorporated into 12B | Builtin-registration Clippy blocker | No independent item, review, or gate remains. |
+| 12D | recorded, not started | Native corpus emission dependencies | Adjudicate checked-read control flow and the complete native diagnostic inventory before Item 12B closure. |
+| 12G | authorized: next worker | Dependency-checker demo path identity | Repair and qualify the stale DLPack project reference. |
+| 12H | authorized: after 12G handoff | Project-wide generated-field identity | Repair declaration/consumer naming consistency across generated files. |
+| 12I | authorized: after 12H handoff | Macro-defined project support visibility | Repair cancellation task-local visibility without blanket exports. |
+| 12J | authorized: after 12I handoff | Async Python error-channel contract | Resolve the authoritative error contracts and preserve async semantics. |
+| 12K | authorized: integration after dependency qualification | Item 12B and Python dependency integration | Qualify the integrated candidate and merge the preserved work; do not reset Item 12B review history. |
 | 12A | pending | Phase closure and whole-phase review | Review the fully merged phase once, reconcile architecture/roadmap/evidence, and archive only when no actionable row remains. |
 
 ## Item Acceptance Contracts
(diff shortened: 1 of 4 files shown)


yaseralnajjar:
Item 12G exact-candidate validation

Base: b475ebdcd37081aa2860d9c348ace4100b546eff
Candidate: 1cb24bdd088bddf42077f6e42112e53bba7c3562
Worktree: /tmp/sifr-item12g.B8fCer/sifr
Date: 2026-09-05

Commands registered in the owner issue before execution:

1. `uv run --project verification --locked python -m unittest discover -s verification/areas/python_interop/runner -p test_dependency_versions.py -v`
   PASS: 4 tests in 0.050 seconds, Python 3.14.7. Computed project paths,
   observed authoritative input reads, both missing inputs, and obsolete
   concatenated-path mutation covered. Initial sandbox attempt stopped at uv
   cache access before any tests; authorized escalation completed successfully.
2. `uv run --project verification --locked python -m sifr_verify areas run --area python_interop --suite dependency-versions`
   PASS: 1 variant, 0 failures; 2 projects, 19 packages, 2 locks, 2 images,
   all 7 existing negative mutations. Report:
   `/tmp/sifr-item12g.B8fCer/sifr/target/verification/areas/python-interop-results.json`.
3. `python3 scripts/check_file_size_guardrails.py`
   PASS: 3754 files, limit 900 lines.

Diff hygiene passed; candidate worktree is clean. No compiler, lockfile,
fixture, or workflow changes. Create-PR and merge-profile gates omitted by
explicit user direction. No Item12B gate or review was repeated.

The phase record was three-way reconciled from common ancestor
2dc4165fd9e7c34432a9b0d098188dc645aaca55, latest main, and the parent's
authorized records. Both main naming findings and retained Item12 checkpoints
are preserved. Parent worktree, stash, and retained Item12B worktree are untouched.

## Initial and only Opus review

Reviewed at exact base `b475ebdcd` → candidate `1cb24bdd0`, worktree clean before and after (read-only; ran with `-B` to avoid bytecode writes).

**What the diff does**

- `dependency_versions.py:46` — `DLPACK_PROJECT_ROOT` changes from the concatenated obsolete `demos/m12_dlpack_demo` to `Path("demos/python_dlpack")`. That constant feeds only `PROJECT_PATHS["dlpack-demo"]`; nothing else in the module or repo references the old root (confirmed: remaining hits are historical `plans/reviews/active/m12-dlpack-*.md` records, outside scope).
- `test_dependency_versions.py` (new, 72 lines) — four regressions over the runtime-computed paths.
- Two `plans/issues/active/*.md` records (authorized doc carry).

**Independently verified**

- `python3 -m unittest discover -s verification/areas/python_interop/runner -p test_dependency_versions.py` — 4/4 pass, using the exact command registered in `ad-hoc-python-interop-qualification-dependencies.md`.
- `dependency_versions.py --self-test` — `projects=2 packages=19 locks=2 images=2 mutations=7`, exit 0. All seven pre-existing negative mutations (version, artifact hash, direct-dependency, retired package, service image) are retained unmodified.
- `demos/python_dlpack/{pyproject.toml,uv.lock}` exist; the pyproject declares exactly `numpy` + `torch`, matching `EXPECTED_PROJECTS["dlpack-demo"]`.
- No fallback, compatibility directory, suppression, `try/except` around the demo reads, or lockfile/version edit appears anywhere in the diff. `validate_repository` still lets `load_toml`'s `FileNotFoundError` propagate, and the third test pins that fail-closed behavior for each demo input.
- `verification_taxonomy.py` passes. The new test's negative case correctly reuses the repo's `"m" + "12_dlpack_demo"` concatenation idiom (required because `verification/` is a taxonomy `ACTIVE_ROOT`, while `plans/` is not — so the plain literal in the issue record is also legitimate).
- File-size guardrail: both touched Python files well under 900 lines.

The fourth test (`test_obsolete_computed_demo_reference_is_rejected`) is the one that closes the actual gap in scope — it patches `PROJECT_PATHS` back to the stale path and asserts `FileNotFoundError.filename` is the old absolute path, so a future textual rename that reintroduces a broken computed path fails loudly instead of silently.

Verdict: SATISFIED

Blocking findings:
- None

Follow-up findings:
- infrastructure issue: `verification/areas/python_interop/manifest.json` registers only `dependency_versions.py` as the `dependency-versions` case; the new `test_dependency_versions.py` is not a manifest case, so it runs only when someone invokes the recorded `unittest discover` command. This matches the existing precedent (`verification/areas/sysroot_release/test_attached_api_certification.py` is likewise unreferenced anywhere in the repo), so it is a pre-existing area-wide gap rather than something this candidate introduced — but the regression is not continuously enforced.
- suggestion: `test_dependency_versions.py:9` does `import dependency_versions` with no `sys.path` setup, relying on the runner directory being the discovery start dir. The sibling precedent file inserts its area root into `sys.path` before importing. Adding the same two lines would make the regression robust under pytest or a different start dir.
- suggestion: `plans/issues/active/ad-hoc-emitted-rust-excellence.md` still lists Item 12G as `authorized: next worker`. Per `AGENTS.md` records are updated after the item merges, so this is expected at candidate time — just flagging it as the outstanding post-merge record edit.

PR #3697, opened 2026-09-05
Fix project-wide generated field identity (12H)

yaseralnajjar:
Item12H implements project-wide generated-field identity through one module-qualified nominal registry, including binary/test projects and generated Rust bridge declarations/consumers. It covers imported literals, patterns, member accesses, owner-local field collisions and module-identity rejection without a PythonError-specific or untyped global substitution. Bridge sources participate in artifact cache identity; materialization does not run another field-renaming pass.

**Disposition: blocked, draft/unmerged.** Reviewed implementation SHA: `9b52ac20094608c8a31f252db99e49ef7c963384`; base: `4ce05473f58716a611ac190581bf0737ba15331e`. Subsequent changes are owner/phase records only.

The [sole remediation Opus review](https://github.com/sifr-lang/sifr/pull/3697#issuecomment-5555345800) is SATISFIED with no blockers. Exact-SHA focused tests pass3/3; final-source driver tests pass581 with77 existing ignored; unchanged-codegen canonicalizer evidence passes115 tests. Native binding execution succeeds, and all264 demo companions pass freshness in the gate. Formatting, HIR and file-size guardrails pass.

The one exact-SHA merge-profile gate failed on the existing SQL coverage registry:9 unclassified packages,13 unclassified targets and1 stale library classification. Later gate stages were not reached. Strict Clippy and the clean-environment binding-authoring suite also remain blocked by the separately recorded12B/12C and Python build/verification dependencies. No qualification was weakened and no failed receipt is a pass.

[Complete validation, review and gate evidence](https://github.com/sifr-lang/sifr/pull/3697#issuecomment-5555393502). The phase, Python dependency issue and SQL coverage owner contain the terminal handoff and concrete12K dependency. Two reviews and one merge gate used; no third review, second gate or create-pr gate. No12I/12J code was implemented. Preserve this candidate for qualified12K integration; do not merge it unqualified.

### DIFF
--- a/crates/sifr_codegen/src/generated_rust_canonicalizer.rs
+++ b/crates/sifr_codegen/src/generated_rust_canonicalizer.rs
@@ -1,5 +1,5 @@
 use quote::ToTokens;
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use syn::visit::{self, Visit};
 use syn::visit_mut::{self, VisitMut};
 
@@ -27,7 +27,6 @@ pub use api_cleanup::{
     discover_project_const_function_names,
     finalize_formatted_generated_rust_source_with_project_consts,
 };
-use field_name_cleanup::canonicalize_generated_field_names;
 pub use identifier_canonicalizer::canonicalize_generated_rust_identifier;
 #[cfg(test)]
 use item_dependencies::IdentifierCollector;
@@ -48,9 +47,37 @@ use syntax_cleanup::canonicalize_syntax;
 /// are escaped too, keeping the mapping injective when user code deliberately uses
 /// a canonical prefix.
 pub fn canonicalize_generated_rust_source(source: &str) -> Result<String, String> {
+    let mut sources = canonicalize_generated_rust_project(&BTreeMap::from([(
+        String::new(),
+        source.to_string(),
+    )]))?;
+    sources
+        .remove("")
+        .ok_or_else(|| "missing canonical crate root".to_string())
+}
+
+/// Canonicalize all physical modules together before per-file cleanup. Keys use
+/// Rust module paths (`a::b`); the empty key is the crate root.
+pub fn canonicalize_generated_rust_project(
+    sources: &BTreeMap<String, String>,
+) -> Result<BTreeMap<String, String>, String> {
+    let fields = field_name_cleanup::canonicalize_fields(sources)?;
+    let names = identifier_canonicalizer::project_name_map(fields.values().map(String::as_str))?;
+    fields
+        .into_iter()
+        .map(|(module, source)| {
+            canonicalize_source_with_names(&source, &names).map(|source| (module, source))
+        })
+        .collect()
+}
+
+fn canonicalize_source_with_names(
+    source: &str,
+    names: &BTreeMap<String, String>,
+) -> Result<String, String> {
     let structurally_pruned = prune_closed_generated_binary(source)?;
     let source = structurally_pruned.as_deref().unwrap_or(source);
-    let mut canonical = identifier_canonicalizer::canonicalize_identifiers(source)?;
+    let mut canonical = identifier_canonicalizer::canonicalize_identifiers(source, names)?;
     for _ in 0..16 {
         let rewritten = rewrite_format_captures(&canonical)?;
         let structurally_pruned = prune_closed_generated_binary(&rewritten)?;
@@ -162,7 +189,7 @@ impl<'ast> Visit<'ast> for FallibleControlUse {
 fn rewrite_format_captures(source: &str) -> Result<String, String> {
     let mut file = syn::parse_file(source)
         .map_err(|error| format!("failed to parse canonical generated Rust: {error}"))?;
-    let field_names_changed = canonicalize_generated_field_names(&mut file);
+    let shorthand_changed = field_name_cleanup::compact_shorthand(&mut file);
     let syntax_changed = canonicalize_syntax_to_fixed_point(&mut file)?;
     let final_syntax = prettyplease::unparse(&file);
     let mut api_file = syn::parse_file(&final_syntax)
@@ -170,7 +197,7 @@ fn rewrite_format_captures(source: &str) -> Result<String, String> {
     let before_api = api_file.to_token_stream().to_string();
     improve_generated_api_items(&mut api_file.items, &final_syntax);
     let api_changed = api_file.to_token_stream().to_string() != before_api;
-    if !field_names_changed && !syntax_changed && !api_changed {
+    if !shorthand_changed && !syntax_changed && !api_changed {
         return Ok(source.to_string());
     }
     let first_api_source = prettyplease::unparse(&api_file);
--- a/crates/sifr_codegen/src/generated_rust_canonicalizer/field_name_cleanup.rs
+++ b/crates/sifr_codegen/src/generated_rust_canonicalizer/field_name_cleanup.rs
@@ -1,197 +1,139 @@
+//! Field spelling belongs to a nominal declaration, not a file's identifier set.
+use quote::ToTokens;
 use std::collections::{BTreeMap, BTreeSet};
-use syn::visit::{self, Visit};
 use syn::visit_mut::{self, VisitMut};
 
-const GENERATED_FIELD_PREFIX: &str = "sifr_generated_";
+mod registry;
+mod rewrite;
+use registry::Registry;
+use rewrite::Rewriter;
 
-/// Remove the compiler namespace from generated field members.
-///
-/// Fields already have a type-owned namespace, so retaining the global generated
-/// prefix is redundant and can make every member of a struct share one prefix.
-/// The rewrite only touches field declarations and member positions; local values
-/// keep their independently canonicalized names.
-pub(super) fn canonicalize_generated_field_names(file: &mut syn::File) -> bool {
-    let mut collector = FieldNameCollector::default();
-    collector.visit_file(file);
-    let names = canonical_field_name_map(&collector.names);
-    if names.is_empty() {
-        return false;
+#[cfg(test)]
+mod tests;
+
+pub(super) fn canonicalize_fields(
+    sources: &BTreeMap<String, String>,
+) -> Result<BTreeMap<String, String>, String> {
+    let mut files = sources
+        .iter()
+        .map(|(module, source)| {
+            syn::parse_file(source)
+                .map(|file| (module.clone(), file))
+                .map_err(|error| {
+                    format!("failed to parse assembled generated Rust: module {module}: {error}")
+                })
+        })
+        .collect::<Result<BTreeMap<_, _>, _>>()?;
+    let registry = Registry::collect(&files);
+    let mut result = BTreeMap::new();
+    for (module, file) in &mut files {
+        let before = file.to_token_stream().to_string();
+        let mut rewriter = Rewriter::new(&registry, module);
+        rewriter.visit_file_mut(file);
+        if let Some(error) = rewriter.error {
+            return Err(error);
+        }
+        let source = if before == file.to_token_stream().to_string() {
+            sources[module].clone()
+        } else {
+            prettyplease::unparse(file)
+        };
+        result.insert(module.clone(), source);
     }
-    FieldNameCanonicalizer { names }.visit_file_mut(file);
+    Ok(result)
+}
+
+fn field_names(fields: &syn::Fields) -> BTreeMap<String, String> {
+    let names = fields
+        .iter()
+        .filter_map(|field| field.ident.as_ref().map(ToString::to_string))
+        .collect::<BTreeSet<_>>();
+    let mut occupied = names
+        .iter()
+        .filter(|name| !name.starts_with('_'))
+        .cloned()
+        .collect::<BTreeSet<_>>();
+    let mut result = BTreeMap::new();
+    for name in names.iter().filter(|name| name.starts_with('_')) {
+        let significant = name.trim_start_matches('_');
+        let base = significant.strip_prefix("sifr_").unwrap_or(significant);
+        let mut candidate = if base.is_empty() {
+            "underscore".to_string()
+        } else {
+            base.to_string()
+        };
+        if syn::parse_str::<syn::Ident>(&candidate).is_err() {
+            let raw = format!("r#{candidate}");
+            candidate = if syn::parse_str::<syn::Ident>(&raw).is_ok() {
+                raw
+            } else {
+                format!("{candidate}_field")
+            };
+        }
+        while occupied.contains(&candidate) {
+            candidate.push_str("_field");
+        }
+        occupied.insert(candidate.clone());
+        result.insert(name.clone(), candidate);
+    }
+    result
+}
+
+fn rename(member: &mut syn::Member, names: &BTreeMap<String, String>) -> bool {
+    let syn::Member::Named(identifier) = member else {
+        return false;
+    };
+    let Some(name) = names.get(&identifier.to_string()) else {
+        return false;
+    };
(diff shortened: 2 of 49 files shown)


yaseralnajjar:
I reviewed the exact diff `4ce05473f..405e3d3c2` (mechanism, driver wiring, tests, regenerated companions, records).

**Verdict: NOT SATISFIED**

The core mechanism is sound and matches the criterion for the paths it covers. `field_name_cleanup` now resolves each field to a module-qualified nominal owner before identifier cleanup (`registry.rs` for nominal/alias/re-export/glob/type-alias/generic resolution, `rewrite.rs` for typed receivers, literals, patterns, loop/closure/match scoping), members are removed from the identifier namespace (`identifier_canonicalizer.rs:169`, `:197`), shorthand is expanded/re-compacted so value bindings keep their own identity, per-owner collisions are injective and keyword-safe, external fields are preserved, unknown receivers fail with a diagnostic, and there is no PythonError special case. The 18 regenerated companions contain only field renames plus reflow, and the records are honest about the blocked binding-authoring suite.

**Blocking findings:**

- **in-scope omission | `crates/sifr_driver/src/build/materialize.rs:232` and `:269-276`** (with `crates/sifr_driver/src/build/project_codegen.rs:77`) — criterion: "repair the shared generated-field naming mechanism with canonical nominal/module identity across declarations [and] imported consumers … Cover binary-project … paths". The generated `__sifr_bridge` module family is a binary-project generated Rust file family, but it is produced *after* `canonicalize_project_fields` runs and is never entered into the shared registry. `generated_bridge_sources` (`rust_interop_bridge_sources.rs:95-113`) emits `pub <field.name>` verbatim from Sifr class attributes (`rust_interop_bridge_contract/generated_types.rs:39-41`, `name: field_name.clone()`, no underscore filter), and each bridge file is then formatted through `write_project_file` → `format_generated_rust` → `canonicalize_generated_rust_source`, which builds a *single-file* registry and renames the declaration. Failure scenario: a bridged record/error class with a private attribute (e.g. `_name`, or an error carrying `__sifr_python_error`) yields `__sifr_bridge/<m>.rs` declaring `pub name: …` while `main.rs`/support modules — canonicalized in the project pass where `crate::__sifr_bridge::<m>::<T>Bridge` resolves to `Ty::Named` outside `registry.fields` — keep `_name` in the struct literal and member access. `Rewriter::member` (`rewrite.rs:1065-1088`) only raises the diagnostic for `Ty::Unknown`, so this diverges silently (fail-open), reproducing the same E0560/E0609 class the item exists to eliminate. It also contradicts the record added in `internal_docs/architecture.md:402-408` ("Binary and test-project materialization use the same registry"). Correction: generate the bridge sources before formatting and pass them into the same `canonicalize_generated_rust_project` call as `main.rs`/support modules, keyed `__sifr_bridge` and `__sifr_bridge::<module>`, then write them without a second independent single-file canonicalization; add a focused regression with an underscore-prefixed bridged record field. (Note honestly: the divergence existed before this item in a different spelling — `sifr_generated_name` vs `name` — so this is not a regression introduced here, but it is the in-scope defect class left uncovered on a binary-project path.)

**Follow-up findings:**

- suggestion: `rewrite.rs:1308-1326` — macros whose body is not a comma-separated expression list (and not `vec![x; n]`) are skipped entirely, so member accesses inside them are never renamed and never diagnosed. Generated code today only emits `println!`/`write!`/`format!`/`format_args!`/`assert*`/`vec!`/`task_local!`, so this is unreachable now; it becomes a silent E0609 if a future emitter adds a macro with pattern or item syntax.
- suggestion: `rewrite.rs:827-962` — the fail-closed diagnostic aborts the whole build whenever a receiver is `Ty::Unknown` and the member name collides with any project generated field. Inference does not model casts, `unsafe`/`async`/`loop` blocks, or iterator adapters outside the fixed list (`filter`, `chain`, `collect`, … all yield `Unknown`), so ordinary future emitter changes can turn a working build into a hard diagnostic. Worth broadening coverage or narrowing the trigger.
- suggestion: `registry.rs:411-419` / `rewrite.rs:849` — `Ty::inner()` is `argument(0)`, which is the key type for map-like receivers, so `map[key]._x` resolves to the wrong owner. Harmless today because both owners map through the same rename, but it can pick a wrong collision suffix.
- suggestion: `crates/sifr_driver/src/build/project_codegen.rs:77-85` — the full project canonicalization now runs once over the project and then again per file inside `format_generated_rust`. It is idempotent (confirmed by the stable companion output), but it doubles canonicalization work on every build.
- infrastructure issue: the `binding-authoring` suite still stops at `runner/binding_authoring.py:362` on clean-environment bytecode state; per the evidence this is a PyO3 interpreter-launcher issue owned by Python build/verification and recorded for 12K. Passing this review is not a claim that suite passed.
- pre-existing issue: the two list-repeat expectations in `generate_rust_with_metadata` and the `project_stdlib_nominals.rs:45` `expect_used` Clippy failure (Item 12C/12B) remain, correctly recorded for 12K rather than absorbed here.

yaseralnajjar:
## Verdict: SATISFIED

I inspected the full `4ce05473f..9b52ac200` diff and the remediation diff `405e3d3c2..9b52ac200` directly.

**Prior blocking finding is resolved at the mechanism level.** Bridge sources are now produced inside `format_generated_binary_project` (`crates/sifr_driver/src/build/project_codegen.rs:78`) with identities `__sifr_bridge` / `__sifr_bridge::<module>`, and are chained into the *same* `canonicalize_project_fields` registry call and every subsequent formatting/const-discovery loop as main and support modules. `generated_bridge_sources` now emits raw field names (`rust_interop_bridge_sources.rs:97`), so declaration spelling is decided solely by the shared registry — the divergence path is gone, not masked.

Supporting checks I confirmed rather than assumed:

- **Single funnel.** `format_generated_binary_project` is reached from exactly one place (`entrypoint.rs:713`), which every `GeneratedBinaryProject` passes through, so `bridge_modules` can't be empty-by-omission at materialization.
- **No second identity pass.** `format_generated_rust` is now `#[cfg(test)]`; `write_project_file` calls `format_canonical_generated_rust` (rustfmt + finalize only). The two identifier-bearing strings that materialization still synthesizes — the root `pub mod` declaration (`materialize.rs:260`) and namespace `pub mod` lines (`materialize.rs:287`) — are explicitly canonicalized, matching `canonical_rust_module_path`'s per-component canonicalization. `__sifr_bridge` → `sifr_generated_bridge` under `canonical_identifier_candidate`, consistent with the file path the test reads back.
- **Collision rejection is real**, not advisory: `rust_formatter.rs:20` errors on duplicate module identity instead of overwriting the source map, and the negative test asserts the exact message.
- **Cache identity** includes bridge sources (`materialize.rs:551`), with a direct mutate-and-compare assertion.
- **The new regression test is genuinely end-to-end**: two modules with the same type basename, record + error bridges, `_name`/`name` owner-local collision, imported consumer, destructuring pattern and member access, then `rustc --edition 2024` on the actually-materialized `src/main.rs` and execution of the binary.
- **No PythonError rule, no untyped global substitution**: `registry.rs` resolution is module-qualified and type-driven; `rewrite.rs:263` renames only through a resolved nominal owner's map, and an unresolved (`Ty::Unknown`) receiver whose name matches any registered generated field raises a compiler diagnostic rather than guessing.
- **Companion deltas are an improvement, not drift.** The five files differing from the initial review lose `_user_<hex>` suffixes (e.g. `sifr_generated_chars_text_user_736966…` → `sifr_generated_chars_text`), which were artifacts of the now-removed duplicate canonicalization. Injectivity is preserved: a genuine user identifier colliding with a canonical prefix is still escaped by the single pass.
- **F3 is latent, not present.** I checked specifically because a live map-receiver defect would be a regression, not a suggestion: `Expr::Index → inner()` takes the first generic argument, but no `x[k].field` index-then-member form occurs in any of the 264 emitted companions, so the wrong-argument selection is unreachable in current emission. Classification as deferred maintenance is accurate.
- Contract-test file changes are confined to initializing `bridge_modules: Default::default()`; no async/callback behavior touched.

No blocking findings.

## Follow-up findings

- **pre-existing issue** — A bridge module literally named `mod` materializes to `__sifr_bridge/mod.rs` and collides with the bridge root file. The new duplicate check keys on *module identity* (`__sifr_bridge` vs `__sifr_bridge::mod` are distinct), not on materialized path, so the collision still resolves by silent overwrite; the candidate only changes which side loses relative to base. Broken identically before this item; worth a path-level identity check whenever bridge layout is next touched.
- **suggestion** — `sifr_codegen::canonicalize_generated_rust_source` and `rust_formatter::format_generated_rust` now have no production caller; ~100 canonicalizer tests exercise the single-source wrapper only. Coverage remains meaningful (it delegates to the project entry point with a lone root key), but the wrapper is now test-scaffolding and should be consolidated rather than left as a parallel API.
- **infrastructure issue** — Qualification is genuinely incomplete and the record states this accurately: `clippy-remediation.log` stops at the inherited 12C `expect_used` failure and is **not** passing lint evidence; the binding-authoring suite stops at the clean-environment bytecode assertion (`runner/binding_authoring.py:362`) with the PyO3 `-B` cause suspected, not proven; and no merge-profile or create-PR gate has run. The single reserved merge-profile gate on this SHA is the remaining requirement before merge — in particular it is what would exercise E2E compilation of the demo corpus, which the emit-only refresh does not.

yaseralnajjar:
# Item12H validation — 9b52ac20094608c8a31f252db99e49ef7c963384

Base: 4ce05473f58716a611ac190581bf0737ba15331e (freshly fetched main).
PR: https://github.com/sifr-lang/sifr/pull/3697 (draft; not qualified to merge).
Owned branch: codex/emitted-rust-excellence-item-12h.
Owned worktree: /tmp/sifr-item12h.afJDbk/sifr.
All paths below are under /tmp/sifr-item12h.afJDbk/.

## Passing mechanism evidence

- Exact candidate: `cargo test -p sifr_driver project_field_identity`: 3 passed. `driver-field-9b52ac20094608c8a31f252db99e49ef7c963384.log`. Includes compiled/executed native bridge record/error regression, owner-local collisions, imported consumer/pattern/member accesses, and duplicate-module rejection.
- Final unchanged source inputs: `cargo test -p sifr_driver`: 581 passed, 77 existing ignored. `driver-remediation-final.log`. Includes native test-project integrations.
- Codegen source unchanged since initial review candidate405e3d3c2: 115 canonicalizer tests passed, including nine focused field tests and positive rustc compilations. `canonicalizer-405e3d3c2adcf018044a2f733ac64ec942f01967.log`.
- `cargo build --locked -p sifr`: passed on final source. `build-remediation-final.log`.
- Native original binding regression, using rebuilt final-source CLI in the generated binding-authoring project: `target/debug/sifr run src/main.sifr --frozen` prints `binding runtime ok`. `native-remediation-final.log`. This is not full-suite qualification.
- Compiler-owned refresh: `python3 scripts/check_demo_emitted_freshness.py --sifr target/debug/sifr --update`: all264 emissions succeeded. `companion-refresh-remediation-final.log`. 21 companions differ from base; no demo Sifr/reference sources changed. The intermediate bridge-listing expansion is absent.
- Exact candidate `cargo fmt --check`, `python3 scripts/check_file_size_guardrails.py` (3758 files), `python3 scripts/check_hir_maintainability_guardrails.py`, and `git diff --check`: passed.

## Qualification blockers — not passes

- Strict affected-crate Clippy stops at unchanged `crates/sifr_codegen/src/project_stdlib_nominals.rs:45`, `expect_used`. `clippy-remediation.log`. This is known Item12C incorporated in preserved Item12B;12K must integrate its repair. This does not establish that driver Clippy completed.
- Full codegen tests reached two unchanged list-repeat expectations upstream of canonicalization; source-path provenance, not an independent base reproduction. `crate-tests-repair.log`. The own parse-diagnostic regression was corrected and covered by final canonicalizer evidence.12K must reconcile inherited tests with preserved12B.
- Named binding-authoring suite passed the original native case and subsequent frozen checks, then stopped on initially clean environment bytecode-state equality at `binding_authoring.py:362`. Remaining assertions were not reached. `binding-authoring-error-flow.log`, preserved `binding-authoring-blocked-results.json` (SHA25691d61c65448fd3475301a0efdece28a6c74373e9d71b55c896ce0ba9a81b8de2). Python build/verification owns this12K input. PyO3 startup is suspected, not independently proven as the cause. No warm retry replaces clean-environment qualification.

## Review and gate accounting

Initial exact-SHA Opus review of405e3d3c2 was NOT SATISFIED for late bridge generation; the candidate above remediates it. Published initial review: https://github.com/sifr-lang/sifr/pull/3697#issuecomment-5555203238.
The sole remediation review returned SATISFIED with no blocking findings: https://github.com/sifr-lang/sifr/pull/3697#issuecomment-5555345800. Raw response: opus-9b52ac20094608c8a31f252db99e49ef7c963384.odYXxZ/response.md. Two completed reviews total; no third review.

The single merge-profile gate on this exact clean SHA FAILED (exit1,362.20s). Cargo setup, all reached guardrails including all264 demo freshness checks, and Rust interop (10 variants,0 failures) passed. Coverage-matrix readiness then failed with9 unclassified SQL packages,13 unclassified targets, and1 stale PostgreSQL lib classification. The other3 coverage variants passed. Later Python-area, crate, and E2E stages were not reached; this is not passing full qualification.

Owner: plans/issues/active/ad-hoc-schema-first-sql-platform-review-follow-ups.md, existing Coverage registry blocker section. That issue already records the same failure on2026-09-05; no SQL source, Cargo target, or coverage classification changed in12H.12K integration needs this owner repair as well as the previously recorded12B/12C and Python build/verification inputs.

Exact gate evidence: merge-9b52ac20094608c8a31f252db99e49ef7c963384.log and .json; coverage-matrix-9b52ac20094608c8a31f252db99e49ef7c963384.json; rust-interop-9b52ac20094608c8a31f252db99e49ef7c963384.json. Reports copied outside target to preserve them.

Disposition: preserve the reviewed implementation, leave PR3697 draft/unmerged, update owner/phase records only, then stop. One merge gate used, no second gate; zero create-pr gates. No12I/12J code was implemented. The review's pre-existing bridge module-to-path collision and API-consolidation suggestions are later maintenance, not changes in this candidate.

Changed paths: `changed-paths-9b52ac20094608c8a31f252db99e49ef7c963384.txt` (48:7 codegen,17 driver,21 generated companions,3 docs).

yaseralnajjar:
B15 final qualification records a recurrence of the existing approved-but-unmerged
H/I/correctingM1 dependencies, not a new repair item. Exact B15candidate
`875a3555a7d1bcd7885ad4151c75a4a7abf17a74`, PR3746, naturally failed its sole
continuation merge gate after2833.76s: Python30variants/26PASS/4blockingFAIL.
Failures are binding-authoring E0560 (12H/PR3697), three callback E0425 cases
(12I/PR3698), async declaration/context SIFR-RESULT-0003 (correctingM1/PR3700).
All earlier13guards,92preparation graphs, Rust10, integratedSQLreadiness4,
core5/CPython2 PASS; remaining15areas/full-crate/E2E/stdlib UNREACHED.

Read-only Git ancestry verifies approvedH `9b52ac20094608c8a31f252db99e49ef7c963384`,
I `f6e8afd964bb214a44c50271dcb2014ee8e828b4`, and correctingM1
`d726ffc11258c49f0185fd2d49697988cf90972c` already exist in preserved fullB13
`4eef8a2bb4dbc24fb7c1d1213652be379047f4b1`. B13/#3744 needs B14capture/#3745,
while standaloneB15 needs those stack repairs: a deliverycycle for parent
assessment, not duplicated implementation or another standaloneB15gate.
No owner/source approval or historical limit is reset; nothing is merged.

[Complete B15 terminal record and evidence](https://github.com/sifr-lang/sifr/blob/44cbbd0b126a63f705d7295cd763802ca0a83959/plans/issues/active/ad-hoc-emitted-rust-excellence.md).
Gate log SHA256 `34a4bd772f4a459b2f5ffa8cfedc538504149279744ba7e53559a8c6a20e4076`.
Record-only update; B15owner stops, no next-item work.

PR #3760, opened 2026-09-07
Fix YAML checkout ownership validation (item 37)

yaseralnajjar:
Item 37 closes checkout-discovery bypasses in the submodule ownership guardrail. It parses actual YAML job steps with the existing Ruby/Psych dependency, rejects malformed or ambiguous workflow structure, and permits nonrecursive governance/evidence checkouts only when their workflow, job, condition, and full input identity match an explicit owner.

Validation passed: `python3 scripts/check_submodule_ownership.py --self-test`, `python3 scripts/check_submodule_ownership.py`, `python3 scripts/check_file_size_guardrails.py` (3,759 files), and `git diff --check`. Self-tests cover named/unnamed/commented and quoted/flow/folded checkout forms, neighboring-step and run-text bypasses, YAML aliases, malformed inputs, and classification mutations.

Only `scripts/check_submodule_ownership.py` changes. This is the runner-only scope of Item 37 in `plans/issues/active/ad-hoc-latest-stable-release-convergence.md`; compiler, workflow, lockfile, fixture, and gitlink inputs are unchanged, so the authorized continuation policy requires no create-PR or merge-profile gate. Exact-SHA Opus review is pending.

### DIFF
--- a/scripts/check_submodule_ownership.py
+++ b/scripts/check_submodule_ownership.py
@@ -5,8 +5,9 @@ from __future__ import annotations
 
 import argparse
 import configparser
-import re
-import tempfile
+import json
+import subprocess
+import textwrap
 from dataclasses import dataclass
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


yaseralnajjar:
**Verdict: SATISFIED**

Blocking findings:
- none

Evidence I re-derived at the exact SHAs (base `b8639d913` → candidate `f011d8e01`, one file, +262/−50, `scripts/check_submodule_ownership.py` only, 559 lines):

- **The bypass is real and is closed.** Replaying the base regex pair on the candidate tree: `release-publication.yml` yielded **0** checkout blocks (indentation-based `WORKFLOW_STEP_RE` swallowed the whole steps region into earlier `- ` items, so five real `actions/checkout` steps were invisible), while `release-publication-prepare.yml` yielded 5 merged blocks. The YAML-parsed checker now enumerates every step per job and correctly flags the four non-recursive `release-publication.yml` checkouts unless they match a classification.
- **Classifications match the workflows byte-for-byte.** The four `release-publication.yml` entries correspond exactly to lines 153–158, 159–163, 172–178, 179–185 (job `publish`, exact `if`, full input identity incl. `fetch-depth: 0`/`persist-credentials: false`); the recursive `stable-source` checkout (164–171) is accepted by the recursive rule, not by classification, so editor-source ownership is unchanged. The two prepare entries match lines 293–300 and 302–309 (job `prepare`, exact `${{ ... }}` conditions).
- **Parser choice is in-scope and precedented.** `ruby -e` + Psych `safe_load(permitted_classes: [], permitted_symbols: [], aliases: true)` mirrors `verification/areas/distribution_release/cases/*workflow*` usage; no Python dependency added. Single-document, scalar-key, duplicate-key and tag rejection all fail closed with a guardrail failure rather than a skip.
- **Adversarial probes beyond the self-test all reject:** same input identity in another workflow, in another job, with one extra `with:` key, `submodules` supplied via job `defaults`, and `uses: "actions/checkout@v7 "` (trailing space) — each produces a failure. Merge-key/anchor-supplied `submodules: recursive` is accepted, correctly, because it is the effective value.
- **Executed on this head:** `python3 scripts/check_submodule_ownership.py --self-test` → PASS; `python3 scripts/check_submodule_ownership.py` → PASS; `git diff --check b8639d913 f011d8e01` clean; no line >100 chars; no workflow, gitlink, lockfile, fixture or compiler change. Guardrail lanes are `ubuntu-24.04` (`profile_runner.py:188`), where `ruby` is preinstalled; the matrix macOS/Windows jobs do not run this guardrail.

Follow-up findings:
- suggestion: `run_workflow_self_tests` spawns roughly 70 separate `ruby` processes; wall clock went from effectively instant to ~50 s locally (measured). Batching the fixtures through one Ruby invocation, or caching the interpreter start, would keep the guardrail lane cheap as classifications grow.
- suggestion: `identity` drops the `submodules` key, so a classified governance/evidence checkout could later set `submodules: true` (non-recursive init) and still be accepted. Harmless today — it is strictly more initialization than the classified intent — but pinning `submodules` absence in the classified identity would make the intent enforceable.
- pre-existing issue: the two `release-publication-prepare.yml` classifications are currently unreachable, because both steps already set `submodules: recursive` and short-circuit before the classification check. They are exercised only synthetically by the self-test and would take effect only if `recursive` were removed from those evidence checkouts.
- pre-existing issue: detection is still anchored on the literal `actions/checkout@` prefix, so a fork or wrapper action (or a composite action performing a checkout) would not be inspected. No composite actions exist today — `.github/` contains only `workflows/` — so this is latent, not active.
- infrastructure issue: the guardrail now has an undeclared `ruby` tool dependency. It fails loudly rather than silently if absent, and current lanes provide it, but the dependency is not recorded anywhere alongside the guardrail registration.

yaseralnajjar:
Item 37 validation evidence

Base: b8639d9131aad26a866936dd0d71e78dc81e14a2
Candidate: f011d8e01d72cfb7cee9c6b72aa57da93348194f
Changed path: scripts/check_submodule_ownership.py

The commands ran sequentially against the exact bytes subsequently committed as this candidate, with no intervening source modification. Combined command exit status: 0.

- python3 scripts/check_submodule_ownership.py --self-test
  Output: submodule ownership self-test: PASS
- python3 scripts/check_submodule_ownership.py
  Output: submodule ownership guardrail: PASS
  Output: file-size guardrails: PASS (3759 files, limit 900 lines)
- git diff --check
  Output: none; passed.

The touched file has 559 lines. No compiler, lockfile, fixture, workflow, or gitlink inputs changed. Per the authorized continuation policy, neither create-PR nor merge-profile gate ran. No later item was implemented or tested.

PR #3769, opened 2026-09-07
test(python): compile and execute generated CFFI probe

yaseralnajjar:
The crypto ABI probe previously verified only generated C text. It now compiles the exact `cffi.gen_src` output using the selected Python interpreter's extension ABI, loads the resulting module in a child interpreter, and checks native calls before and after rejecting an overflowing argument. An intentionally invalid generated source verifies compiler-error propagation and temporary-directory cleanup on failure as well as success.

Closes Item 54 of `plans/issues/active/ad-hoc-latest-stable-release-convergence.md`. CFFI/Cryptography selections are unchanged; the only implementation path is the existing feature runner.

Validation passed on the unchanged committed bytes:

- `uv run --project verification --locked python -m sifr_verify areas run --area python_interop --suite crypto-abi-features --suite callbacks`: two variants, zero failures.
- `git diff --check`.
- `python3 scripts/check_file_size_guardrails.py`: 3,761 files; touched runner 148 lines.

No Sifr create-PR or merge gate applies to this runner-only item under its dispatch instructions. One exact-SHA Opus review is pending; final evidence will be published outside the reviewed tree.

### DIFF
--- a/verification/areas/python_interop/runner/crypto_abi_features.py
+++ b/verification/areas/python_interop/runner/crypto_abi_features.py
@@ -1,40 +1,148 @@
 from __future__ import annotations
 
+import shlex
 import subprocess
 import sys
+import sysconfig
 import tempfile
 from pathlib import Path
 
 
+MODULE_NAME = "_sifr_cffi_probe"
+COMPILE_FAILURE = "sifr_intentional_compile_failure"
+INVOKE_PROBE = """
+import importlib.util
+import sys
+from pathlib import Path
+
+extension = Path(sys.argv[1]).resolve()
+spec = importlib.util.spec_from_file_location("_sifr_cffi_probe", extension)
+if spec is None or spec.loader is None:
+    raise RuntimeError("could not load the generated CFFI extension")
+module = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(module)
+if Path(module.__file__).resolve() != extension:
+    raise RuntimeError("loaded a different CFFI extension")
+if module.lib.sifr_add(19, 23) != 42:
+    raise RuntimeError("generated native addition returned the wrong value")
+try:
+    module.lib.sifr_add(1 << 80, 0)
+except OverflowError:
+    pass
+else:
+    raise RuntimeError("CFFI accepted an out-of-range long long argument")
+if module.lib.sifr_add(-19, 23) != 4:
+    raise RuntimeError("generated native addition failed after argument rejection")
+print("cffi-native-invocation=42; overflow=rejected; recovery=4")
+"""
+
+
+class ProbeFailure(RuntimeError):
+    def __init__(self, stage: str, detail: str) -> None:
+        super().__init__(f"{stage} failed: {detail}")
+        self.stage = stage
+        self.detail = detail
+
+
+def run_checked(stage: str, command: list[str], root: Path) -> str:
+    completed = subprocess.run(
+        command, cwd=root, text=True, capture_output=True, check=False
+    )
+    if completed.returncode != 0:
+        raise ProbeFailure(stage, (completed.stdout + completed.stderr).strip())
+    return completed.stdout
+
+
+def config_flags(name: str) -> list[str]:
+    value = sysconfig.get_config_var(name)
+    if not isinstance(value, str):
+        raise RuntimeError(f"Python build configuration is missing {name}")
+    return shlex.split(value)
+
+
+def compile_and_invoke(root: Path, *, invalid_source: bool) -> None:
+    cdef = root / "probe.cdef"
+    csrc = root / "probe.c"
+    output = root / "probe_module.c"
+    suffix = sysconfig.get_config_var("EXT_SUFFIX")
+    if not isinstance(suffix, str) or not suffix:
+        raise RuntimeError("Python build configuration is missing EXT_SUFFIX")
+    extension = root / f"{MODULE_NAME}{suffix}"
+    cdef.write_text("long long sifr_add(long long, long long);\n", encoding="utf-8")
+    csrc.write_text(
+        (f"#error {COMPILE_FAILURE}\n" if invalid_source else "")
+        + "long long sifr_add(long long left, long long right) "
+        "{ return left + right; }\n",
+        encoding="utf-8",
+    )
+    command = [
+        sys.executable,
+        "-m",
+        "cffi.gen_src",
+        "read-sources",
+        MODULE_NAME,
+        str(cdef),
+        str(csrc),
+        str(output),
+    ]
+    run_checked("cffi.gen_src", command, root)
+    generated = output.read_text(encoding="utf-8")
+    if MODULE_NAME not in generated or "sifr_add" not in generated:
+        raise RuntimeError("cffi.gen_src output omitted the module or function")
+    # Compile the CLI's exact output, using this interpreter's extension ABI.
+    object_file = root / "probe_module.o"
+    includes = {
+        sysconfig.get_path("include"),
+        sysconfig.get_path("platinclude"),
+    }
+    run_checked(
+        "compile",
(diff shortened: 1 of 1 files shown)


yaseralnajjar:
Item 54 validation evidence

Candidate: `d2292f65182c38191ae717e5b1e16475edd518b9`
Base: `d038782a81c91b5560a5447ff05885ff8a915521`
PR: https://github.com/sifr-lang/sifr/pull/3769
Owned worktree: `/private/tmp/sifr-item54.VhP27O/codebase`
Only changed path: `verification/areas/python_interop/runner/crypto_abi_features.py` (148 lines).

The following named command passed on the unchanged bytes committed at this candidate, using the locked Python 3.14.7 environment:

```text
uv run --project verification --locked python -m sifr_verify areas run --area python_interop --suite crypto-abi-features --suite callbacks
python crypto ABI features ok: cffi-gen-src=compiled-loaded-invoked; overflow=rejected; compile-error=rejected; cleanup=success-and-error
[sifr-case-timing] bucket=python_interop case=crypto-abi-features elapsed_ms=2587 status=pass
python interop matrix-passed ok: report=target/verification/areas/python_interop/callbacks.latest.json
[sifr-case-timing] bucket=python_interop case=callbacks elapsed_ms=298 status=pass
result_json=target/verification/areas/python-interop-results.json
python interop verification ok: variants=2, failures=0, blocking_failures=0, non_blocking_failures=0
```

The generated C extension was compiled and linked using the interpreter's sysconfig, loaded by its exact artifact path in a child interpreter, returned 42, rejected an integer exceeding C long long with OverflowError, and returned 4 on a subsequent valid call. A second generated source contains an intentional #error; the probe requires failure at the compile stage with that diagnostic marker. Both temporary roots were confirmed absent after context exit. The callback suite is existing static matrix coverage (five certified package entries), not a claim of newly compiled Sifr callback evidence.

Common checks passed on the same unchanged implementation bytes:

```text
git diff --check
python3 scripts/check_file_size_guardrails.py
file-size guardrails: PASS (3761 files, limit 900 lines)
```

No compiler, lockfile, fixture, workflow, gitlink, CFFI release, or Cryptography release changed. Under the explicit Item 54 dispatch policy, neither create-PR nor merge-profile Sifr gate applies. No Cargo build or broad/host-sensitive test ran. E1/E2 work and evidence remain externally owned.

yaseralnajjar:
Review of `verification/areas/python_interop/runner/crypto_abi_features.py` at candidate `d2292f65182c38191ae717e5b1e16475edd518b9` against base `d038782a81c91b5560a5447ff05885ff8a915521`. Only that one file changed (`git diff --stat`: 1 file, +136/-28); no lock, project, manifest, fixture, or workflow input moved, so the runner-only dispatch policy applies and I ran no gates.

Acceptance mapping (registration row, `plans/issues/active/ad-hoc-latest-stable-release-convergence.md:462`):

- **Compile the actual generated source** — `crypto_abi_features.py:98-113` compiles `probe_module.c`, the exact `cffi.gen_src read-sources` output (line 88), with this interpreter's own `CC`/`CFLAGS`/`CCSHARED`/`LDSHARED`/`LDFLAGS` and its `include`/`platinclude` paths, then links to `EXT_SUFFIX`. Not a re-generation check: the pre-existing string check at lines 89-91 is retained *in addition to* real compilation.
- **Load and invoke** — `crypto_abi_features.py:115-119` runs the extension in a separate interpreter; the probe loads it by explicit file location (lines 18-23), rejects a `None` spec/loader, asserts the loaded `module.__file__` resolves to the produced artifact (lines 24-25, so an unrelated same-named module cannot satisfy it), and calls `lib.sifr_add(19, 23) == 42` (line 26). The parent asserts the exact evidence line (line 118), so a silent partial run cannot pass.
- **Error path** — two distinct negative paths. (a) ABI argument rejection: `sifr_add(1 << 80, 0)` must raise `OverflowError`, with an `else: raise` so acceptance fails loudly (lines 28-33), followed by a recovery call proving the module is still usable after the rejection (lines 34-35). (b) Compiler rejection: `#error sifr_intentional_compile_failure` is prepended to the C source (line 73) and the failure is **positively identified**, not merely observed — line 138 requires both `error.stage == "compile"` and the `COMPILE_FAILURE` marker in the captured stdout+stderr (`run_checked`, line 52), and re-raises otherwise. A `cffi.gen_src`, link, or invoke failure, or any unrelated compiler error, therefore propagates rather than being credited as the expected rejection; `main`'s `else` branch (line 141) fails if the invalid source compiles. The generated-content assertion also still runs on the invalid variant, so the `#error` variant cannot pass by gen_src emitting nothing.
- **Cleanup** — `run_probe:122-130` verifies deletion on *both* outcomes: the `with temporary:` block is nested inside a `try`, and the `finally` asserts `not root.exists()` whether `compile_and_invoke` returned or raised. `main` exercises both (lines 134, 136), matching the recorded `cleanup=success-and-error` marker. The comment at line 114 explains why invocation is a subprocess — the loaded extension is released before directory removal, which is what makes the cleanup assertion meaningful on the success path rather than platform-lucky.
- **Releases unchanged / runner-only** — no dependency, lock, or fixture edit; the only marker consumer is the runner itself (`grep cffi-gen-src` finds no stale expectation elsewhere), and `manifest.json:79-90` only asserts `expect_exit_code: 0`, which the new marker string does not affect.

No behavior present at the base was dropped, and no later item's work was pulled in.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `verification/areas/python_interop/README.md:139` still describes `crypto-abi-features` as "CFFI 2.1 source generation through `cffi.gen_src`", which now understates the suite (compile, load, invoke, ABI-rejection, compiler-rejection, cleanup). Docs are outside this item's declared changed path, so this is not an omission against acceptance.
- suggestion: `crypto_abi_features.py:94-97` consumes `sysconfig.get_path("include"/"platinclude")` without the explicit validation applied to `EXT_SUFFIX` (line 68) and to `config_flags` (line 59); a `None` would degrade to a literal `-INone` include flag instead of a clear configuration error. Cosmetic asymmetry only — on CPython both paths are always strings.
- infrastructure issue: `crypto_abi_features.py:100-111` depends on a usable `CC`/`LDSHARED` toolchain in `sysconfig`, so on a host or Python build without a C compiler the suite hard-fails rather than reporting a missing prerequisite. That is the correct choice under the no-fallback rule, but the failure will surface as a raw compiler diagnostic; if a non-toolchain lane is ever enrolled, that lane needs its own owner rather than a fallback here.
- suggestion: `crypto_abi_features.py:128-130` — if temp-directory removal itself failed, the `finally` guard's `RuntimeError` would replace the in-flight `ProbeFailure`, losing the original cause. Chaining (`raise ... from error`) or checking before re-raise would preserve both. Narrow: reachable only when cleanup is already broken.

yaseralnajjar:
Item 54 exact-SHA review provenance

- Base: `d038782a81c91b5560a5447ff05885ff8a915521`
- Candidate: `d2292f65182c38191ae717e5b1e16475edd518b9`
- Reviewer: Claude Opus (`claude-opus-5`), read-only plan permissions, medium effort.
- Verdict: `SATISFIED`; no blocking findings; no remediation review.
- Raw response: `/tmp/sifr-item54-opus.uAGPwv/response.md`
- Raw response SHA-256: `04bc6b88853cbc03b5d4a120f0a004cba75c0b01f130cb0bb7922cf68502a441`
- Review prompt: `/private/tmp/sifr-item54.VhP27O/review-prompt.md`

The reviewed Git tree does not contain its own final review. Focused validation and reviewer approval cover the same unchanged candidate. No Sifr gate applies to this runner-only dispatch. The four nonblocking observations are recorded as separate follow-up work; none alters Item 54's acceptance or blocks merge.

PR #3783, opened 2026-09-08
[12K-B20] Correct formatter discovery and preserve startup attribution

yaseralnajjar:
## Scope

Only item12K-B20; stacked on exact B19 full record
`fb66ed00d7672b9f5c623e3291d1b8498d50ec45`. Candidate
`ac4c277b30c6cac046ab1746fe4020c04df7eaf1`.

Correct formatter discovery: root `.gitignore` must not exclude files merely
because the absolute checkout path contains `/tmp/`. Preserve explicit-file,
force-exclude and no-respect controls; add focused regression coverage and
avoid eager irrelevant matcher construction/repeated rule scans.

NOT full B20 closure or merge approval. Startup variability was localized before
Sifr initialization, but no complete causal runtime correction was established.
Diagnostic chained linking lowered mean work but retained variation and was not
adopted. All temporary probes are removed; no workload/budget/limit weakening.
Owner https://github.com/sifr-lang/sifr/issues/3776 remains open. No original
PR3717/corpus48 main delivery.

## Exact-candidate evidence

- PASS benchmark/budget selftests;4 discovery unit tests;2 CLI regressions;
  cargo fmt; file-size guardrail; working-tree and base diff checks.
- One representative subset attempt: INVALID_ISOLATION_ABORTED. My launcher
  omitted an owned TMPDIR. First project-build warmup timed out at120004.340ms;
  time wrapper ended but owned compiler child85215 survived, overlapping the
  next sample3862 in ownedPGID84008. Descendant Cargo targeted shared system-temp
  Rust probe storage. No external cache defect is inferred from my omission.
- Exact owned group stopped/snapshotted/terminated; no shared cleanup. Formatter,
  remaining cases and dependent budget were UNREACHED. No canonical aggregate
  report or controlled-host PASS is claimed. No replay; zero gates.

Preserved evidence root `/private/tmp/sifr-b20.bj6Au6/evidence`: final-tests,
final-profile/abort-record.json and raw timeout/process receipt, all registrations,
raw observations and profiler traces. Exact-SHA read-only Opus review pending.
Parent will scope any later owner from the evidence; no next-item implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7539,6 +7539,7 @@ name = "sifr"
 version = "0.0.0"
 dependencies = [
  "clap",
+ "ignore",
  "ruff_python_ast",
  "ruff_text_size",
  "serde",
--- a/crates/sifr/Cargo.toml
+++ b/crates/sifr/Cargo.toml
@@ -25,6 +25,7 @@ sifr_sysroot = { workspace = true }
 sifr_syntax = { workspace = true }
 sifr_sql_contract = { workspace = true }
 clap = { version = "4.6.6", features = ["derive"] }
+ignore = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 sha2 = { workspace = true }
--- a/crates/sifr/src/check_and_package_commands.rs
+++ b/crates/sifr/src/check_and_package_commands.rs
@@ -7,6 +7,7 @@ use super::diagnostic_rendering_and_run::{
     current_session_package_id, execute_cargo_plan, render_diagnostics,
 };
 use super::formatter_cli::FmtArgs;
+use super::formatter_discovery::FormatterGitignore;
 use super::package_graph_context::load_package_graph_context_for_entrypoint;
 use super::package_session_cli::package_session_for_cwd;
 use super::python_runtime_context::{package_python_runtime, package_python_runtime_for_check};
@@ -578,9 +579,11 @@ pub(super) fn fmt_entrypoint(
         args.paths.clone()
     };
     let mut diagnostics = Vec::new();
+    let gitignore = FormatterGitignore::load(&cwd, config.respect_gitignore, &mut provider)?;
     for target in targets {
         let explicit_target = provider.is_file(&target);
-        let files = select_formatter_files(&target, &config, explicit_target, &mut provider)?;
+        let files =
+            select_formatter_files(&target, &config, &gitignore, explicit_target, &mut provider)?;
         for file in files {
             if args.check {
                 diagnostics.extend(sifr_format::check_path_with_options(
@@ -694,46 +697,21 @@ fn flag_override(enable: bool, disable: bool) -> Option<bool> {
 fn select_formatter_files(
     target: &Path,
     config: &EffectiveFormatConfig,
+    gitignore: &FormatterGitignore,
     explicit_target: bool,
     provider: &mut impl SourceProvider,
 ) -> Result<Vec<PathBuf>, Vec<RenderedDiagnostic>> {
     let files = sifr_format::collect_sifr_files(target, provider)?;
     let mut selected = Vec::new();
-    let ignore_patterns = if config.respect_gitignore {
-        read_gitignore_patterns(provider)?
-    } else {
-        Vec::new()
-    };
     for file in files {
         let excluded =
-            pattern_matches(&file, &config.exclude) || pattern_matches(&file, &ignore_patterns);
+            pattern_matches(&file, &config.exclude) || gitignore.is_ignored(&file, provider)?;
         if excluded && (!explicit_target || config.force_exclude) {
             continue;
         }
         selected.push(file);
     }
     Ok(selected)
-}
-
-fn read_gitignore_patterns(
-    provider: &mut impl SourceProvider,
-) -> Result<Vec<String>, Vec<RenderedDiagnostic>> {
-    let path = Path::new(".gitignore");
-    if !provider.is_file(path) {
-        return Ok(Vec::new());
-    }
-    let source = provider.read_file(path).map_err(|err| {
-        vec![formatter_cli_diagnostic(format!(
-            "could not read .gitignore for formatter discovery: {err}"
-        ))]
-    })?;
-    Ok(source
-        .as_str()
-        .lines()
-        .map(str::trim)
-        .filter(|line| !line.is_empty() && !line.starts_with('#'))
-        .map(str::to_string)
-        .collect())
 }
 
 fn pattern_matches(path: &Path, patterns: &[String]) -> bool {
@@ -854,6 +832,6 @@ fn formatting_drift_for_path(source: &str, path: &Path) -> RenderedDiagnostic {
     }
 }
 
-fn formatter_cli_diagnostic(message: impl Into<String>) -> RenderedDiagnostic {
+pub(super) fn formatter_cli_diagnostic(message: impl Into<String>) -> RenderedDiagnostic {
     diagnostic_with_code(message, DiagnosticCode::FMT_FORMATTING_DRIFT)
 }
--- a/crates/sifr/src/formatter_discovery.rs
+++ b/crates/sifr/src/formatter_discovery.rs
@@ -1,14 +1,14 @@
 //! Interpret the working directory's gitignore at its own path boundary.
 
 use super::check_and_package_commands::formatter_cli_diagnostic;
-use ignore::gitignore::{Gitignore, GitignoreBuilder};
+use ignore::gitignore::GitignoreBuilder;
 use sifr_diagnostics::RenderedDiagnostic;
 use sifr_frontend::SourceProvider;
 use std::path::{Component, Path, PathBuf};
 
 pub(super) struct FormatterGitignore {
     root: PathBuf,
-    matcher: Gitignore,
+    rules: Vec<String>,
 }
 
 impl FormatterGitignore {
@@ -20,6 +20,7 @@ impl FormatterGitignore {
         let root = normalize_path(cwd);
         let path = root.join(".gitignore");
         let mut builder = GitignoreBuilder::new(&root);
+        let mut rules = Vec::new();
         if enabled && provider.is_file(&path) {
             let source = provider.read_file(&path).map_err(|error| {
                 vec![formatter_cli_diagnostic(format!(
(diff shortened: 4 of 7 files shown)


yaseralnajjar:
Reviewed exact SHA `ac4c277b3` against base `fb66ed00d`, read-only: full diff, the first B20 progress section of `plans/issues/active/ad-hoc-emitted-rust-excellence.md`, `evidence/final-tests/results.json` + `binary.json`, `final-qualification-registration.md`, `final-profile-resolution.md`, `final-profile/abort-record.json`, `experiment-09.md`, `run_final_profile.py`, and the unchanged `run_benchmarks.py` timeout path. No tests, builds, profilers, gates, or network actions were run.

Verdict: NOT SATISFIED

Blocking findings:

- **in-scope omission — required qualification never produced a valid result (measurement accounting)**: `evidence/final-profile/abort-record.json` records `status: INVALID_ISOLATION_ABORTED`, `formatter_case: UNREACHED`, `budget_subset: UNREACHED`, 2 of 10 representative cases launched. Criterion: acceptance covers "process/runtime and measurement accounting" along `formatter-corpus-001-project-check`, and step 8 of `final-qualification-registration.md` is the registered representative/dependent subset. Reason: the invalidation is self-attributed, not external — `evidence/run_final_profile.py:6` sets only `ROOT` and passes no owned `TMPDIR`/artifact-cache environment, so the generated Rust bridge probe resolved to the shared `/var/folders/.../T/sifr_generated_artifact_cache/rust_bridge_probe_target` while a private compiler target was in use (`setup_omission` field states this explicitly). The abort was honest and no aggregate was fabricated, but the acceptance-critical measurement remains unproduced. Correction: set an owned `TMPDIR`/generated-artifact-cache in the launcher and execute the already-registered subset once, unchanged in every other respect; only then can the formatter case's retired-instruction accounting be evaluated.

- **in-scope omission — no defensible causal explanation for the retired-instruction variation**: phase doc B20 section (`plans/issues/active/ad-hoc-emitted-rust-excellence.md`, first section) and `final-qualification-registration.md` ("Current unresolved requirement") both state it. Criterion: "identify where variation arises … establish a defensible causal explanation … If an external condition demonstrably needs an unavailable intervention, name the supported condition/intervention; uncertainty alone is not an external blocker." Reason: experiment09 localizes 26.04M–30.36M instructions before the first Sifr initializer against ~17.8M for main→CLI completion — a location, not a mechanism. Experiments 10–14 (suspended spawn, dyld rebase with 107 COW/108 file-backed faults, chained linking lowering the mean ~6M while retaining variability, sparse CPU profiles with effective high-frequency 0, 100µs timer profile with SHA-256/VM/security kernel stacks) are explicitly recorded as not establishing a corrective mechanism, and no specific unavailable host intervention or OS defect is named. The independently proven discovery defect (root `/tmp/` substring-matching the host `/private/tmp/...` absolute path, checking zero files; controls 0/0/0/1 → both drift cases exit 1) is a real, separate in-scope cause that is fully corrected — but it is correctly not claimed to cover startup variation, so the primary acceptance question stays open. Correction: either establish the mechanism for the pre-initializer variation with a bounded adopted correction, or name the supported condition and the specific unavailable intervention; absent both, B20 acceptance cannot close.

- **regression — `crates/sifr/src/formatter_discovery.rs:42` (and its position at `check_and_package_commands.rs:582`)**: gitignore content that Git accepts can now abort `sifr fmt` entirely, including for explicit-file targets that must bypass exclusions. Criterion: "without weakening formatter semantics" / preserve existing explicit-file, `--force-exclude`, `--no-respect-gitignore` controls. Reason: `load` runs before target selection and hard-fails on any `GitignoreBuilder::add_line` error; that error is reachable for lines Git treats literally — e.g. `{a,b` yields `globset` `UnclosedAlternates` (`globset/src/glob.rs:599`, test `err_alt1`), as do `[z-a]` and a dangling `\`. Unclosed classes are tolerated (`allow_unclosed_class: true` by default) but alternates/range/escape errors are not. Base behavior: `read_gitignore_patterns` never parsed patterns (only a read error could fail), and for `explicit_target && !force_exclude` the ignore result was computed then discarded, so gitignore content could never block an explicit file. Now such a repo gets a `SIFR-FMT-0001` diagnostic and formats nothing, even for `sifr fmt path/to/one.sifr`. This is a strictness divergence rather than a leniency one, and the 4-control CLI test at `crates/sifr/tests/formatter_discovery.rs:31-48` only covers well-formed rules, so it does not catch it. Correction: treat an individual unparsable rule as non-fatal (collect and skip, as the `ignore` crate's own file-loading API does) or defer any rule error until that rule actually participates in a selection decision, so the unconditional explicit-file bypass is preserved.

Follow-up findings:

- **pre-existing issue**: `verification/areas/performance/run_benchmarks.py:605-624` — on `subprocess.TimeoutExpired` the wrapper returns a failure dict but never terminates the timed wrapper's descendants. This matches the observed overlap (first `sifr` build child PID 85215 surviving as PPID 1 with `first_warmup_duration_ms: 120004.34` while sample PID 3862 launched). The blob is byte-identical at base and candidate (`7dc3270d0226f68ead596c347d15c38cc21fa33b`), so this is not a B20 change; it is a cleanup deficiency for the owning scope to file, and it does not by itself establish the original timeout cause or the startup CV cause. Termination in this session was correctly limited to owned PGID 84008 (`stopped-process-group.txt`), with no shared-path removal or ownership transfer.
- **infrastructure issue**: the shared `sifr_generated_artifact_cache/rust_bridge_probe_target` observation should not be recorded as an external cache/compiler defect — per `abort-record.json` it follows from the launcher's own `TMPDIR` omission. Do not infer ownership of that shared path.
- **suggestion**: `formatter_discovery.rs:87-110` rebuilds a `GitignoreBuilder` per candidate path, with `mandatory_literal`/`could_match` (`:123-137`) as a pre-filter. The patch's own equivalence test (`:236-271`) asserts the result equals a single matcher built once from all rules — i.e. one `Gitignore` built in `load` is semantically identical, strictly cheaper per file, and would delete ~60 lines including the filter. Given B20's subject is per-invocation instruction cost, building once is the better shape.
- **suggestion**: `root` is `normalize_path(cwd)` (`:25`) and is never canonicalized, while `is_ignored` canonicalizes the candidate's parent (`:78`). This is safe in production only because `std::env::current_dir` returns a physical path; if `load` is given a non-physical root (as the unit tests' `tempfile` paths under `/var/folders/...` are), `strip_prefix` at `:91` fails and everything reports not-ignored. The unit tests call `matches_resolved_path` directly and so do not cover the mismatch. Canonicalizing `root` at load, or asserting the invariant, would close the gap.
- **suggestion (patch quality, distinct from acceptance)**: the discovery correction itself is well-targeted and evidenced — rooted relative matching, directory-alias resolution with file-symlink name preservation, engine-owned pattern/negation ordering, and CLI-level regressions covering the absolute/relative drift pair and the explicit-file × `--force-exclude` × `--no-respect-gitignore` matrix. All eight registered short checks exit 0 (`evidence/final-tests/results.json`), `sifr_format` has no source change, and the file-size guardrail is satisfied (`formatter_discovery.rs` is 273 lines). Patch quality is good; that is not B20 acceptance.

yaseralnajjar:
12K-B20 terminal: NEEDS-NEW-SCOPE / NOT COMPLETE / NOT MERGED. Stopping under the user's explicit bounded-review/terminal direction; no further implementation, replay, gate or merge.

Draft PR: https://github.com/sifr-lang/sifr/pull/3783
Base: fb66ed00d7672b9f5c623e3291d1b8498d50ec45
Frozen candidate: ac4c277b30c6cac046ab1746fe4020c04df7eaf1
Separate pushed phase-only terminal record: 96d051e4c16771ae1902f6e8aa77ecd836d45fa6 on codex/item12k-b20-terminal-record
Merge SHA: none.

Single initial exact-SHA Opus review: NOT SATISFIED, https://github.com/sifr-lang/sifr/pull/3783#issuecomment-5577108250
Raw review SHA256: 17d7c16179b389a74730ae04a16d00c685dc3aee9e5e83287763b123cffcaf92
One provider request, zero retries, zero remediation reviews, zero gates; reviewer repeated no tests. Raw review says "2 of 10 representative cases launched": correction is TWO TARGET LAUNCHES OF THE SAME FIRST BUILD-PROJECT CASE, not two distinct cases. Its proposed future rerun is not authorization to replay here. Suggestions are not adopted causal mechanisms.

Outstanding acceptance: supported causal explanation/correction for pre-initializer instruction variability (no specifically unavailable intervention established); valid unchanged formatter/representative/dependent-budget qualification; reviewer-identified regression where eager fatal gitignore rule parsing can abort explicit-file formatting despite exclusion bypass. No remediation performed under terminal-only direction. Discovery's rooted /tmp/ correction is independently evidenced but does not solve startup CV.

Keep three facts separate:
(a) MY launcher omitted owned TMPDIR. Private compiler target did not isolate the generated bridge probe target, which accessed shared default temporary storage. This invalidates qualification; it is not automatically an external dependency or proven cache/compiler defect. No shared cleanup or ownership transfer.
(b) Actual first warmup timeout120004.34008298907ms, unavailable counters/empty output; child85215 survived as PPID1 while next sample3862 launched. Preserved commands/PPID/PGID prove both belonged to ownedPGID84008 with their Cargo descendants. Only that verified group was terminated; exit143 and no members remained.
(c) Independently evidenced pre-existing timed-wrapper descendant-cleanup deficiency in unchanged run_benchmarks.py blob7dc3270d0226f68ead596c347d15c38cc21fa33b; observed survival/overlap does not prove initial timeout or original CV causation. Parent scopes the later owner, not a new inferred dependency here.

Qualification INVALID_ISOLATION_ABORTED. Formatter, later representative cases and budget UNREACHED. No aggregate/host-final report fabricated. Eight short-command receipts PASS on exact candidate, including benchmark/budget selftests, four discovery unit tests, two CLI regressions, fmt/file-size/diff guards. No silent replay.

Experiment02 chronology: symbol setup02=0 launches, embedded-Python monitoring setup02b=0, separately preregistered external-system-Python monitoring02c=6 actual exact-command launches. Old unused-budget snapshot predates02c. Failed setups retained; changed continuation used original six-target allocation, not an unchanged acceptance retry.

Evidence: /private/tmp/sifr-b20.bj6Au6/evidence/terminal.json
Terminal SHA256: 0b4fe5b32beb604527091fdfbeab9ff3b4b7dbd6d4d194e27ebb17d3260999dc
Inventory: /private/tmp/sifr-b20.bj6Au6/evidence/evidence-inventory.json
Inventory SHA256: 31e42892a35d7d3bc5ef06b5fe826b66aea2b99c8d66b2ab83ab0b628cf97bf9 (1193 artifacts).
All registrations, raw failures/limited traces, exact controls/sample budgets and abort/process receipts retained.

Owner3776 stays OPEN. No original3717/B19 integration or next-item implementation. All16gitlinks/corpus8bcbe7ab/excluded8ad089 boundary preserved. SQL builtin3f422/PR3694 independent coupling not assessed; Item65/67/release governance not absorbed. Remaining B20 acquisition/build/profiler/review processes: 0; host capacity released. Parent scopes next continuation from these receipts. STOP.

PR #3802, opened 2026-09-08
B23: reap timed benchmark process groups before the next sample

yaseralnajjar:
Timed benchmark wrapper timeouts previously killed only the wrapper, allowing its descendants to overlap the next sample. Each timed command now owns a new process group; cleanup terminates and reaps it before returning, escalates resistant members to SIGKILL, and fails closed if the group or output pipes remain. Success/nonzero status and captured text remain available.

12K-B23 only. The registered helper keeps process ownership and its deterministic tests together under the file-size limit. This does not establish the cause of the original120s timeout or formatter CV; B24 owns that qualification.

Validation on exact candidate `1af54249499ebdf42dcbb42c1fe3272810d9ee3a`, base `16a97d07327c3c3b0f153acb3662bd2bd89925a3`: both named benchmark/budget Python self-tests, exact diff hygiene, HIR maintainability and file-size guardrails PASS. Real child/grandchild tests cover cooperative/resistant timeout, exited leaders, closed pipes, retained UTF-8 output and a following sample checking old PIDs are gone; unreaped-group failure is also locked.

Only two Python harness files and the scoped phase record change. Per explicit B23 dispatch, no compiler tests/builds, benchmark acquisitions, create-pr or merge-profile gates. One initial exact-SHA Opus review SATISFIED, no blockers, zero remediation/retries. Review SHA256 `62889dd3be376f38ea9b7e268cb9f06ec4811e09f27cb3fbd4b013bc55fe84ab`; full response published in the PR. Five nonblocking follow-ups are deferred to the record-only update, with no next-item implementation.

External receipts: `/private/tmp/sifr-b23.inohbu/evidence/validation.1af54249499ebdf42dcbb42c1fe3272810d9ee3a.json`, SHA256 `0dbbbccfae532dcbd3942cf36bb9ff9d72262a88793be696ec10d9bedabad4fa`.

### DIFF
--- a/plans/issues/active/ad-hoc-emitted-rust-excellence.md
+++ b/plans/issues/active/ad-hoc-emitted-rust-excellence.md
@@ -2,6 +2,62 @@
 
 Status: active
 
+## Item12K-B23: timed subprocess group ownership (2026-09-08)
+
+Bounded implementation awaiting its named validation and exact-SHA review.
+Base: `16a97d07327c3c3b0f153acb3662bd2bd89925a3`; independent owner root
+`/private/tmp/sifr-b23.inohbu`, branch `codex/item12k-b23-timeout-lifecycle`.
+The parent and every predecessor clone, index, target and reference are read-only.
+Owned launch environment uses sibling `tmp` and `evidence`, with
+`CARGO_TARGET_DIR` unset. No historical parent ledger is copied into this record.
+
+Scope follows the dispatched B23 registration and the merged
+[B26 assessment](ad-hoc-joint-emitted-rust-delivery-assessment.md#existing-b23-timeout-descendant-lifecycle).
+Only `run_benchmarks.py`'s timed process boundary and its registered
+`benchmark_process.py` helper change. The helper owns each command's new session
+and process group, terminates descendants on timeout, drains output, reaps the
+leader and waits for the group to disappear before returning. Resistant members
+receive SIGKILL after a bounded TERM grace. Failure to drain or finish reaping
+raises a benchmark error, preventing another sample. Ordinary success/error
+status and output survive; timeout output is text and unavailable metrics remain
+unavailable. Background descendants left by an exited leader are cleaned too.
+POSIX descendants inherit this group; deliberate session/group escape is outside
+this benchmark-command contract. Orphan reaping belongs to the OS adopter, and
+failure to reap is a blocking error rather than an overlapping next sample.
+
+All deterministic tests are implemented before execution: ordinary exit0/7;
+ready-handshaken cooperative and TERM-resistant child/grandchild trees;
+leader-exit with inherited pipes; closed-pipe background descendants; UTF-8
+stdout/stderr retention; PID/group disappearance and a following sample proving
+no overlap; simulated failure to reap must fail closed. These checks are called
+by the existing benchmark `--self-test`, with no compiler or benchmark workload.
+
+Named validation only:
+
+- `python3 verification/areas/performance/run_benchmarks.py --self-test`
+- `python3 verification/areas/performance/check_budgets.py --self-test`
+- `git diff --check <exact-base> <exact-candidate>`
+- `python3 scripts/check_hir_maintainability_guardrails.py`
+- `python3 scripts/check_file_size_guardrails.py`
+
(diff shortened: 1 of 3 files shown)


yaseralnajjar:
I read the exact diff, both changed Python files in full, every direct caller of `run_subprocess` (`run_case`, `run_frontend_query_case`, `run_lsp_query_case`, `ensure_sifr_binary`, `ensure_frontend_query_bench`, and the `controlled_sampling.run_controlled_case` → `run_case_fn` path), the metrics parser (`process_metrics.timed_command` / `parse_process_metrics`), the candidate's B23 record section, and the merged "Existing B23" registration. I verified the five named logs' SHA-256 values against `validation.1af5….json` (all match, all exit 0) and ran no tests, builds or gates.

Coverage substantiation:
- **Ownership**: `benchmark_process.py:72` launches the timing wrapper with `start_new_session=True`, so pgid == leader pid and `signal_group`/`group_exists` (`:22`, `:29`) can never touch the caller's group. All timed benchmark commands (measured, warmup, query, and helper-build) funnel through the single `run_subprocess` boundary; `controlled_sampling` and `work_baseline` hold no timed subprocess of their own, so no timed path bypasses the new owner.
- **Timeout termination/reaping before the next sample**: `finish_group` (`:36`) TERMs the group, drains, waits out a bounded grace, SIGKILLs, then re-drains and blocks until `killpg(pgid,0)` fails. Failure to drain or to see the group disappear raises `BenchmarkError` ("refusing to start another sample"), so an incomplete reap blocks sampling instead of being hidden. The `else` branch also closes a group left behind by an exited leader, and the `BaseException` branch closes it on interrupt.
- **Evidence preservation**: `Popen.communicate` resumes its existing per-fd buffers on the second call, so partial timeout output is retained, and the values returned are decoded `str` (the base path returned `TimeoutExpired.stdout`, which is `bytes` under `text=True`); `parse_process_metrics(completed.stderr)` and the success/nonzero-exit result shape are otherwise unchanged, and timeout metrics stay `None` / `"unavailable"`.
- **Tests through the existing entrypoint**: `run_benchmarks.run_self_test` calls `run_benchmark_process_self_test(run_subprocess)` first; the log shows all seven checks plus the inherited suites — ordinary exit 0 and 7, fail-closed unreaped group (mocked, no leaked process), and cooperative/resistant/leader-exit/closed-pipes 3-deep trees each asserting group disappearance, UTF-8 stdout/stderr retention, and a following sample proving the prior PIDs are gone. Readiness handshaking, not a guessed sleep, gates the 100 ms timeout.
- **Guardrails**: 816 and 233 lines, both under 900; file-size PASS (3762 files) and HIR PASS logs authenticated. Record update is confined to the B23 section of the active issue.

Limits: evidence is read, not reproduced; the tree tests exercise the Darwin `/usr/bin/time -l` wrapper on this host only; nothing here speaks to the original 120 s timeout or formatter CV (B24's scope), as the record itself states.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `benchmark_process.py:22`/`:29` — `group_exists`/`signal_group` key on the leader's pid after `communicate` has reaped it, so an extremely unlikely PID recycle could make the post-success `group_exists` check consult (or the subsequent SIGKILL hit) a foreign group; inherent to POSIX pgid cleanup and not introduced defectively here, but a `waitid`-free hardening note if this helper grows.
- suggestion: `benchmark_process.py:24` — only `ProcessLookupError` is handled; an `EPERM` from `killpg` (unsignalable group member) would escape as a raw `PermissionError` rather than the module's fail-closed `BenchmarkError`.
- suggestion: `benchmark_process.py:196` — `assert owned[0].pid != os.getpgrp()` only proves the owned group differs from the caller's; `os.getpgid(owned[0].pid) == owned[0].pid` would assert session-leadership directly.
- pre-existing issue: `run_benchmarks.py:432` — a timed-out *warmup* sample is `continue`d before the `timed_out` check, so only measured-sample timeouts raise; unchanged by this diff, and such warmups are now at least cleaned up before the next sample.
- infrastructure issue: `benchmark_process.py:13` imports `unittest.mock` at module scope inside a production runner import path; consistent with this area's inline `--self-test` convention, but it couples the runtime helper to test-only machinery.

PR #3804, opened 2026-09-08
docs: B33 static module-instance identity contract

yaseralnajjar:
B32's mapped handover stopped on two dyld representations whose recorded metadata could not establish module object identity. B33 records a static/offline external resolver contract that requires native SBModule equality across all aliases, then checks current mapped header, section, exact symbol and address round trips before binding. Historical B32 identity remains unknown.

Only Markdown changes enter this PR. The scoped document binds the external helper paths and SHA256 manifest. Validation: named encompassing offline suite 299/299 (231 inherited + 68 module cases), Python3.9 syntax parsing, exact candidate diff, HIR and file-size guardrails. The initial 297/299 receipt and corrected source are preserved. No live debugger/inferior proof or broad gate ran. Exact-SHA Opus review precedes merge.

This closes only 12K-B33's static prerequisite. B34 custody evidence, the later combined proof, B24, SQL/B27 and the full emitted-Rust phase remain open.

### DIFF
--- a/plans/issues/active/ad-hoc-emitted-rust-b33-module-instance.md
+++ b/plans/issues/active/ad-hoc-emitted-rust-b33-module-instance.md
@@ -1,0 +1,162 @@
+# 12K-B33: module instance identity at mapped handover
+
+Date: 2026-09-08. Owner: performance / issue3776. Exactly B32-F1.
+State: static/offline contract implemented; 299/299 PASS, awaiting exact-SHA review.
+
+## Scope and ownership
+
+Sole session owns `/private/tmp/sifr-b33.W1oIqs/sifr`, independent fresh-main
+clone, branch `codex/item12k-b33-module-instance`, base
+`491ba4ede1609ce476831015dc209a9064cd8ffc`; private index/object store and TMPDIR
+`/private/tmp/sifr-b33.W1oIqs/tmp`. External apparatus/evidence lives in
+`/private/tmp/sifr-b33.W1oIqs/evidence` (E). No shared targets or alternates.
+The parent dirty phase ledger's top B33/B34 execution registration and direct
+B33 onboarding authorize this scope; lower historic headers do not supersede it.
+All parent/predecessor files, indexes, refs and raw remain read-only.
+
+B32 native worker closed. Terminal SHA256
+`cbcf4bd5ed995dc3453e46294c12ddb620bf7fec1a188b9378081e6c6a25f74a`, scoped
+document SHA256 `f4f2116b378ae9506596e2fe7c275fa438ef1d3ef17a3a3ef3d4d17a93161d00`,
+remote record `9aad3dd68a77671193523ae503e4251215a13afc`, branch
+`codex/item12k-b32-output-custody`, issue3776 comment5579599977. Its stop7
+duplicate dyld rows have equal recorded metadata but unrecorded object identity.
+Historical identity stays unknown and historical coverage stays INCONCLUSIVE.
+
+Implement only the owned external module resolver/observer integration and
+offline coverage. Retain the inherited 231 address/symbol/handover/output cases,
+all existing guards, canonical exclusive output files and outer consumer.
+B34 alone owns custody rejection evidence. No B34 implementation, live debugger
+probe, inferior launch/attach/continue, build, Cargo/Clippy/Sifr gate, CV/counter
+or acceptance acquisition. This merge is a static prerequisite only. B34's
+contract and integrated negatives must also complete before a later combined
+proof can be proposed; that proof remains unallocated. B24 causal/full budget,
+SQL/B27 joint delivery/approved65, and the full phase remain open.
+
+## Registered owned files and named commands (before execution)
+
+Owned apparatus under E: `coverage_module.py` (resolver and API snapshot),
+`coverage_module_fakes.py` (explicit synthetic API objects),
+`coverage_module_tests.py` (same resolver entry points), `coverage_symbols.py`
+(inherited suite and encompassing entry), `lldb_coverage.py` (integration),
+`coverage_address.py`, `coverage_output.py`, `coverage_output_tests.py`,
+`run_coverage.py`, `coverage.lldb`, `identity-contract.json`, and
+`module-inventory.json`. The outer launch apparatus is retained for static
+integration review only; no executable or live command is registered for B33.
+`record_b33.py` authenticates/hashes/serializes evidence and runs only the named
+documentation checks; it never invokes behavioral tests or debugger execution.
+Source references: installed LLDB Python wrapper, pinned public `SBModule.cpp`,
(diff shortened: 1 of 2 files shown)


yaseralnajjar:
Exact-SHA Opus review for item 12K-B33.

Base: `491ba4ede1609ce476831015dc209a9064cd8ffc`
Candidate: `3446f57dc2d21f584d30188f91ede95540af67f7`
Frozen external manifest SHA256: `81597f166178476699cb73a6ca7b13c980624125a5603c89c6b7cafc5e185a31`
Review response SHA256: `432ab8c8586843136b21f7c9a77b74d8075faeeb96998ed996d5c9fdc9a0b0b8`
Validation: 299/299 named offline checks, exact-candidate diff/HIR/file-size PASS. One initial review, zero remediation reviews, zero live proofs or broad gates.

I authenticated the exact delta and the frozen external candidate directly (all 17 manifest digests recomputed and matching; manifest self-hash matching; installed LLDB wrapper `08d7c4c6…` and the four pinned `SB*.cpp` copies re-hashed and read).

## Verdict: SATISFIED

**What I verified against the authoritative scope**

- **Native identity basis, not metadata.** `coverage_module.py:10` `equal()` is tri-state: it rejects invalid/mismatched-type operands and returns UNKNOWN on non-`bool`, asymmetric (`forward != reverse`), or self-contradictory (`forward == different`) results. I confirmed the installed wrapper dispatches `SBModule.__eq__`/`__ne__` to `_lldb` at lines 10034/10040 exactly as documented, and that the pinned `SBModule::operator==` compares `m_opaque_sp.get()` (module pointer), `SBSection::operator==` the `SectionSP`, `SBSymbol::operator==` the `Symbol*`, and `SBAddress::GetModule` the owning module. `FindSymbols` sets `sc.module_sp = module_sp` (`SBModule.cpp:347`), backing the per-context ownership check.
- **No first-match selection.** `select` (`coverage_module.py:154`) requires an all-`True` pairwise matrix over every mapped path row, then derives the handle from the single agreed header load via `ResolveLoadAddress(load).GetModule()` and re-proves `same(module, anchor)` for each row. Test `module-synthetic-native-alias-count-*` asserts `current_module(...) is c.anchor`, not the enumerated wrapper. `symbols.pop()` (`:224`) takes the anchor's own lookup, not a match choice.
- **B32 duplicates preserved as UNKNOWN.** `coverage_module_tests.py:11-20` re-hashes B32 `events.json` (`503a9225…`, which I recomputed), asserts the stop-7 dyld rows are still two identical rows, replays them with `token=None`, and requires the receipt matrix be `[[None,None],[None,None]]` with `state == 'REJECTED'` — no retroactive acceptance, no UUID/metadata fallback anywhere in the path.
- **Rejection coverage is real.** 39 negatives plus 3 file-roundtrip negatives cover distinct/unknown/invalid identity, conflicting module UUID/triple/path, header file/load conflict, contradictory unloaded alias, null header, section owner/instance/size/file/load, context owner and count, symbol instance/name/type/extent/owner/section, load- and file-address roundtrip module/section/offset/symbol, snapshot state, mid-operation stop and epoch drift, and region proof. Case arithmetic checks out: 1+3+13+39+3+2+1+1+2+2+1 = **68**, and `offline-result.json` reports 299 cases, 0 failing, 68 `module-` prefixed.
- **Correction claim is exactly as stated.** `diff failed-source/coverage_module.py coverage_module.py` is a **single line** (`:222`, `module` → `alias`), and `offline-initial-failed.json` fails exactly `module-file-roundtrip-foreign-start` and `-end` (297/299). Nothing else changed.
- **Inherited guards preserved.** `coverage_symbols.py` diff vs B32 is purely additive (owned root, item id, `require(len(results) == 231)`, module hook, three new 3.9-parse targets, provenance/hash fields); the `182` inherited assertion is untouched. `coverage_address.py`, `coverage_output.py`, `coverage_output_tests.py` are byte-identical to B32; `run_coverage.py` and `coverage.lldb` differ only in the owned-root path. The exclusive-file route and outer consumer live in the byte-identical `coverage_output.py`; the new `module-instance-resolutions.json` only lands in `inventory.json`, which enumerates rather than expects.
- **Observer integration is complete, not partial.** `lldb_coverage.py` routes `current_module`, `identity`, `resolve` (hence `arm`, `verify_sites`, `refresh_later`) through the shared `Resolver`, and adds `initial_address` at the initial-PC stop. The instance token flows through `bind` → `binding['module']`, which `verify_sites:134` already compares, so a same-metadata replacement is detectable (`module-same-metadata-replacement-visible-in-binding`). The removed `ambiguous current module` check is strictly superseded, not weakened: aliases are now admitted only after native equality *plus* identical module/header metadata, identical section trees, per-section identity, and header address agreement.
- **Capability basis exists.** Every API member used (`GetObjectFileHeaderAddress`, `FindSection`, `FindSubSection`, `GetNumSubSections`/`GetSubSectionAtIndex`, `GetParent`, `ResolveFileAddress`, `sections`, `ResolveLoadAddress`, `SBAddress(SBSection, offset)`, `GetContextAtIndex`, region accessors) is present in the installed wrapper. No missing-capability blocker; strong `SBModule` refs in `Resolver.instances` keep `IsValid()` true across module removal, so token retention does not degrade to UNKNOWN.
- **Bounds honored.** Only the two Markdown paths changed (186 insertions); tree clean; `checks-3446f57d….json` shows `diff --check` exit 0, HIR guardrails PASS, file-size PASS 3762 files, external helpers max 588 lines; `input-authentication.json` PASS 47; `inferior_launches: 0`, `live_proofs: 0`; `record_b33.py` is hash/serialize/named-checks only with no test or debugger entry point.

## Blocking findings

None.

## Follow-up findings

- **suggestion** — `plans/issues/active/ad-hoc-emitted-rust-b33-module-instance.md:100`: the `SBModule.cpp#L185` anchor points at a comment inside `GetUUIDString` in the frozen copy; `SBModule::operator==` is at line 197. The three sibling anchors (`SBAddress.cpp#L203`, `SBSection.cpp#L229`, `SBSymbol.cpp#L82`) are exact. The substantive semantic claim is correct as written; only the line anchor drifts.
- **suggestion** — `lldb_coverage.py:13-15`: `bind` and `resolve_identity` are now dead imports after the move into `coverage_module.py` (`expected_path`, `check_site`, `stop_evidence` remain used). Cosmetic.
- **infrastructure issue** — `coverage_module.py:96-98`: `self.record(...)` runs in the `finally` of `operation`, so a persistence failure during error propagation would mask the original rejection reason. Mirrors the inherited B32 event-save pattern; not introduced by this item.
- **pre-existing issue** — `run_coverage.py:184-187` still expects `prelaunch.json` under the owned root, which does not exist in `E` for B33. Inert here (no executable command is registered for this item) and B34 owns that custody path.

I did not create a plan file or exit plan mode: this was a read-only review with a fixed required output, and there is no implementation to propose.

PR #3816, opened 2026-09-08
docs: plan selected R2 archive adapter and custody proof

yaseralnajjar:
Plans the selected R2 Standard archive boundary over the merged F1B ArchiveStore: conditional create-only PUT, direct full-byte reads, indefinite complete-root lock/no-expiry admission, and explicit reader/receipt-writer/custodian separation.

Registers the bounded offline 70-F1C1 implementation paths and nine synthetic cases, with a standard-library-only Python 3.14.7 command. Real SDK delivery, account access, protection and independent-copy proof remain explicit later 70-F1C2 prerequisites; no provider implementation or cloud action is included.

Validation on candidate `cae41b5aaf98f132d6eebe39f761d4a514293e38`: exact-base diff and working-tree diff checks passed; file-size guard passed (3,767 files); local links/proposed-path absence, five base-tree blobs, five official-source snapshot hashes and API identities passed. Only two Markdown files changed; no Sifr gates per the item rule. Exact-SHA Opus review pending.

### DIFF
--- a/plans/issues/active/ad-hoc-latest-stable-item70-f1c0-r2-plan.md
+++ b/plans/issues/active/ad-hoc-latest-stable-item70-f1c0-r2-plan.md
@@ -1,0 +1,341 @@
+# Item 70-F1C0: selected R2 contract and offline implementation plan
+
+Planning assessment, 2026-09-08. Owner: release/distribution,
+[issue #3775](https://github.com/sifr-lang/sifr/issues/3775).
+This implements the coordinator's registered documentation-only F1C0 scope.
+R2 Standard is the selected primary design, with indefinite bucket locks,
+configuration credentials segregated from producer credentials, and a
+separately controlled full-byte copy. Provider selection is settled. No actual
+account, bucket, credential, copy medium or custodian is established here.
+
+The [F1A assessment](ad-hoc-latest-stable-item70-f1a-backend-readiness.md) and
+[merged F1B contract](ad-hoc-latest-stable-item70-f1b-offline-archive.md) are
+prerequisites. The [Item 70 disposition](ad-hoc-latest-stable-item70-evidence-recovery.md)
+and complete four-native/two-structured-skip matrix remain authoritative.
+Historical custody remains INCOMPLETE/UNRECOVERED; this plan creates no
+qualification, approval, publication or retrospective evidence.
+
+## Canonical boundary and capability evidence
+
+Use the existing [ArchiveStore](../../../verification/areas/distribution_release/governance/archive_store.py)
+protocol unchanged: `location`, atomic `create(key, content) -> bool`, and
+fresh `read(key) -> bytes`. Existing `seal`, `readback`, and `attest_readback`
+continue to own inventory validation, hashing, root sealing and append-only
+receipts. Provider code transports exact bytes; it does not duplicate the
+manifest validator or declare administrative independence.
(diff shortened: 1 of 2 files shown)


yaseralnajjar:
Item 70-F1C0 validation

Base: 0c0f4f046b769deeb8cf91a83d62fbb7ae983ec1
Candidate: cae41b5aaf98f132d6eebe39f761d4a514293e38
Changed paths: plans/issues/active/ad-hoc-latest-stable-item70-f1c0-r2-plan.md; plans/issues/active/ad-hoc-latest-stable-release-convergence.md

PASS git diff --check HEAD^ HEAD
PASS git diff --check
PASS python3 scripts/check_file_size_guardrails.py: 3,767 files, 900-line limit
PASS python3 /private/tmp/sifr-item70-f1c0.XM898D/check-docs.py: four local plan links, ledger link, three absent proposed paths, five exact base blobs, five official-source hashes, API capability markers.
Clean working tree after candidate commit/checks.

No compiler, lockfile, fixture or workflow changes; zero create-PR/merge gates. No other item tests, Cargo, native execution, historical search, cloud actions or qualification. Python/dependency/source/API reads are the named planning investigation, not offline adapter tests or live access proof.

yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: F1A's registered nonblocking follow-up "recheck pricing when an actual provider is selected" (`plans/issues/active/ad-hoc-latest-stable-item70-f1a-backend-readiness.md:251`) is the one of its three F1C follow-ups that this selection item does not route anywhere; the other two are satisfied (full UTC timestamps are used at `...item70-f1c0-r2-plan.md:187`/`:326`, and role separation with observed protection metadata is assigned to F1C2 at `:288`–`:296`). Cost was not part of the coordinator's F1C0 scope, so this is not an in-scope omission — worth attaching to F1C2 when the account fact is resolved.
- suggestion: the snapshot table (`...item70-f1c0-r2-plan.md:325`–`:336`) says the HTML was fetched "from the exact Cloudflare URLs in the capability table", but only five of the nine Cloudflare URLs in that table have retained snapshots (the Boto3 reference, both `api.cloudflare.com` resource pages, limits, and error-codes pages have none). The claims sourced from the unsnapshotted pages (5 GiB single-part, 1,024-byte keys, `IfNoneMatch="*"`/412, control-plane response shapes, error classes) and the uncited R2 bucket-name 3–63 character contract at `:72` are individually consistent with current provider documentation, but a future reader cannot re-verify them from retained bytes. Naming the snapshotted subset explicitly, or snapshotting the remainder in F1C2, would close that gap.
- suggestion: spot-checked source claims all hold against the retained bytes — `If-None-Match` present in `sources/r2-s3-api.html`, indefinite/prefixless/new-and-existing coverage and lock-over-lifecycle precedence in `sources/r2-locks.html:138`,`:259`–`:261`, and the exact default/`eu`/`us`/`fedramp` endpoint derivation in `sources/r2-auth.html:169`–`:175`. The locks page additionally states that with multiple overlapping rules "the strictest (longest) retention requirement takes precedence" (`sources/r2-locks.html:256`); adding that to the admission rationale would strengthen the "inspect all rules" requirement at `...item70-f1c0-r2-plan.md:132`–`:143` without changing it.

PR #3820, opened 2026-09-08
Qualify B38 coherent acquisition and batch corroboration

yaseralnajjar:
Closes only the B38 acquisition qualification: coherent initial parent reconciliation and shared batch corroboration preserve every native/ps receipt within the existing two-round/six-second/64-subject bounds. The executable diagnostic apparatus is frozen outside the Git tree and bound by the candidate's record; compiler and runtime product files are unchanged.

Validation: named offline suite493/493 (443 retained+50), one integrated live proof PASS23.17s with one target, handover/all8entries, exit0/exact2files and independent fresh zero-process release. Diff/HIR/file-size checks pass; no Sifr gates under the explicit docs-only rule. Actual live scheduling required no resampling; race coverage is the offline sequence16/26 replay and injected negatives.

One exact-SHA read-only Opus review: SATISFIED, no blockers, zero remediation/provider retries, against f7e0f1fdb5a74862f5c3be1671263e66264db693 and the actual frozen external sources/receipts. [Complete review](https://github.com/sifr-lang/sifr/pull/3820#issuecomment-5583126928), response SHA256404c9a4aeb793936a714caba04301b73b00af2601814fff454c78f2ff17a6b41. Merged as ab16c6687bb65bed22e1d3bead3a0682ce67ef2b. Full B24/B27/SQL/original phase remains open.

### DIFF
--- a/plans/issues/active/ad-hoc-emitted-rust-b38-acquisition-reconciliation.md
+++ b/plans/issues/active/ad-hoc-emitted-rust-b38-acquisition-reconciliation.md
@@ -1,7 +1,41 @@
 # 12K-B38: coherent acquisition and batch corroboration
 
 Date: 2026-09-08. Owner: performance / issue3776.
-State: implementation and offline qualification complete; sole live proof pending.
+State: offline and sole live qualification PASS; exact-SHA review/merge pending.
+
+## Sole live outcome and verified release
+
+Prelaunch candidate `ad59a78f0e4ac79cdfbb247c3d93845e5f80284f` was committed
+and pushed before both callbacks and the sole actual command. PASS in
+23.17234345903853 seconds: one LLDB session/target84958;15 events; one
+mode3→mappedmode0 handover; all8 ordered entries; actual exit0 and exactly2 file
+checks. Canonical stdout is empty and stderr1361 bytes. Full output completion
+passed after actual exit and fresh runner process release. All3 AC admission
+snapshots passed. No repair, retry, second proof or counter/CV acquisition.
+
+Custody retained71 observations with no rejections. This live scheduling did
+not require acquisition resampling; F2/F3 race coverage is the named offline
+suite, including actual sequence16/26 initial receipts and synthetic fresh
+rounds. The live result qualifies the integrated apparatus, without claiming
+that those races recurred in this proof.
+
+Independent post-command native-before/full-ps/native-after verifies all owned
+PIDs and groups84299/84866/84958/84959 absent; group queries return ESRCH,
+no matching rows, watcher/monitor false. No signals were delivered by this fresh
+verification. Capacity-release callbacks were sent to parent and coordinator.
+
+Evidence under E=`/private/tmp/sifr-b38.tJeQq1/evidence`:
+
+- `coverage/outcome.json`: `9f4b4255badb50c1c00ec63b55e1584de3b6588ecb344cc62c5e3b5b3da8ede5`
+- `coverage/custody.json`: `585449af7c6d1c8f2934f99eb1f5f2eb921bd269da40a6f46c8d3b1b8101cdd7`
+- `coverage/output-completion.json`: `1357d564c2eb5ce985799b462b3d967dc03e18f7f47240db3d1d2bde547cf47c`
+- `coverage/process-release.json`: `9ec9345ef5926386e83c26ee5adf1b8a76445222069cc3f1ed578ed7eda69d1d`
+- `fresh-process-release.json`: `8e252fea58fd176550699c29143b55bb81c7b2f03906fec940146b24b9190978`
+
+All tested executable files remain unchanged. Success-only named checks and
+exact-SHA external Opus review follow. Only this diagnostic record changes in
+the repository, so create-PR and merge Sifr gates are not applicable under the
+user's explicit item instructions. Full B24/B27/SQL/original phase remains open.
 
 ## Ownership and authority
 
--- a/plans/issues/active/ad-hoc-emitted-rust-b38-acquisition-reconciliation.md
+++ b/plans/issues/active/ad-hoc-emitted-rust-b38-acquisition-reconciliation.md
@@ -1,0 +1,135 @@
+# 12K-B38: coherent acquisition and batch corroboration
+
+Date: 2026-09-08. Owner: performance / issue3776.
+State: implementation and offline qualification complete; sole live proof pending.
+
+## Ownership and authority
+
+Sole owner: independent fresh-main clone `/private/tmp/sifr-b38.tJeQq1/sifr`,
+branch `codex/item12k-b38-tJeQq1`, base
+`5614f06c8ff49411dd8d0b8107e6479e4274ec96`. Private sibling `tmp` and
+`evidence` (E); independent index, refs and object store, no alternates or target.
+Parent dirty phase/dependency records and all predecessor paths remain read-only.
+Exact commands: `/private/tmp/sifr-b38.tJeQq1/registration.md`.
+
+The parent/coordinator authorized both B37-F2 initial parent disagreement and
+B37-F3 corroboration exhaustion, followed by ONE changed-apparatus qualification.
+The authoritative scope is the parent's top B38 phase registration. No new API
+research, arbitrary parent acceptance, dropped subjects, blanket helper-name
+exemption, no-event exit inference or full target retry is authorized.
+
+B37 continuation terminal authenticated SHA256
+`388d4049799db2f4c03b15142019b18cc7125c534f749cfc9ed97a2790746fc7`, remote
+record `0dcf02aaa3f7230f3c8c4c6e5da79f26367c90ba` on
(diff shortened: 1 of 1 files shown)


yaseralnajjar:
## Review — Sifr item12K-B38 (initial exact-SHA read-only Opus review)

**Base SHA:** `5614f06c8ff49411dd8d0b8107e6479e4274ec96`
**Candidate SHA:** `f7e0f1fdb5a74862f5c3be1671263e66264db693`
**Prelaunch:** `ad59a78f0e4ac79cdfbb247c3d93845e5f80284f`

## Verdict: SATISFIED

## Blocking findings

None.

## What I inspected

**Git.** Exactly one changed path, base→candidate: `plans/issues/active/ad-hoc-emitted-rust-b38-acquisition-reconciliation.md` (A, +169). Prelaunch→candidate is a pure addition of the outcome section (`State:` line + "Sole live outcome and verified release"); no other repository byte changed. Worktree clean. No compiler/lock/fixture/workflow path touched, so the create-PR and Sifr merge gates are correctly excluded and I did not run them.

**Frozen apparatus integrity.** All 351 files in `E/frozen-manifest.json` verified byte-identical (`shasum -c`: 351/351 OK), including `sifr-experiment09`, the LLDB framework Python, and every attempt source. `frozen-manifest.source_hashes` == `offline-result.json.source_hashes` == the 20 live-on-disk files exactly. `E/authentication.json` self-hash `41ac93…742c68` matches `frozen-manifest.authentication_sha256`; `terminal_bound_files` 226 (distinct union), `prior_frozen_files` 144, `runtime_files` 10, `prior_terminal_sha256` `388d4049…46fc7`. All five record-cited hashes plus `offline-result.json` / `frozen-manifest.json` / `prelaunch.json` reproduce exactly.

**Inherited-module relocation.** Diffed all 14 supposedly-unchanged inherited files against `E/prior-b37-cont`: 8 byte-identical; 6 differ only by the registered paths/run-identity relocation (`sifr-b37-cont.bJIsfC`→`sifr-b38.tJeQq1`, `RUN_ID` `12K-B37-cont-bJIsfC-output-1`→`12K-B38-tJeQq1-output-1`). No assertion or behavior change.

**The scoped change.** Read `coverage_acquisition.py` (full rewrite, 258 lines), `coverage_native.py`, `coverage_custody.py`, `coverage_symbols.py`, `coverage_native_fakes.py` diffs, and `coverage_b38_tests.py` in full.

- *B37-F3 batch:* `batch()` (`coverage_acquisition.py:19`) samples **all** subjects around **one** ps snapshot; `require(len(rounds) < 2)` before append and `receipt['end'] - native['begin'] <= 6` bound F2+F3 jointly to two snapshots / six seconds; the 64-subject bound is retained at `coverage_native.py:118`, `coverage_acquisition.py:58` and `:113`. The receipt is appended *before* any fallible sampling (`:26`), so partial receipts survive. `confirmed_departures():210` requires `departed == set(acquisition['missing'])` — no subject can be dropped, and there is no basename exemption anywhere.
- *B37-F2 coherence before authority:* `initial_chain()` is consumed at `coverage_custody.py:275`, before `self.first` admission, `acknowledge()` and `cleanup_allowed()`. It requires exactly two candidate parents, both resolvable to the live target→debugserver→LLDB→root chain by native path, correct groups, and `ppid` restricted to that chain **in every intermediate round** (`:242`–`:257`). The original disagreement is retained intact under `native['acquisition']['original']` and re-derived independently in `validate_rounds():122`–`:127`. Chains are recorded as `kind: coherent-initial-acquisition, historical_cause: UNKNOWN`, separate from `transitions` — no invented transition.
- *Canonical debugserver identity:* `DEBUGSERVER = os.path.realpath(...)` is the right root-cause fix, not a name exemption. Verified against the host: `LLDB.framework/Resources → Versions/Current/Resources`, `Current → A`. Verified against raw evidence: `coverage/custody.json` `native_initial["84959"].registration.after.path` is `.../Versions/A/Resources/debugserver` while the same PID's ps row shows `.../LLDB.framework/Resources/debugserver`. With the old symlink constant `role(84959)` would have been `None`.
- *Negatives:* root disappearance (`coverage_native.py:127`, now unconditional — stricter than B37), reappearance (`:134`), PID reuse/path/group via `stable()`, exec/watch-loss (`:76`, `:151`), registration replacement (`:146`, `:162`), new unobserved descendant (`:140`), count/time exhaustion (`for…else raise`, `:82`).

**Offline evidence (reused, not rerun).** `offline-result.json`: `state PASS`, `count 493`, `inherited_cases 443`, `lifecycle_cases 50`, `inferior_launches 0`, 0 failures, 493 unique names. Diffed the 443 non-B38 names against `prior-b37-cont/offline-result.json` — **zero drift**; nothing dropped, renamed or deferred. `coverage_symbols.py` now hard-asserts `len(results) == 443` after the B37 suite, which locks this in. Attempts reproduce 480/491, 490/491, 491/491, 493/493; `attempt-4-sources` is byte-identical to the active sources, and across attempts only `coverage_acquisition.py`, `coverage_native_fakes.py` and `coverage_b38_tests.py` changed (`coverage_native/custody/symbols` never did) — same-suite in-scope correction only. The final +2 cases are exactly `B38-initial-reject-transient-{native,ps}-parent`. Attempt-1's three inherited failures were fixed, not suppressed.

**Live proof.** `coverage/outcome.json` `state PASS`, `elapsed 23.17234345903853`, `debugger_sessions 1`, `target_launches 1`, `event_count 15`, `exit 0 / exited true`, `handover_count 1`, all 8 hits in order, `file_check_phases 2`, `finalized_after_exit_and_release true`, `zero_processes true`, `failures []`. `inferior.json` `flags 134`, pid/pgid 84958. Stream hashes on disk match `output-completion.json` (stdout 0 bytes, stderr 1361). `admission.json`: 3 snapshots, `rejections []` each, AC required. `custody.json`: 71 observations, 0 rejections, `ACTIVE`, `target_ids [84958]`, `tainted []`. `fresh-process-release.json`: `ABSENT`, `signals_delivered 0`, `matching_rows []`, ESRCH for 84299/84866/84958/84959, `native_before`/`native_after` all null, watcher/monitor false. I independently confirmed the record's honest claim that this scheduling did **not** exercise resampling: all 71 observations have zero `rounds` and no `acquisition` key, and `acquired_chains` is `{}` — ps and native agreed on the target's parent (both 84959) at first sight. Scope explicitly assigns the race cases to the offline suite, so this is accepted, not an omission.

**Validation record.** `validation-f7e0f1…json`: `git diff --check` 0, HIR guardrails PASS, file-size PASS (3767 files, limit 900), largest maintained external source 611 lines (`coverage_symbols.py`). `gates: 0` with the correct stated reason.

## Follow-up findings

- **suggestion** — `E/coverage_native.py:129`: `require(pid != self.root, 'root disappearance ambiguity')` is now unreachable-as-effective, because the identical check at `:127` runs unconditionally for every native-absent subject. Harmless, but the duplicate obscures which guard is load-bearing.
- **suggestion** — `E/coverage_b38_tests.py:8,9,10`: `DEBUGSERVER`, `replay_document` and `validate_rounds` are imported but never referenced in the module body (`validate_rounds` is reached only via `confirmed_departures`). Dead imports.
- **suggestion** — `E/coverage_custody.py:299`: `chain = acquired[pid]` shadows the ancestry `chain` bound at `:291` inside the same loop iteration. Currently safe because `:299` is the last use before the iteration ends, but it is fragile against any future insertion below it; a distinct name would remove the hazard.
- **suggestion** — `E/coverage_acquisition.py:149`: the `receipt[field].get(pid)` int-key fallback is dead, since `event_history()` (`:16`) always emits `str(pid)` keys. Only the `events` tamper case at `coverage_b38_tests.py:243` needs the dual lookup, and that is on the tampered document.
- **test-coverage suggestion** — `E/coverage_acquisition.py:140`: the `new unobserved acquisition descendant` guard is implemented and reachable (a new root descendant appearing in a fresh round is rejected via `observed_rows` → `custody.observe`), and the `/bin/ps` probe PID is correctly exempted per round, but none of the 50 named B38 cases drives it directly. Not part of the approved evidence enumeration, so out of scope for this item.
- **pre-existing issue** — `E/coverage_acquisition.py:105`–`:107`: `validate_rounds` returns early when `acquisition` is absent, so on the no-reconciliation path (the shape of this entire live proof) the new `initial_sample` receipt and the `native['end'] - native['begin'] <= 6` total bracket are never asserted. This mirrors the inherited B37 `confirmed_departures` shape and is not a regression; the per-subject `before`/`after` evidence on that path is still fully validated by `checked_record`.
- **infrastructure issue** — The changed F2/F3 producer path (`reconcile`/`batch`) carries zero live coverage from this qualification, since the scheduling never hit a ps/native disagreement or an in-flight departure. Scope explicitly accepts this and assigns the races to the offline suite; worth carrying into the phase record so a future item does not read this proof as live qualification of the batch path.

No new requirements, no broad revalidation, no whole-phase review. B24 causal/representative budget, B27 joint source delivery/builtin fix/exact65, SQL and the original emitted-Rust phase remain OPEN and untouched.

PR #3840, opened 2026-09-16
Establish DX.1 compiler measurement lanes and repair LSP transport

yaseralnajjar:
This establishes DX.1 baseline and policy contracts before compiler optimization: installed optimized and contributor compiler lanes bind actual artifacts, toolchains, profiles and sysroots; legacy budget mappings preserve active comparable guards; fixture/assertion authorities and the demanded-stdlib memory workload are frozen.

The canonical LSP measurement client now handles coalesced and fragmented frames with deadline-aware reads and retains the original failure when cleanup also fails. Benchmark report paths support evidence outside the reviewed tree.

Validation covers candidate 5c7502f7aea3cd887a217c2fb3e289c7a1f2c653:
- Seven independent LSP transport regressions pass.
- Benchmark self-test, including Q07/Q08 and seeded budget failures, passes; manifest/schema checks and file-size guardrail pass (unchanged-input evidence reused where applicable).
- Installed optimized baseline passes CLI/native and 21 demanded-stdlib LSP samples (one warmup), coefficient of variation 0.0059124.
- Contributor CLI/native, unchanged/edited sessions and bare Cargo stdlib-cache tests pass. Completed contributor evidence explicitly reuses unchanged successful rows with paths/digests and keeps failed preparation/report-display captures intact.
- No compiler, lockfile, actual fixture or workflow changes; broad gates are omitted under the user's explicit batch policy.

Evidence root on the owned remote host: /home/yaser5/projects/sifr/dx1-evidence.
Product report SHA-256: 63630af41872a6592af2a242336b99b0228ad938a1adad76e389d730d03e7994.
Consolidated contributor report SHA-256: 13d1ae567054848ec6329dc24ad6b58daab7f0db80b068961919661858fb9393.

This records pre-migration observations; it does not claim optimized targets or implement DX.2. Scoped Opus review: SATISFIED, no blocking findings. Review: https://github.com/sifr-lang/sifr/pull/3840#issuecomment-5705061104. Post-merge records completed in #3841.


yaseralnajjar:
I reviewed the batch read-only: the full base→candidate diff, the transport implementation and its 7 regression tests (re-run at the candidate: OK), the DX.1 lane/policy code (`compiler_lane_tests.py` re-run at the candidate: Q07/Q08 negative controls pass), the frozen contract (`dx_contract.json`), and the evidence root — verifying digests, preserved failures, and Q09 sample integrity.

**Traceability spot-checks performed**

- Transport: `_read_message` now drains `_receive_buffer` first and only calls `_read_chunk` when bytes are actually needed; `_read_chunk` recomputes `remaining` per call, treats `select` non-readiness as timeout, and uses `os.read(fileno)` so no frame can be stranded in a `BufferedReader`. Header and body stages share one deadline. No `.read()` on `process.stdout` remains anywhere in the file, so the coalescing defect cannot recur through another path. `close()` captures `sys.exception()` and demotes cleanup failure to `add_note` only when a primary exists, re-raising otherwise — both directions are tested. No timeout was raised, no protocol operation dropped, no alternate client, no compiler edit (`git diff` touches no Rust/lock/fixture/workflow).
- Q09 inputs match the frozen contract exactly: 20 iterations + 1 warmup (`range(21)`, CV over `records[1:]`), `rss_poll_seconds` 0.01, operation sequence, `allocation_delta_bytes`/`decoded_counts` recorded `null` with an explicit unavailability reason rather than substituted by RSS. CV 0.0059 → `performance_status: "observed"`, `comparison_status: "baseline-only"`, report-level `"baseline observation; no optimized target claimed"`. Nothing is promoted to qualification.
- Evidence: `product-baseline-5c7502f7a/baseline.json` = `63630af4…` (functional pass, `source_commit` 5c7502f7a, 21 checks + native first/noop/edited + Q09, zero null RSS fields, 105 RSS samples/iteration). `dev-completed-5c7502f7a/baseline.json` = `13d1ae56…`, `fixture-inventory.json` = `201476d3…` (3215 records: 2436 fixture / 616 imperative / 158 declarative / 5 profile, pinned to the phase base with the application-profile authority named). All three digests match the claims.
- Consolidation is honestly framed, not a re-capture: `evidence_kind: "consolidated-completion"`, per-row `evidence_source` with `reuse` flags and row index, `original_capture` retaining `functional_status: "failed"`, artifact/lock/rustc equality asserted against the prior receipt, and `validate_report` re-run on the result. The three historical failed reports still carry their original digests (`4c34e566…`, `b27c216…`) and the dev cold-helper failure is retained; the 180 s helper deadline in `ensure_frontend_query_bench`/`ensure_sifr_binary` is unchanged — preparation was moved outside measurement and logged, not relaxed.
- Policy: `reference_host.execution_details` no longer hardcodes `build_profile: "dev"`; `validate_measurement_rows`/`assert_comparable` bind lane, compiler profile, application profile and verification selection per row; the Mac/dev-versus-Linux and dev-versus-product rejections are enforced and seeded negatively, while `budgets.json` stays fully mapped (65/65 IDs, digest-pinned) rather than disabled. File-size guardrail holds for every touched source file.

---

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `plans/roadmap.md`, `plans/phases/index.md` and the `Execution Status`/`Current Handoff` sections of `ad-hoc-compiler-dx-and-toolchain-reuse.md` still read "DX.1 blocked on LSP verification transport" and list the demanded-stdlib LSP and contributor baselines under "Missing acceptance", while `ad-hoc-lsp-protocol-buffered-reader.md` keeps `status: open`. This batch resolved that blocker and completed those baselines. The completed captures postdate the candidate SHA, so they cannot be digest-bound from inside it, and AGENTS.md directs the record update to follow the merge — hence not blocking. The post-merge record must correct the blocked-on-transport status in all three rows and add `product-baseline-5c7502f7a/baseline.json` (`63630af4…`), `dev-completed-5c7502f7a/baseline.json` (`13d1ae56…`) and `consolidate-5c7502f7a.py` to the evidence table.
- suggestion: `verification/areas/performance/dx_capture.py` hardcodes the Q09 sample count, poll interval, source path and CV bound that `data/dx_contract.json` freezes; it records `dx_contract_sha256` but never reads the contract, so a future contract edit would silently diverge from the capture. Reading the frozen values (or asserting them) would make the freeze self-enforcing.
- suggestion: `verification/areas/performance/check_budgets.py:89` rejects any `product-installed-optimized` results unconditionally, including when a named reference profile is selected. That is correct today (no product contract exists), but it means adding a product reference profile in a later milestone requires editing this gate, not just adding profile data.
- pre-existing issue: `compiler_lanes.configure` rejects `SIFR_SYSROOT`/`SIFR_SYSROOT_MODE` overrides only on the receipted path; the receipt-less `contributor-dev` default still inherits an ambient sysroot override. Pre-existing exposure (there was no check at all before), and out of scope here since DX.2 owns ambient-toolchain replacement.
- infrastructure issue: `verification/areas/performance/run_benchmarks.py` is now 877 lines against the 900-line guardrail. Under the limit, but the next addition will likely force a split by responsibility.
- pre-existing issue: dropping the `process.poll()` probe from `_read_message` means an EOF mid-frame now raises the bare `"LSP closed stdout while reading <stage>"` without the exit code / stderr tail that `_diagnostic_context` used to attach. In practice `close()` in the caller's `finally` still surfaces that context as a note on the primary error, so no diagnostic is actually lost in the measurement flows — only in direct `_read_message` use outside a client lifecycle.

PR #3844, opened 2026-09-17
Own generated caches and compiler subprocess lifetimes (DX.3)

yaseralnajjar:
Generated native/test artifacts now publish immutable entries into persistent storage with OS leases, same-filesystem staging, ownership checks, concurrent-winner validation, and pressure-based pruning that protects active readers and auxiliary roots. Existing native/test consumers use these entries; warm test hits rerun test binaries.

Compiler subprocesses retain bounded raw output and original cancellation/timeout causes while cleaning owned descendants. User programs stream complete output without compiler limits. Verification separates safety deadlines from performance budgets, rejects child status forgery, and retains child metrics. Initial doctor, cache inspection/pruning, and timing surfaces make setup and cache behavior inspectable.

Validation on candidate `63640031f7b66e067b2cbb26849277faed725668`: C01–C03, C05–C07, B10, R01–R03 acceptance and focused regressions pass (8 Rust storage/process tests, 10 Python process/report tests, actual cold/warm CLI consumers including umask 002 and 9 MB output, missing-payload recovery and pressure cleanup). CLI schema, production driver Clippy, formatting, step-budget self-test and source-size checks pass. Scoped Claude Opus 5 review: SATISFIED, no blockers. Optional all-targets Clippy exposed pre-existing test lints retained for separate follow-up.

Full create-PR/merge gates are prospectively deferred to phase end under the canonical Phase DX policy; no full-gate pass is claimed. Before/after preparation observations are functional evidence, not a host-sensitive speedup claim. Evidence is retained at `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx3-evidence/`, indexed by the exact candidate SHA. No DX.4 implementation.


yaseralnajjar:
**Verdict: SATISFIED**

All four prior blocking findings are remediated on `63640031f`, and I found no new regression that meets the blocking bar.

What I verified in the source (no re-running of supplied evidence):

1. **Umask/ownership** — `cache_storage::seal()` (`crates/sifr_driver/src/cache_storage.rs:107`) strips `0o022` from the producer-owned staged tree and is invoked first in `PendingCachedArtifact::commit` (`build/workspace.rs:83`), before `payload()` validation; staging/entry roots themselves are created through `directory()` at mode `0700` (`workspace.rs:219`, `:293`). Auxiliary roots are now created privately (`cargo_resolution.rs:143`, `rust_interop_probe.rs:171`, `rust_interop_probe_cache.rs:54`), and `inspect()` skips any family lacking a `.locks` namespace into `protected_roots` (`cache_storage.rs:212`), so inspection/prune no longer walk or delete them; `size()` now checks only uid. The permanent regression test sets `umask(0o002)` inside its dedicated child and asserts an auxiliary root survives pruning (`workspace_tests.rs:17`, `:53`).
2. **User programs** — `run_program` (`process_execution.rs:80`) inherits stdio, applies no deadline and no `STREAM_LIMIT`, preserves exit status, and disarms group teardown only on normal exit; cancellation still terminates the owned group and returns a cause-bearing `ProcessFailure`. `cmd_run_file`/`run_binary_artifact` use it and map spawn errors to the same exit code as base (`failure_exit_code` default 2 == `EXIT_USAGE_OR_CONFIG`). Compiler paths keep bounded capture plus deadlines, with raw bytes retained in `ProcessFailure`.
3. **Deadline vs budget** — `step_budgets.py` no longer exports `SIFR_VERIFY_SAFETY_DEADLINE_SECONDS` from `budget_ms`; the safety bound is the generic 2400 s default in `profile_commands.py:30` (largest configured budget is 1 200 000 ms), and the contract line is labeled `kind=performance_budget`. `ProfileRunner` tracks `functional_exit_status`/`performance_exit_status` separately and `run_profile_with_report` writes both statuses independently.
4. **Child metrics** — the `[child:stdout|stderr] "<json>"` decode in `reports.parse_log:227` runs strictly after the `LANE_STEP`/`LANE_STEP_BUDGET`/`LANE_STEP_CACHE` branches, so forged lane status cannot authorize aggregate status while case timings, e2e timing/group stats and artifact-cache counters survive.

Prior suggestions are also addressed: reference-counted `sigaction` save/restore with per-scope cancellation reset (`process_signals.rs`), post-conversion `valid_entry` revalidation on the reader-hit path (`workspace.rs:261`), and a human vs `--json` `cache inspect` surface. Touched files are all well under the 900-line guardrail.

Blocking findings:
- none

Follow-up findings:
- **suggestion**: `crates/sifr_driver/src/build/rust_interop_probe.rs:171` applies the cache-root policy (`cache_storage::directory`) to a *user-specified* `CARGO_TARGET_DIR`. That path rejects any symlinked ancestor component and any pre-existing leaf with group/other-write. A developer who symlinks their target dir, or who (under umask 002) already has `…/rust_bridge_probe_target` created by cargo on the base binary, gets a hard `unsafe Rust probe target` failure on @rust-interop builds where base worked. Consider restricting the strict policy to driver-owned cache paths and only asserting uid/symlink on user-configured roots.
- **suggestion**: `crates/sifr_driver/src/process_signals.rs:38` installs handlers with `sa_flags = 0` (no `SA_RESTART`), while `capture()` (`process_execution.rs:14`) and the `try_wait()` loop propagate `EINTR` verbatim. Linux normally delivers to the main thread, but if the signal lands while the main thread is inside `waitpid`, or on a capture thread blocked in `read`, the cancellation `ProcessFailure` (cause + partial streams) is replaced by a bare `Interrupted` error — the exact DoD property this path exists for. Set `SA_RESTART` or retry `EINTR`.
- **suggestion**: `verification/runner/sifr_verify/profile_reporting.py:147` derives `functional_status` solely from `runner.functional_exit_status`, which is only assigned inside `ProfileRunner.execute_step`. A nonzero lane status originating elsewhere (summarization failure at `:129`, release-report failure at `:139`, or an exception raised by `prepare_step_budget`/`record_step_success` outside `timed_step`) yields `exit_status != 0` with `functional_status: "pass"`. Fall back to the lane status when no step recorded a failure.
- **suggestion**: the generic safety bound is a bare `"2400"` literal in `profile_commands.py:30` — only 2× the largest configured cold budget (1 200 000 ms). Worth naming it as a policy constant with the DX.1 rationale so the separation stays auditable.
- **pre-existing issue**: the `--all-targets` clippy lints in `stdlib_interop_startup_tests.rs:150` and `stateless_python_codegen_tests.rs:250/260` remain, unrelated to this diff.

PR #3848, opened 2026-09-17
DX.5: Add indexed stdlib schema and bounded decoder

yaseralnajjar:
DX.5 defines a private indexed stdlib container and explicit payload schema without serializing live compiler objects. Signatures use shared type references, nominal instances preserve declaration/package/view identity, and type parameters retain binder identity. The bounded decoder checks compatibility, offsets, digests, typed references, structural cycles and source ranges; its cache preserves active handles and evicts only unreferenced records.

The checked consumer inventory classifies current semantic, codegen, mutation, clone and lifetime sites. Metadata production and normal CLI/LSP consumers remain assigned to DX.6/DX.7.

Validation on `c7734025c242631a326901ce08c846ac42afd7ae`:

- 14 focused metadata tests pass for I05, I06, M06 and M16, including all 134 record families, generated cases, malformed seeds, package/binder/view identity and recursive graphs.
- The 500-field class scaling case grows from 167,449 to 677,209 encoded bytes for 10 versus 1,000 signatures; loaded signatures retain 11 versus 1,001 records without loading the class view. Allocation figures are conservative bounds, not RSS measurements.
- Scoped clippy, formatting, payload/site inventory, file-size guardrail and whitespace checks pass.
- Evidence: `/home/yaser5/projects/sifr/dx5-evidence/c7734025c242631a326901ce08c846ac42afd7ae/receipt.json` on the owned Linux host. Under the approved Phase DX exception, the full merge gate runs once at phase end.

[Scoped Claude Opus 5 review](https://github.com/sifr-lang/sifr/pull/3848#issuecomment-5712915895): SATISFIED, no blocking findings.


yaseralnajjar:
Scoped Claude Opus 5 review for `c7734025c242631a326901ce08c846ac42afd7ae` (base `3f070fcdbc8a205104291d32b91edd8e3c59b12f`). Evidence is retained outside the reviewed tree.

Reviewed read-only against the stated scope, DoD, and evidence receipt.

## What I verified

- **Schema completeness / no silent omission**: All 134 record kinds are unique, contiguous 1..=134, and every one has a `decode::references` arm (134/134). I independently diffed enum variants between `crates/sifr_ir/src/{hir_nodes,specialization_metadata,rust_interop,python_interop,template_strings,sql_queries,sql_migrations}.rs` and their wire twins — no missing enum or variant (the Python checker only compares struct fields and family-name sets, so this closed that gap). I also hand-mapped `sifr_type_system::Type` (51 variants) onto `metadata::Type`: `Class/Protocol/Newtype/Enum` fields land in `NominalView` (name, identity, fields, methods, parent_class, enum_variants, newtype_inner), `TypeVar(String)` → binder+slot, `LiteralStr`/`LargeIntLiteral` → `Ref<Text>`, `FloatLiteral(f64)` → `u64` IEEE bits. Nothing dropped.
- **Bounded decode**: header (120B) / entry (92B) offsets match encoder writes exactly; `open` enforces size, count, kind range, reserved-flag zeroing, strict ascending IDs, contiguity (`offset != next`), non-zero bounded lengths, exact `decoded_bound` recomputation, no trailing bytes. `read_payload` allocates only `entry.len` after that check, and digests are verified per payload. No `unwrap`/`expect`/`panic!`/slice-index in non-test metadata code.
- **Identity / determinism**: `Ref::anchor` is length-delimited and kind-domain-separated; `intern` is content-addressed with conflict rejection; the encoder emits in ID order so payload bytes are allocation-order independent (I05 test shuffles 32 insertion orders and compares container hashes).
- **Retention**: `get` evicts only cache-exclusive `Arc`s under the same lock that issues handles; exhaustion by live handles is an explicit error, not an eviction of pinned data.
- **Graph validation**: iterative color/depth/work DFS with `Module`/`Declaration`/`Binder` (and `NominalView` edges) as boundaries — I traced the diamond and mutual-recursion cases; no false cycle reports.
- **Non-activation**: diff touches only `sifr_sysroot`, docs, and the new checker. `lib.rs` adds `pub mod metadata` and nothing wires it into a command path.
- **Inventory**: 1010 generated site rows spanning driver (274), codegen (324), lowering (130), frontend (152), analysis (29), lsp (98), cli (3), with read/enumerate/mutate/clone/borrow-retain classification, regenerated and diff-checked by `scripts/check_dx_metadata_inventory.py`.
- Largest touched source file is 790 lines. I did not rerun the covered gates.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_sysroot/src/metadata/container.rs:26-31` — the default `Limits` are internally asymmetric. `MetadataEncoder` accepts a record up to `record_bytes` (2 MiB), but `MetadataStore::open` rejects any entry whose `decoded_bound` (`4096 + len*32`) exceeds `retained_bytes` (64 MiB). Records larger than ~2 MiB − 128 B therefore encode successfully and then make the whole container unopenable under the same limits. Fail-closed with an actionable error, so not a defect today (no producer exists yet), but DX.6 should either derive one bound from the other or reject at `insert`.
- suggestion: `scripts/check_dx_metadata_inventory.py:83-99` — the parity guardrail covers `ExternalDefs`/`SemanticExports` and the seven `sifr_ir` groups, but not `sifr_type_system::Type` → `metadata::type_records::Type`, nor enum-variant sets anywhere. A future compiler `Type` variant or a new variant on a mirrored enum would not be flagged, and `record_cases.json` coverage is asserted only per-kind (`kinds.len() == KIND_COUNT`), not per-variant. Worth extending before DX.6 makes the mapping load-bearing.
- suggestion: `crates/sifr_sysroot/src/metadata/store.rs:161-186` — the pressure path re-sums every retained entry's bound on each exhausted `get`, i.e. O(cache) per pressured lookup while the retention mutex is held. Fine at current sizes; may want an incremental counter once real stdlib volumes flow through in DX.7.

PR #3850, opened 2026-09-17
Implement DX.6 canonical metadata production and shared preparation

yaseralnajjar:
Implements DX.6 canonical indexed stdlib production through the existing source frontend/codegen. `sifr sysroot build-metadata`, linked driver tests, verification preparation and native packaging share one keyed write-through ensure operation with process/thread coordination, complete validation, atomic publication and retryable failures. Packaging binds metadata to the exact compiler and target. DX.7 consumer activation remains separate.

Validated candidate: `5581af0884edcb46d88419b9b7d66d1ce97531df`.

- M09/M12–M14 and canonical 89-module inventory: 11 passed; one ignored subprocess-protocol helper is exercised by the process tests. Includes repeated exact bytes, every successful waiter's independent assertions, two distinct configurations, producer death/failure/cancellation, source mutation, strict overrides and output diagnostics.
- R10: full bare `cargo test -p sifr_driver` with normal CLI paths unavailable and no injected verification environment, empty and prepared metadata caches: 649 passed / 78 ignored on each run. Wall times 266.45s and 242.41s.
- Real CLI: four cold/warm semantic target configurations, direct launch and output errors. Real Cargo-reported compiler preparation and native Linux archive verification pass; separately produced staged-source metadata matches the complete source-tree artifact identity.
- Constructor-view decoder, preparation and distribution regressions, focused driver clippy, generator/inventory/file-size guards and format checks pass. Unchanged focused test inputs reuse their existing evidence.

[Read-only Claude Opus 5 review: SATISFIED, no blocking findings](https://github.com/sifr-lang/sifr/pull/3850#issuecomment-5717512054). Nonblocking observations are deferred to the owning follow-up record, not absorbed into this producer item. Evidence and SHA-256 receipts are retained on the remote host under `/home/yaser5/projects/sifr/dx6-evidence/5581af0884edcb46d88419b9b7d66d1ce97531df/`.

The approved Global Execution Rules defer the full gate to phase completion; no intermediate broad gate or release qualification was run or claimed. Historical failed development evidence is preserved separately.


yaseralnajjar:
Read-only Claude Opus 5 review of DX.6.

Candidate: `5581af0884edcb46d88419b9b7d66d1ce97531df`

Base: `03c4b0d0671afe7e96f054cf0a39dcab63e56db9`

Review response SHA-256: `ae9dab23539f905748bda3efbce525c74e7c3fd74b2c9b6208e3d70c09d2afb2`

Verdict: SATISFIED

**Basis (verified directly against the diff and evidence):**

- Producer is source-only and linked: `crates/sifr_driver/src/metadata_producer/production.rs:25` captures the canonical inventory via `load_stdlib_tooling_sources_from_sysroot` and rejects a non-source-tree sysroot (`crates/sifr_sysroot/src/layout.rs:21` `SysrootMode`); no `build.rs` references the producer, and the packaging/verification adapters invoke the compiled CLI or the already-linked test entry, never Cargo from inside the producer (`verification/runner/sifr_verify/metadata_setup.py:57`, `cargo_crate_setup.py:26`).
- Identity/binder/export completeness: input identity binds semantic target, producer policy, mapping-component tokens, every module/kind/relative path/source, `sysroot.toml` and `Cargo.lock` (`production.rs:57-95`); exports resolve to a real checked origin or fail (`exports.rs:4-46`, no invented signatures); nominal declarations keep the declared kind while class-shaped constructor views are accepted narrowly (`crates/sifr_sysroot/src/metadata/validation.rs:70-86`, regression at `metadata/tests.rs:666`).
- Publication correctness: bytes are decoded and `validate_complete()`-checked before staging; inputs are recaptured before and after staging; same-directory `create_new`+`O_NOFOLLOW` stage, `sync_all`, `rename`, directory fsync (`ensure.rs:176-205`). Failures/cancellation never enter `SUCCESSES` and release the OS lock by dropping the `File`; `remember` only memoizes successes (`ensure.rs:31-45`).
- Coordination: process-local keyed store plus per-key `entry_lock` with double-checked entry validation before and after ownership (`ensure.rs:120-170`), proven by M13 (4 threads: 1 production/4 assertions; 4 cold + 4 warm processes; two distinct configurations produce distinct keys) and M14 (staged producer killed, failing owner with two waiters, wait cancellation, cancelled-producer retry).
- Output/diagnostics: `publish_output` re-reads and digest-checks the prepared bytes before copying, and yields the required "output parent"/"writable output directory" diagnostics; read-only source with writable cache passes (`tests.rs:337-399`).
- Packaging/verification integration: `--require-native-target` gate, ELF/Mach-O native check before any execution, descriptor binding compiler bytes/identity/target/input identity/metadata digest, archive verification with a bounded extract (`scripts/distribution/metadata_artifact.py`, `verify_release_archive.py:194-201`). DX.7 consumers are untouched.
- Evidence checks out on this host: worktree HEAD is the candidate and clean; `final-evidence.json` digests match the files I hashed (`driver-final.log`, `bare-empty.log`, `clippy-focused-final.log`, `cli-final.log`); named driver run shows 11 passed / 1 subprocess-protocol ignored with the 89-module inventory; both bare runs are full 649-passed/78-ignored suites with no CLI binary present; file-size guardrail re-run by me passes (generated files carry a real `@generated` marker with a `--check` generator, so the exclusion is legitimate, not a bypass).

**Answer to the self-review question:** publication is validated — produced bytes are decoded and completely validated before staging/rename, `publish_output` re-verifies the digest before copying, and the `output == cache path` short-circuit only skips a copy of already-validated bytes. The allocation bound, however, is not preserved on the read side (below).

Blocking findings:
- none

Follow-up findings:
- suggestion: `crates/sifr_driver/src/metadata_producer/ensure.rs:53` (also `:127`, `:159`, `:248`, `:291`) reads the cached/override container with `fs::read` before `MetadataStore::open` enforces `Limits::file_bytes` (256 MiB, `crates/sifr_sysroot/src/metadata/container.rs:26`). A multi-GB corrupt or wrongly-copied entry is fully allocated before the bounded rejection, so the M06-style "no runaway allocation" property holds for the decoder but not for its DX.6 callers. Not blocking: the entry directory is uid-checked, 0700, symlink- and other-writable-rejecting (`cache_storage::payload`/`check_owned`), and the override path is test-only until DX.7. Correction is small — check `fs::metadata(path).len()` against `Limits::default().file_bytes` (or hand the `File` to `MetadataStore::open`, which already accepts `Read + Seek + Send`, and stream the digest) before reading. Worth closing before DX.7 exposes the override to ordinary consumers; `verify_release_archive.py:194` already does exactly this size-first check.
- suggestion: `crates/sifr_driver/src/compiler_context.rs:31` `with_metadata_override` is not part of the `(identity, sysroot)` stdlib-cache key. Harmless today because the override only validates, but two contexts with the same identity and different overrides would share one memoized stdlib owner once DX.7 makes consumers read the artifact.
- infrastructure issue: a failed or cancelled production leaves `{key}.stage` in the cache until the next production of that same key reclaims it (`ensure.rs:189-191`), and published metadata entries are pinned with no pruning. Both are consistent with the documented DX.6 boundary (`internal_docs/compiler_dx_architecture.md` §6.5) and general cache GC is out of scope, but per-configuration 18 MB entries will accumulate across test identities.
- suggestion: `internal_docs/architecture.md:2007` — the appended DX.6 paragraph runs the pre-existing sentence "Fragment validation identity never replaces final application checks." onto the end of a long line; purely cosmetic.

PR #3852, opened 2026-09-17
DX.7: share external baselines and load metadata on demand

yaseralnajjar:
DX.7 routes ordinary checks, builds, native tests and LSP analysis through demanded metadata modules. Tiny checks load semantic imports without whole-stdlib HIR or Rust reconstruction; code generation loads complete selected support and retains transformed/final Rust validation.

External definitions share an immutable baseline with isolated mutable project exports. Complete nominal payloads use shared storage with copy-on-write mutation, and metadata projections are cached by artifact-local identity. Navigation loads source text on demand. Installed execution requires matching metadata and has no source bootstrap fallback.

Validation on `036e69592164b87d309f31ba43ee406a58ed79ca` passes M01, M02, M04, M17 and focused source-overlay, nominal-sharing, exports, analysis and LSP regressions. Six representative fixtures preserve exact public emit bytes, newlines and preamble markers against both the source reference and installed source compiler; twelve native assertion executions and the installed native test runner pass. Missing/corrupt/incompatible metadata rejects and restored metadata retries. Selected source removal does not break installed checking.

Paired optimized-installed DX.1 workloads, 20 measured samples after one warmup: steady/post-close RSS 128.262 → 65.350 MiB; peak RSS 128.264 → 95.766 MiB. Separate allocator sampling reports steady allocated chunks 89.364 → 44.054 MiB and post-close 72.234 → 43.258 MiB. Fresh checks improve 10.76%; LSP open/diagnose/close improves 10.67%, timing CV below 1.3%. All samples preserve the frozen DX.1 package diagnostic. These are workload measurements; allocator chunk counts include retention and are not exact object attribution. Candidate observes 13 demanded modules, 13 shared nominal projections and zero HIR-module/Rust payload reads. The scoped 128 MiB cap passes.

Evidence: `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx7-evidence/036e69592164b87d309f31ba43ee406a58ed79ca/evidence.json`, indexed by `final-evidence.json`; raw reports, command receipts and preserved failed/incomplete history remain outside the reviewed tree. [Scoped Opus 5 review](https://github.com/sifr-lang/sifr/pull/3852#issuecomment-5719741485) returned SATISFIED with no blocking findings on the exact candidate. Under the approved phase policy, no intermediate full create-PR/merge gate was run; the full gate remains at phase end. Full target/corpus qualification belongs to DX.8.


yaseralnajjar:
Scoped DX.7 review via [phase-closure-loop](https://github.com/sifr-lang/sifr/blob/036e69592164b87d309f31ba43ee406a58ed79ca/.cursor/skills/phase-closure-loop/SKILL.md) and [talk-to-claude-opus](https://github.com/sifr-lang/sifr/blob/036e69592164b87d309f31ba43ee406a58ed79ca/.cursor/skills/talk-to-claude-opus/SKILL.md).

Base: `963eed6419d8947822030451017fde6212a1b86f`
Candidate: `036e69592164b87d309f31ba43ee406a58ed79ca`
Response SHA-256: `1e159c464005c26e3df1beb40711b6783b78b2878f7cb08e9173be4fb1e91d79`
Prompt SHA-256: `547ac2c1a1c0b67012c000ade366e3ba561ae65ee6e974ec62fc37ffd1d3fef5`

Read-only Claude Opus 5 returned SATISFIED with no blockers after reviewing the named validation and paired installed evidence. Follow-ups will be retained in the record-only phase update; no implementation changes were made after this approval. Raw review and validation evidence remain outside the approved Git tree at `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx7-evidence/036e69592164b87d309f31ba43ee406a58ed79ca/`.

The empty allocator baseline increased by 0.107 MiB (0.070 → 0.177 MiB); steady, peak and retained totals decreased as reported. The frozen DX.1 package diagnostic is unchanged. No intermediate full gate or release qualification is claimed.

---

**Verdict: SATISFIED**

Blocking findings:
- none

Scope traced against the acceptance criteria (read-only; no files modified):
- **Ownership/layering:** `ExternalDefs` fields are all `ModuleMap` (immutable `Arc<HashMap<K,Arc<V>>>` baseline + mutable overlay) at `crates/sifr_lowering/src/lower/external_layers.rs`; `collect_module_exports` writes overlay only and calls `remove_module_overlay` before re-inserting (`crates/sifr_frontend/src/query_diagnostics.rs:485`). Lowering does no provider I/O — demand happens at the frontend boundary (`external_overlay.rs::prepare_external_defs`, `graph_cache_and_queries.rs:302`, `ensure_lowered` at `:708`). Staged order (source-backed first at `e7e89bab2`, then metadata activation) is confirmed by the commit range.
- **Stale export isolation:** `clear_module_caches` removes only the project overlay; the `source_overlay_failed_and_deleted_exports_are_invalidated` and `..._projects_share_only_immutable_baseline` tests cover failed/deleted/repaired and two-project pointer-identity isolation, and `dx7_m17_metadata_projects_keep_failed_deleted_and_repaired_exports_isolated` repeats it over metadata.
- **Artifact identity / error recovery:** the process-global `(identity, sysroot)`-keyed `OnceLock<Result<…>>` slot is gone (`compiler_context.rs`); the cache is per-`CompilerContext` and stores successes only, so a transient selection failure is not retained (`dx7_m17_failed_override_retries_and_never_reuses_another_owner`). Installed selection checks schema/compiler/target/target-id, hex-validates all digests, bounds the descriptor to 8 KiB, and `open_consumer` re-checks `metadata_id`; `MetadataStore::open` still rejects `actual != expected`.
- **Selected semantic/support completeness:** `semantic::project` writes every `SemanticExports` field (inventory script enforces field parity, now minus the non-payload `provider`), and `materialize` rebuilds signatures, class fields/templates, constants, generators, transitive deps and interop plan from the decoded HIR.
- **Preserved validation:** no bypass introduced — reusable Rust is checked for bytes digest, compiler/target identity, grammar identity and `FragmentBoundary::StdlibModule` (`support.rs:55-68`); templates are role/owner-checked against the demanded HIR inventory *and* completeness-checked (`templates.rs`); interop sources are length+digest-verified against the artifact. `for_codegen` routes every codegen consumer (`entrypoint.rs:143,723`, `project_codegen.rs:63`, `test_runner/orchestrator.rs:135`) through the same materialization.
- **Navigation lifetime:** `Weak`-valued cache keyed by stdlib root, swept on miss, source text behind `OnceLock` and digest/length-checked on first read; `loaded_sources()==0` asserted.
- **Evidence:** counter assertions (not elapsed time) are present — `loaded_semantic_modules()`, `decoded_count::<HirModule/RustPayload/NominalView>()`, `payload_read_count`, `projected_nominal_views()`, `shares_storage` — in `metadata_reader/tests.rs` (M01, M02/M04/M15, M17, navigation, nominal sharing). Installed report has 30 rows: 12 emit byte comparisons (metadata vs. the DX.1 installed reference compiler, sha-matched to the recorded `.source.rs`), 12 native executions, source-unavailable check, and missing/corrupt/incompatible/restored artifact handling. Q09 is paired and separate from dev/test measurements, with allocator caveats stated. I re-ran the read-only checks here: file-size guardrail PASS (4040 files), `sifr_driver`/HIR maintainability PASS, DX metadata inventory PASS, `cargo fmt --check` clean, worktree clean.

Follow-up findings:
- suggestion: `crates/sifr_lowering/src/lower/external_layers.rs:39` vs `:48/:54/:57` — read/write asymmetry. `get` gives the baseline priority, while `insert`, `entry`, `get_mut` and `remove` touch only the overlay. Today no caller mutates a baselined key (`insert_error_type` runs pre-freeze in `stdlib/re_exports.rs:229`; `take_module_specialization_metadata` is called with `"main"`; SQL profiles use synthetic `sifr.sql.schemas`/`sifr.sql.requirements` names), but a future mutator targeting a demanded stdlib module — or a project module literally named `sifr.*`/`_sifr.*` — would be silently discarded rather than rejected. A debug assertion in `entry`/`get_mut` when the key is baselined would make the invariant enforceable.
- suggestion: `external_layers.rs:29` — `freeze()` silently no-ops when the baseline is already non-empty, leaving overlay entries unshared. Double-freeze or freeze-after-extend degrades sharing without any signal.
- suggestion: `external_layers.rs:76` — `len()`/`is_empty()` iterate, and `iter()` does a baseline `contains_key` per overlay entry. Fine at current overlay sizes; worth noting before incremental-engine work raises call frequency.
- pre-existing issue: `crates/sifr_driver/src/metadata_reader/provider.rs:47` — `closure` silently skips names absent from the artifact. Combined with the existing `continue`-on-missing shape in `stdlib_demand_plan.rs:30,85` and `support_plan.rs:435`, a stdlib module missing from a valid-but-incomplete artifact degrades to absent Rust and surfaces only as a downstream `rustc` failure. The `continue` arms predate this item and completeness is otherwise enforced by `validate_complete`/template checks, but the failure mode is newly reachable.
- infrastructure issue: validation breadth is narrow by design under the phase exception — `cargo test -p sifr_driver dx7_` ran 6 of 733 tests, and the manifest has no clippy receipt (fmt, encoder, decoder, inventory, guardrail, diff-check only). Permitted here; the residual risk for a ~6.2k-line change rests on the phase-end merge gate and DX.8 corpus qualification.
- suggestion: `q09-comparison.json` — `heap_empty_bytes` regresses +153% (73 KB → 185 KB) from the pinned metadata container. Negligible in absolute terms and already implied by the third caveat, but it is the one metric that moved the wrong way and is not called out by name.

PR #3854, opened 2026-09-17
Qualify full metadata and pin installed toolchain generations

yaseralnajjar:
DX.8 adds complete metadata traversal, installed-toolchain qualification, and immutable installation generations so live sessions retain matching semantic records, sources and payloads. Doctor now reports selected metadata identity and explicit package integrity. Mutable legacy installations reject with an empty-root remedy; no migration or release is introduced.

Full qualification exposed and fixes private imported constant projection, local generic-class shadowing, builtin-open payload demand, protocol ordering, host-derived semantic identities, and CPU-dependent NaN HIR bits. Actual macOS qualification also fixes Darwin statvfs block-count widening.

Validation covers M03, M05, M07, M08, M10, M11, M15, R09, Q05 and Q09:

- 89 canonical modules and 50,410 complete records per target; four-target source/metadata exact Rust and source-map comparisons, 727 cases each.
- Actual native Linux/macOS producers: all complete portable record streams equal byte for byte under each matching semantic target. Compiler envelopes remain strict; no foreign executable runs.
- 727 native E2E cases pass. Exact emitted-input/runtime equivalence retains these behavioral oracles after the NaN repair; actual installed native focused cases also pass.
- Actual installed public emit: 727 exact stdout/newline/preamble/source-map cases; relocation, live LSP generations, rollback, rejection/recovery, lazy source access, installer/self-update tests and guardrails pass.
- Paired optimized installed Q09: steady RSS 128.36→65.30 MiB, peak 128.36→95.71 MiB; LSP latency −9.88%, fresh check −10.40%; measured CV below 2%, candidate under 128 MiB. Separate heap cohorts and fixed-input limitations retained.

Evidence: `/home/yaser5/projects/sifr/dx8-evidence/qualification-manifest.json`, SHA256 `59c1578ad47ba67ef3898ca5852b7d8c579c3c56db9ff486aa85270cb2aa23b1`, candidate `53e9a8f05ccebd794311cef39c34c0059ceb98b4`. Failed and interrupted attempts remain preserved. Successful installed stages are joined only after exact compiler/metadata/payload equivalence checks; the earlier unsupported standalone `--offline` invocation is not relabeled as passing.

The canonical prospective Phase DX policy defers full create-PR/merge gates to phase end. No release was requested. Scoped Opus review: [SATISFIED, no blockers](https://github.com/sifr-lang/sifr/pull/3854#issuecomment-5722224549).

Final candidate: `78f12c5cdd357f1601926f3fa78ec96f004108dd`. Post-review corrections change only four verification files: executable installer fixture, committed profile-wired development doctor regression, and generation-pinned receipt assertions. The real distribution area passes all 59 cases with pinned Node/npm; metadata-structural and the final native Mac revalidation pass. Compiler/product/corpus inputs are unchanged, so the original product evidence remains valid. Final evidence wrapper: `final-evidence-78f.json`, SHA256 `2ac0d40ea6b501011a961145e613820d3fd48d031e52a7c3c57d59de492d5f0f`.


yaseralnajjar:
Scoped Claude Opus 5 review of exact candidate `78f12c5cdd357f1601926f3fa78ec96f004108dd` against `e96028d97fdc2efe4389485dfcb88981d17f32a5`. Final response SHA256: `3f859a2ef2524dda2648215792aba88033c44a04b666d0b5c4924f82316a0fb9`. First review found two blockers; both are corrected and verified below. Nonblocking observations are deferred to the phase follow-up record.

Verification complete. All evidence hashes match the wrapper (`2ac0d40e…`), and the wrapper's own SHA256 matches.

**What I verified directly (read-only):**
- `git diff 53e9a8f05..78f12c5cd` touches exactly 4 verification files; no compiler/product/script/lock/vendor change.
- Blocking #1: `git ls-tree 78f12c5cd` shows `metadata_immutable_generation.sh` at `100755`, matching all sibling cases. It executed in the real harness — `78f12c5cd-distribution-representative.log:101,108` `status=pass`, and the full area run is `total_variants=59, total_failures=0`.
- Blocking #2: `verification/areas/sysroot_release/development_metadata_doctor.py` is committed and reached from `runner.py:192-200` inside the `metadata-structural` branch, which is wired in `create-pr.json:839`, `merge.json:724`, `nightly.json:732`, `release.json:723`. All five cases drive the real CLI (`missing`/`ensure`/`ready`/`corrupt`/`recovered`), and cover every gap named in the first review: `build-metadata` remedy text, `compiler_build_id` in the message, source root in the message, `development_metadata_path` selection (`selected.is_relative_to(cache)`), no silent production (no `*.sifrmeta`), no silent repair (bytes unchanged), and the `sysroot_cli.rs` early-return exit code 2 + JSON error envelope. `a4d278ea1-metadata-structural.json` is `pass`; its log shows the in-suite source compiler build; `committed-development-doctor/report.json` records all five real exit codes with `status: passed`.
- Regression check on the correction: `run_metadata_qualification` is only ever called with `metadata-structural`/`metadata-corpus` (`runner.py:134`), so restructuring the old `if result.returncode or suite != "metadata-corpus"` guard leaves the corpus path and its failure summaries unchanged. The receipt fixture's new assertions hold against the real layout — `immutable_installation.sh:44-53` symlinks `sysroot.toml` and `bin/sifr` through `.sifr-current`, so the retained `sysroot.toml` digest check resolves into the generation rather than a stale mutable file, and `binary_path.parent.parent == generation` is consistent with the `../.sifr-current/bin/sifr` link. No migration or back-compat path was added.
- `mac-final-78f/report.json`: `status: passed`, all 4 targets, 50,410 records each, revalidated on unchanged bytes with a byte-identical native binary.

---

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- infrastructure issue: `verification/areas/sysroot_release/runner.py:195` — `metadata-structural` now calls `build_source_sifr()`, adding a source-tree compiler build to that suite (13m03s cold in `a4d278ea1-metadata-structural.log`). In `merge`/`nightly`/`release` this shares the target dir with `boundary-equivalence` and is mostly cached, but `create-pr.json:839` wires only `metadata-structural` for this area, so that profile gains a cold build it did not have before. Correctness is unaffected; worth a look if create-PR wall clock matters.
- infrastructure issue: `verification/areas/sysroot_release/runner.py:196` — `tempfile.mkdtemp(prefix="metadata-doctor-", dir=ACTUAL_ROOT)` is never removed, so each run leaves a full doctor evidence tree (plus a published `.sifrmeta`) under `target/verification/actual/sysroot_release`. Same pattern as the existing `metadata-installed-` dirs, so this compounds an existing growth source rather than introducing a new class of one.
- suggestion: `verification/areas/sysroot_release/runner.py` is now 891 lines against the 900-line guardrail (which passes, 4046 files). The next addition to this adapter will trip it; a responsibility split is a better answer than shaving lines.
- pre-existing issue: the four non-blocking suggestions from the first review remain open and were correctly not promoted to acceptance requirements — `production.rs:57` missing `sifr_rust_interop_catalog` portable token, duplicated cache-key derivation at `ensure.rs:100`/`ensure.rs:346`, the narrow exception tuple still guarding only the `metadata-corpus` `Qualification.installed()` branch (`runner.py:207`, while the new structural branch at `runner.py:198` widened to include `ValueError`/`KeyError`), and `integrity.rs:32-41,63,70` hardcoding sysroot-relative paths.
- pre-existing issue: `scripts/distribution/generate_version_installer.sh:19` — the `'%s/%s\n'` → `'%s/%s\\n'` change is still a no-op inside the unquoted heredoc.

PR #3856, opened 2026-09-18
feat(driver): reuse native Cargo roots and finalize artifacts

yaseralnajjar:
Closes DX.9 only. Ordinary build/run/test/probe requests now use compatible owned Cargo storage and ask Cargo for freshness before returning independently captured executable/runtime/debug bundles. Stable generated roots retain unchanged input mtimes, isolate same-name targets, and serialize mutation/capture across native contexts.

Current source-lock authority, trust, build-script inputs and native environment remain authoritative. Portable exports resolve the local generated root before freezing exact Git sources/locks. Python startup checks the actual loaded library path/content; source exports explicitly declare their external runtime contract.

Validation: I04, B01–B04, B07–B09 and focused regressions pass on `2f1f3fb215d5263c0d1fb49d8c3346aafa299b33`. This includes real error/lint/link/runtime failures, concurrent capture, revoked trust, changed native flags/build-script inputs, and independently compiled native/Python exports with an actual alternate-library mismatch.

Paired optimized installed measurements (first pair discarded): cold 12.628→12.707s; no-op 0.252→0.513s; edited 12.389→0.592s. Edited builds rebuild 1 rather than 31 units; no-op builds confirm 32 fresh units with zero rebuilds. The no-op increase buys the required current Cargo check.

The canonical user-approved phase-end policy defers the full gate; no intermediate full gate or release qualification is claimed. Supplemental Clippy has zero changed-file errors; unrelated existing metadata/Python findings are recorded in their owning issues. All historical failed attempts remain preserved.

Evidence: remote `/home/yaser5/projects/sifr/dx9-evidence/qualification-manifest.json`, SHA256 `deedd24f44e6ada7896c963af6ce2a442a6e1df93351fd72d1ee8eeb187a62ac`. Scoped Opus review will be attached before merge.


yaseralnajjar:
Scoped DX.9 review completed on `2f1f3fb215d5263c0d1fb49d8c3346aafa299b33` with **SATISFIED; no blocking findings**.

All named acceptance and focused evidence covers this candidate. Qualification manifest SHA256: `deedd24f44e6ada7896c963af6ce2a442a6e1df93351fd72d1ee8eeb187a62ac`. Original read-only Opus response SHA256: `38c3078487ec037b914c18cdde22057a78ad2e96315113e62dc90b49dca462d7`. Evidence root: `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx9-evidence/`.

The user-approved canonical phase policy defers the full gate; no full Clippy or release-gate pass is claimed. Follow-up suggestions will be recorded as separate work in the post-merge record, without changing this approved implementation.

---

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: the feature-enabled supplemental Clippy run reports 498 errors, all in unchanged `sifr_runtime` Python modules (`python/object_ops.rs`, `async_value.rs`, `buffer_ops*`, `arrow_ops.rs`, `callbacks/*`, `recursive_ops.rs`) and metadata-owner files. I re-derived the span ownership from `2f1f3fb21-scoped-clippy.jsonl`: zero error spans land in any of the 18 changed source files. Correctly recorded in `ad-hoc-python-interop-qualification-dependencies.md` and `ad-hoc-dx-metadata-review-followups.md`; no full Clippy pass is claimed, and the phase-end gate owns it.
- suggestion: `rust_interop_probe.rs:102,186` — with the `cache_file.is_file()` early return removed (correct per 8.3), `probe_cache_file`/`mark_probe_cache_hit` now only write a receipt that nothing reads. The probe receipt is dead state; either drop it or give it an explicit observability role.
- suggestion: `materialize.rs:354-370` — in the cached flow, `materialize_binary_project_at_path_with_target` still copies the executable to `cached_binary_path(project_root.parent(), project_name)`, i.e. `<family>/roots/<project_name>/target/release/<name>`, a path keyed by project name rather than root identity. It is redundant there (the finalized bundle is captured from `report.native_executable`) and is non-aliasing only because the family lease serializes it. Root-id keying would make the isolation structural rather than lock-dependent.
- suggestion: `test_runner/execution.rs:56-68` — the test-runner family key omits `normal_seed_cache_fragment()` and the Python environment that the binary family key includes, and passes `generated_project.cache_scope` uncanonicalized while `materialize.rs:210` canonicalizes. Not a correctness defect (the lock lives in the per-scope root, and the test path never bound `PYO3_PYTHON`), but the asymmetry will produce duplicate roots for symlinked scopes.
- suggestion: `native_storage.rs:127-137,181-194` and `python/loader.rs:24-26` — every bundle file is read in full twice (identity, then capture), and the generated Python binary SHA256s the entire `libpython` on every process start. Measurable fixed startup cost for large interpreters; an mtime/size pre-filter or a mapped-read would keep the same guarantee.
- infrastructure issue: `native_storage.rs:33,72` — the family and publication leases are unbounded blocking exclusive `File::lock()` calls with no timeout or ownership diagnostic, so a wedged sibling process stalls a build silently. Consistent with the existing `cache_storage` primitives, so DX.3 storage ownership is the right home for a bounded-wait/diagnostic policy.
- infrastructure issue: runtime-library capture (`is_runtime_library`, `.so`/`.dylib` only) and the `$ORIGIN`/`@loader_path` loader script cover Unix only, and `validate_loaded_library` returns an error under `#[cfg(not(unix))]`. Unreachable today — `cache_storage.rs` is already `std::os::unix`-only, so the whole driver storage path is Unix-bound — but the Windows gap should be closed together with that, not separately.

Evidence I independently confirmed: manifest SHA256 `deedd24f…a62ac` structure and per-log digests; clean worktree at `2f1f3fb2`; all 9 `dx9` named tests plus the loader alternate-process test pass on the exact candidate; measured `edit` 12.39 s → 0.59 s with rebuilt units 31 → 1 and noop fresh units 0 → 32 (noop cost rises 0.25 s → 0.51 s, which is the required always-consult-Cargo behavior, not a regression); CLI report shows real family repartition on flag change, failure on invalid target, restoration, and repeated failing tests re-executing; both portable exports build offline `--locked` and run, and `python-portable-loader-mismatch` exits 1.

PR #3858, opened 2026-09-18
Implement DX.10 application profiles and compiler configuration reuse

yaseralnajjar:
Default `build` and `run` now select development applications, `test` selects the test profile, and all three accept explicit `--release`. Generated manifests pin overflow/unwind behavior, native reuse includes profile identity, and reports expose Cargo's actual application profile. Generated application, test and bridge-probe workspaces own their profile authority independently of the caller's Cargo project. Compiler distribution optimization remains independent.

Compiler integration preparation and execution share one configuration plan while retaining native, minimal, feature and target isolation. Release selections remain explicit, including the complete 727-fixture native run-pass corpus; development coverage is additional. All owned native verification consumers retain their release selection, including performance, sysroot, Python, differential, golden and hardening callers. Binary consumers read finalized artifacts. The consumer inventory and 21 actual operations qualify this migration while reusing the unchanged installed ff1 compiler.

DX.10 B05, B06, B11, B12 and R06 are covered on the 12 GB Linux host with Rust 1.98.1: 727/727 actual release fixtures, 5/5 development fixtures, fresh installed-compiler semantics/FFI/profile measurements, default CLI checks, unsafe/malformed caller-manifest and nested-cache probe regressions, configuration/assertion inventories, concrete grouped artifact identity, and focused guardrails. Every measured warm profile return rebuilt zero units; each edit rebuilt one application unit. Unchanged-input reuse is explicit.

The size consumer's predetermined real decrease/increase controls pass with expected exits 0/2. A separate real-ref whole-file size comparison remains a **failed** non-increase observation: +6,720 bytes (+0.10%), in non-allocated debug/string metadata; `.text` is unchanged. Its assertion was not relaxed, its failure is preserved, and no whole-file non-increase is claimed for DX.10.

Candidate: `d4b80812086538f6d26bf7830f30da54d52bffb0`. Evidence: `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx10-evidence/d4b80812086538f6d26bf7830f30da54d52bffb0/evidence-summary.md` and `qualification-manifest.json`. The [adjudicated final scoped review is SATISFIED, no blockers](https://github.com/sifr-lang/sifr/pull/3858#issuecomment-5725527463); both prior failed reviews and the explicit continuation adjudication remain recorded. The canonical prospective phase-end policy requires named acceptance here; the full gate remains deferred to phase completion.


yaseralnajjar:
DX.10 preserved review history and adjudication.

Initial candidate `6b093deaadc38f569dbaf7ce7b224e21503ac224`: **NOT SATISFIED** (R1 release consumer/path migration, R2 generated-manifest authority).
Remediation candidate `ff1f3b84b2b0ba9f174b2623a1837e90ff086ffb`: **NOT SATISFIED** (R2 satisfied; R1 repeated at additional first-party consumers). The agent stopped under the phase-closure-loop rule and the orchestrator explicitly adjudicated bounded continuation of the same mechanism, a complete consumer inventory, focused regressions and one additional exact-SHA review. Both failures remain preserved.

Current candidate `d4b80812086538f6d26bf7830f30da54d52bffb0` completes the first-party native consumer migration, including the additional differential/Python/golden/hardening/namespace callers identified by the ownership inventory. It has 13 focused consumer/profile checks, 8 generated preparation checks and 21 successful actual operations (16 observed native release application profiles). No compiler/runtime/stdlib/lock input changed after ff1; named corpus/semantic/profile/grouping evidence is explicitly reused. Final adjudicated review is pending.

The separate whole-file size non-increase observation stays **failed** (+6,720 bytes); the second reviewer confirms it is not a new DX.10 non-growth acceptance requirement. Actual decrease/increase consumer controls pass expected exits 0/2. Ambient uv0.12.5 failed the existing Python project's pin; owned uv0.12.10 corrected only the test environment, and the actual live-package build then passed. Neither failure is reclassified as passing.

<details><summary>Original first review</summary>

I reviewed the base→candidate diff against the DX.10 scope and the linked evidence (release corpus/artifact inventory, configuration inventories, metrics), and verified in-scope mechanisms directly in the tree: profile selection and identity (`ApplicationProfile`, CLI `--release`, `CompilerContext`), artifact identity (`NativeBuildContext.profile`, test-runner cache key), captured-override validation (`validate_application_profile`, `native_profile::validate_flags`), retained native/release selections (`application_profiles.json`, `area_adapter.py`, `run_pass_suite` complete-corpus guard), grouping inventory equivalence (replayed `configuration_plan` across 4 profiles × 2 modes), and preparation/execution command derivation.

Verdict: NOT SATISFIED

Blocking findings:
- regression: `verification/areas/performance/tools/check_codegen_binary_size.sh:96` (and `:98`), criterion "Implement explicit … profiles throughout commands, manifests, effective configuration, artifact identity and reports" / retain existing release assertions. The finalized-copy path moved from `target/release/<bin>` to `target/final/<bin>` (`crates/sifr_driver/src/build/materialize.rs:280`) and the default `sifr build` profile moved to development, but this first-party performance-area tool still runs `sifr build` with no profile flag and then reads `$out_dir/sifr_output/target/release/sifr_output`. That path no longer exists in any profile, so `measure_size` hits `error: binary not found for baseline` and exits 1; even once the path is fixed, the tool would compare unoptimized development binaries and no longer measure the release binary-size budget it exists to guard. Correction: pass `--release` in `measure_size` and read `$out_dir/sifr_output/target/final/sifr_output` (and `.exe`), consistent with the `area-native` "release" selection already declared in `verification/runner/application_profiles.json` and applied in `area_adapter.run_sifr_variant`.
- regression: `crates/sifr_sysroot/src/native_context.rs:432-442`, criterion "validate effective overrides" without conflating compiler/host project profiles with generated application profiles. `validate_application_profile` walks `invocation_root.ancestors()` — the user's process CWD, not the generated Cargo root — and applies the application unwind/overflow boundary to the first `Cargo.toml` it finds. That manifest does not govern the generated application's profile (the build runs with `--manifest-path <generated>/Cargo.toml`, and the generated root carries its own pinned `[profile.*]` tables), so the check validates the wrong authority in both directions. Failure scenario: a user runs `sifr build app.sifr` from inside any Rust project whose root manifest sets `[profile.dev] panic = "abort"` (or `overflow-checks = false`); the build is rejected with SIFR-BUILD-0005 "application profile requires panic=unwind and overflow-checks=true" even though that setting has no effect on the generated binary. The same loop also hard-fails with "invalid profile manifest"/"cannot read profile manifest" on any unrelated ancestor `Cargo.toml` that this `toml` parser cannot read. Conversely, when the generated project *is* absorbed into an outer workspace, the loop `break`s at the first ancestor manifest, which need not be the workspace root whose profiles Cargo actually applies. Correction: scope the manifest check to the manifest that governs the generated build (the generated project's own `Cargo.toml` and, if it resolves into one, its workspace root), and keep the captured `.cargo` configuration, environment and `RUSTFLAGS` checks as they are — those are genuinely CWD-derived and are correct today.

Follow-up findings:
- suggestion: `verification/runner/sifr_verify/profile_runner.py:342` — grouping dropped the previously deliberate `Planned crate test red-blocker <id>: must_be_executed_by=…` line (it was an explicitly reviewed reporting behavior). `configuration_plan` now filters those suites silently, and `dx10_profile_checks` excludes them from both sides of the inventory diff, so nothing detects the loss. Re-emit the line from `run_crate_tests` or from the plan.
- suggestion: `verification/runner/sifr_verify/profile_runner.py:22` — `crate_test_suites_for_mode` is now an unused import in that module; `cargo_command` is likewise no longer used for crate tests.
- suggestion: `crates/sifr/tests/e2e_support/batch_execution.rs:490` — the release-corpus guard indexes `authority["selections"][0]`, while the Python check selects by `id == "e2e-run-pass"`. Reordering `application_profiles.json` would silently compare against the wrong selection; select by `id` on the Rust side too.
- pre-existing issue: `crates/sifr_driver/src/tests/package_rust_interop_static_program_support.rs:95` and `crates/sifr_driver/src/tests/package_project_build_check.rs:474` still describe the layout as `<project>/target/release/<bin>` in their `expect` messages. The three-parent traversal is still correct, so this is stale text only.
- suggestion: `crates/sifr_driver/src/build/rust_interop_probe.rs:164` — `execute_direct_cargo_probe` selects the application profile for `cargo check` but does not call `validate_application_profile`, unlike `run_cargo_build` and the test-runner path; the probe can therefore run under an override the build would reject.

</details>

<details><summary>Original second review</summary>

## Verdict: NOT SATISFIED

**R2 is satisfied.** `validate_application_profile` now takes the manifest that governs the build (`crates/sifr_sysroot/src/native_context.rs:382-450`); both producers pass the generated root's own `Cargo.toml` (`materialize.rs:589`, `test_runner/execution.rs:183`), and both generated-manifest renderers already emit `[workspace]` plus the pinned profile tables, so the ownership guard cannot false-reject any current caller. Captured `.cargo`/env/RUSTFLAGS checks are untouched; the probe manifest stays valid TOML; the new e2e regression covers default/release × run/test under unsafe and malformed caller manifests with the cache nested beneath them.

**R1 is satisfied only for the one file the first review named.** `check_codegen_binary_size.sh:94-100` is correct and the growth assertion at `:127` is unchanged. Preserving the +6720-byte observation as failed is right — no DX.10 criterion asserts non-growth, B12 asks for measurement, and the tool is not wired into any gate. But R1 was remediated file-by-file instead of by criterion, and the default-profile + `target/release`→`target/final` change broke a cluster of first-party consumers; only one member was fixed.

### Blocking findings

- **in-scope omission** — `verification/areas/performance/run_benchmarks.py:623` (also `:436`, `:612-618`). Criterion: explicit profiles "throughout commands … artifact identity and reports"; arch §8.4 "Runtime-performance and release-qualification cases | Explicit declared profile, never an accidental default"; Q07. `command_for_case` builds with no profile flag (no `global_args` on any `mode:"build"` case), so all 15 `build-project` cases now produce a **development** binary while `:436` records `application_profile = "release"`; `collect_build_size_metrics` then stats `sifr_output/target/release/<bin>` and raises `BenchmarkError("… did not produce release binary")` — every build benchmark hard-fails. Correction: `--release` for `mode=="build"` and read `target/final/<bin>`.

- **in-scope omission** — `verification/areas/sysroot_release/runner.py:265` and `:521` (builds `:259`, `:514`) plus `attached_api_certification.py:57-72`. Criterion: "retain every existing release assertion". These release-qualification builds lost the release profile and still read `sifr_output/target/release/sifr_output`, which no longer exists; `run_checked([binary])` fails, so the stdlib-boundary, compile-smoke and attached-API certifications cannot run. Correction: `--release` on the three builds, read `target/final`; `metadata_qualification.py:100` is the same accidental default and belongs in the same pass.

- **in-scope omission** — `verification/areas/python_interop/runner/live_packages.py:172` (build `:130-145`). Same criterion. The live build gates on `target/release/<bin>.is_file()`, so every live case now reports `"failed to build compiled Sifr live binary"` regardless of the real exit code.

- **in-scope omission** — `verification/areas/performance/dx_capture.py:28` (build `:124`). Q07: `dx-native-*` rows hardcode `application_profile: "release"` while the captured command has no flag, making future captures incomparable to retained baselines.

### Follow-up findings

- **infrastructure issue** — `…/corpora/leetcode/benchmarks/bench.py:118-165` (submodule `sifr-lang/leetcode`) builds without `--release`, and its `executable_files` fallback silently benchmarks the single `target/final` unoptimized binary against `rustc -O`. Externally owned; record in that repo's owning issue.
- **suggestion** — `internal_docs/compiler_dx_architecture.md:505-516` still frames override validation as captured config/env/flags only; one sentence should state that the generated manifest is the profile authority and generated applications own their workspace.
- **pre-existing issue** — the five first-review non-blocking items remain open as agreed separate follow-up work.

No rescope is needed: findings 1–4 are the unfinished remainder of the already-adjudicated R1 mechanism, each correctable by the same two-line change per site. Review record written to `/home/yaser5/.claude/plans/read-only-second-scoped-remediation-adaptive-flute.md`. No repository files were modified and no validation was run beyond reading the tree.

</details>

Evidence host/root: `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx10-evidence/`. Current exact-SHA summary and hash manifest are in `d4b80812086538f6d26bf7830f30da54d52bffb0/`; prior raw responses and the adjudication remain outside the Git tree. No full intermediate gate or actual release was run.

yaseralnajjar:
DX.10 final adjudicated scoped review: **SATISFIED, no blockers**.

Exact candidate: `d4b80812086538f6d26bf7830f30da54d52bffb0`
Base: `b540824a208b47aab5bf15d254c9fa6ab6f22984`
Prior failed reviews/adjudication: https://github.com/sifr-lang/sifr/pull/3858#issuecomment-5725480175

B05/B06/B11/B12/R06 combine explicitly inherited unchanged-input corpus/profile/group evidence with the final consumer regressions. The complete 727-fixture release corpus, 175 actual hashed binaries, additional default-dev coverage and final installed semantic/FFI/metrics remain qualified. Fresh remediation evidence includes 13+8 focused checks and 21 successful actual operations with 16 observed release application profiles. No full intermediate gate was run under the canonical phase-end policy.

Evidence: `yaser5@yaser.tailaa73b4.ts.net:/home/yaser5/projects/sifr/dx10-evidence/d4b80812086538f6d26bf7830f30da54d52bffb0/`.
Review SHA256: `1c7a0acb69c19f56fdffec0b6ca899e27f89b98f37aa87457fd77b2b0b0c1857`.
Qualification manifest SHA256: `c22780206faa6168cf7f92ede0b742f652c732c05969de33b96d1eb1a90d7e93`.
The +6720-byte separate size non-increase failure stays failed; no DX.10 non-growth claim. Initial uv mismatch and both failed reviews remain preserved. Suggestions are deferred to owning followup records.

<details><summary>Original final Opus response</summary>

## Verdict: SATISFIED

The remediation diff `ff1f3b84b..d4b808120` is verification-only (18 files, all under `verification/`), and it closes all four blocking findings from the second review with the same R1 mechanism, no assertion relaxed:

| Second-review blocker | Candidate state |
| --- | --- |
| `run_benchmarks.py` build cases had no profile and stat'd `target/release` | `command_for_case` adds `--release` for `mode=="build"` only (`run_benchmarks.py:622`); `collect_build_size_metrics` reads `target/final` (`:625`); `check`/`fmt-check` unchanged, so the `application_profile="release"` record at `:436` is now truthful for all 15 build cases |
| `sysroot_release` boundary/compile-smoke/attached-API/metadata builds lost release and read `target/release` | `runner.py:259,265,514,521`, `attached_api_certification.py:60,73`, `metadata_qualification.py:100` all explicit `--release` + `target/final` |
| `live_packages.py` gated on `target/release/<bin>` | `:143` `--release`, `:173` `target/final` |
| `dx_capture.py` recorded `release` with an unflagged command | `:126` explicit `--release`, plus a guard at `:22-23` that refuses any `dx-native-*` sample without `--release` |

Independently verified beyond the named blockers:
- Repo-wide sweep for generated-application `target/release` consumers leaves only the external LeetCode submodule and *compiler*-binary paths (`cpython_differential/data/generated_seed_manifest.json`, `catalogue_lint.py`) — correct, since the compiler stays optimized. `{sifr} run` templates missing `--release` in any verification manifest: none.
- Newly migrated owners are the pre-DX.10-release ones (differential prep/execution/hand-seeded, integer readiness, python example/binding runtime, namespace `run` only, hardening `run`/`build` only, 4 golden runtime commands); `check`/`fmt`/`test`, the default-dev CLI tests and compiler Cargo invocations remain distinct. `area_adapter.py:575` release handling from `ff1` is intact.
- `native_release_consumer_checks.py` is wired into the runner self-test via `dx10_profile_checks.policy_checks` → `selftest.py:21`, so it is gate-selected, and its negative controls are real (compiler-only `--release` after `cargo`, stale `target/release` intermediate, mislabeled dev capture).
- I re-ran the focused suites read-only: 8 new consumer tests + 5 `ProfilePlanTests` = 13 pass; 3 + 5 = 8 generated preparation/selection tests pass. No repository file was modified and no broad gate was run.
- Crate/stdlib/Cargo/vendor blob identities in `consumer-profile-inventory.json` match the tree, so inheriting the `ff1` B05/B06/B11/B12/R06 evidence (727-fixture release corpus, 175 hashed binaries, dev/release/alternating storage) is legitimate; nothing in a consumer-only diff can move those inputs. The `+6720` whole-file observation correctly remains recorded as FAILED against no DX.10 criterion.

### Blocking findings

None.

### Follow-up findings

- **suggestion** — `verification/areas/runtime_platform/sanitizer_manifest.json` case `generated-binary-asan-smoke` (scope `generated_binary`, suites `sanitizer-smoke`/`sanitizer-full`) still invokes `… -- run demos/hello.sifr` on the new development default, so ASan now covers the unoptimized generated binary instead of the optimized one it covered pre-DX.10. Its assertions are exit/sanitizer-clean only, so no release assertion or record was weakened and no named criterion is unmet — but this is a first-party native invocation owner absent from `consumer-profile-inventory.json` entirely (not even in `unchanged_owners`). Worth one explicit decision (declare `--release`, or record the dev choice) in the owning follow-up.
- **suggestion** — `verification/areas/core_language/data/validation_suites/manifest.json` and `verification/areas/project_workspace/data/validation_suites/manifest.json` declare ~12 literal `sifr build`/`sifr run` argv executed through `area_adapter.py:286-306`. They are behavior-matrix cases with profile-agnostic assertions, so default-dev is defensible, but they are likewise unclassified in the inventory, which claims coverage of "all 10 literal application argv producers".
- **suggestion** — stale wording only: `run_benchmarks.py:648` still says "did not produce release binary" while stat'ing `target/final`, and `crates/sifr_driver/src/tests/package_project_build_check.rs:474` / `package_rust_interop_static_program_support.rs:95` `.expect` strings still read `<project>/target/release/<bin>`.
- **infrastructure issue** — `verification/areas/cpython_differential/checks/` tests only import when run from that directory (`python3 -m unittest test_generated_suite` errors from the repo root, passes from `checks/`). Pre-existing, unrelated to this diff.
- **pre-existing issue** — external LeetCode submodule profile migration, the five first-review non-blocking items, and the `+6720` whole-file size observation remain separate owning work, as stated.

</details>

PR #3860, opened 2026-09-18
Release closed editor state and reject superseded results

yaseralnajjar:
Incremental edits could update the frontend while leaving the session's authoritative overlay unchanged, allowing a later reload to restore old text. DX.11 now pins source/host/toolchain generations, commits overlay updates atomically, and rejects superseded responses, diagnostics and clears independently of cancellation. Explicit configuration changes recover metadata in the same LSP process, and closed projects release metadata and Python environment caches. Formatting, folding and selection work without semantic metadata.

Validation on candidate `c0d490d460ca4f8a286226da804b31744b5a0b3c`:

- Nine focused Rust regressions; installed E01–E06 transport tests, including all three position encodings, logical aliases, 20 stale terminal responses, and missing/incompatible metadata recovery; protocol smoke and an actual two-document push-progress regression pass. Incremental edits preserve diagnostic publication without workspace progress.
- Applicable predecessor evidence for toolchain upgrade/rollback, large-session and Q09 is reused with the exact remediation diff and capability/mode proof: progress-disabled clients already took the same no-progress path. Original dd282f9d measurement provenance is retained.
- The unchanged large-session smoke limits pass: 42 operations, p95 7.128 ms, peak RSS 53.1 MiB against 1000 ms / 128 MiB limits.
- Paired optimized installed Q09 cohorts on the 12 GB Linux host use 21 samples each, excluding the predetermined warmup. Median post-close RSS falls 65.51 → 47.80 MiB and allocated heap chunks 43.26 → 0.51 MiB. Steady/peak RSS show no improvement; those deltas, allocator retention and switched-project measurements are reported explicitly. Elapsed trace time is comparable.

Evidence is retained outside the reviewed tree under `/home/yaser5/projects/sifr/dx11-evidence/c0d490d460ca4f8a286226da804b31744b5a0b3c/qualification-manifest.json`. The Phase DX policy requires named tests and scoped Opus review at intermediate milestones; the full gate remains at phase end. No release qualification/publication or DX.12 implementation is included. The separate DX.10 size-budget failure remains unchanged in its owning issue.


yaseralnajjar:
## Review: DX.11 — Editor Snapshots, Recovery and Memory Discipline

Base `67005d246` → candidate `dd282f9d8`, branch `codex/dx11`. I read the full diff, architecture §10/§10.1, the canonical DoD, and cross-checked the external evidence (manifest hashes all verified, E01–E06 cases present, Q09 figures match the reported deltas).

**Verdict: NOT SATISFIED**

### Blocking findings

- **regression | `crates/sifr_lsp/src/notifications/mod.rs:164` (with `crates/sifr_lsp/src/diagnostics.rs:72`)** — criterion: "Preserve syntax-only responsiveness" / architecture §10 latency-sensitive editor work.
  `text_document_did_change` was changed from `publish_document` (one document, no progress scope) to `publish_all`. `publish_all` opens a `ProgressKind::FullDiagnostics` work-done scope whose threshold is 2 documents (`progress.rs:3`), so with two or more publishable open `.sifr` files in a push-diagnostics mode, every settled edit now emits a `$/progress` begin/end pair titled "Checking Sifr workspace" to any client advertising `window.workDoneProgress`. The generation early-return in `flush_ready` suppresses this only during an unsettled burst; a normal edit-then-pause publishes.
  Republishing dependent open documents after an edit is itself defensible (and arguably required by the stale-reconciliation goal) — the regression is the workspace-load progress scope on an incremental edit path. Correction: suppress the `FullDiagnostics` progress scope for change-driven reconciliation (e.g. a non-progress reconcile entry point, or open the scope only for genuine full/workspace publication such as `initialized`, `didChangeConfiguration`, `didChangeWatchedFiles`), leaving the publication behavior itself unchanged.

### Follow-up findings

- **suggestion** — `crates/sifr_lsp/src/diagnostics.rs:148` the comment "reconciles them without publishing every open file per edit" does not describe the code: the early return fires only when the captured generation is stale, so the settled-edit case does publish every publishable open file. Either the comment or the filtering should be corrected; there is also no "already published at this version" dedupe in `DiagnosticJobs::schedule`, so unchanged documents are recomputed and republished on each edit.
- **infrastructure issue** — the per-edit O(open-documents) publication above is unevidenced: the only multi-document latency fixture (`large-candidate.json`) runs with `diagnostics_mode: "off"` and `diagnostics_seen: 0`, and E01–E06 use pull diagnostics with at most two open documents. The 128 MiB/1000 ms budget therefore does not bound the new push path. Worth a scoped multi-document push-mode latency sample in DX.12–DX.14, not a re-run here.
- **suggestion** — `crates/sifr_lsp/src/requests/folding_range.rs:11` and `selection_range.rs:19`: the new `syntax_queries::facts` uses `parse_source_module`, which returns `Err` on any syntax error, so folding/selection now return `-32603` for transiently invalid buffers. The previous path went through `AnalysisHost::editor_facts`, which used `FrontendContext::parse_module` and degrades to `ParsedModule::empty()` (i.e. `Ok([])`). No information is lost either way and the terminal-response contract holds, so this is not blocking, but degrading to an empty result would match the prior contract and avoid client-visible error toasts while typing.
- **suggestion** — `crates/sifr_driver/src/compiler_context.rs:49` `refreshed_toolchain()` rebuilds via `CompilerContext::new`, silently dropping `metadata_override` and any explicitly supplied `ResolvedSysroot`. Unreachable today (the LSP workspace builds its context without either), but it is a latent hazard if the LSP ever gains a metadata-override or explicit-sysroot path; a debug assertion or explicit field carry-over would make the contract safe.
- **pre-existing issue** — the scope bullet "reduce demonstrated layout/clone/interner costs" produced no active-state reduction; the demonstrated win is post-close only (retained RSS −27.03 %, allocator chunks −98.83 %). The Q09 `attribution` field correctly explains this (same metadata container/index and 13 demanded modules in the active state) and claims no unmeasured win, which is what §10.1 requires, so this is a note for the DX.12–DX.14 baseline rather than an omission.
- **pre-existing issue** — the DX.10 F6 `+6720`-byte whole-file non-increase failure remains failed in its owning issue, unrelabelled and unweakened, as required. Confirmed in the manifest's `deferred` field; not touched here.

### What I verified as satisfied

Generation ordering is genuinely cancellation-independent: the message pump bumps the shared counter at ingress (`server.rs:97`) before the main loop applies the event, and both responses (`server.rs:231`) and diagnostic publication go through `Generations::publish`. I checked the request-queue drain hole specifically — `select_next_lane` imposes no in-flight cap and `drain_queued_requests` loops to exhaustion, so no request survives to be published under a later message's generation. Suppressed jobs are re-queued rather than dropped, and `diagnostic_clears` correctly refuses to clear a reopened document while still clearing on mode-Off transitions. Snapshot identity (`snapshot_owner` `Arc` ptr-eq plus `shares_metadata_generation`) prevents an old snapshot from querying a new host; `record_analysis_overlay` closes the overlay/reload byte mismatch; `retain_open_projects`, `refresh_existing_projects` and `without_cached_metadata` release closed-project environment caches and the idle metadata owner. Syntax handlers no longer construct a semantic host, matching E05 and the `metadata_stats().is_none()` unit assertion. All touched first-party sources are under 900 lines (largest: `host/implementation.rs` 884).

yaseralnajjar:
**Verdict: SATISFIED**

### Blocking findings

None.

The prior blocker is correctly and minimally resolved. `text_document_did_change` (`crates/sifr_lsp/src/notifications/mod.rs:164`) now calls `DiagnosticsController::reconcile_changes`, which routes to `publish_workspace(..., report_progress: false)` (`crates/sifr_lsp/src/diagnostics.rs:64-112`); `begin_progress(ProgressKind::FullDiagnostics, ...)` is never reached on the edit path, so no `$/progress` begin/end pair is emitted per settled edit regardless of `window.workDoneProgress`. This is exactly the "non-progress reconcile entry point, publication behavior unchanged" correction the first review prescribed.

Verified as unchanged elsewhere: `publish_all` (progress retained) is still used by `initialized`/`server.rs:142`, `workspace/didChangeConfiguration`, `workspace/didChangeWatchedFiles`, and `didClose`; `didOpen`/`didSave` still use `publish_document`. Job scheduling, generation gating in `flush_ready`, `flush_clears`, and every snapshot/analysis/metadata/closed-owner path are byte-identical to the qualified predecessor — `git show c0d490d46` touches only `diagnostics.rs`, `notifications/mod.rs`, and the new test. The misleading `flush_ready` comment (prior follow-up) is corrected to describe the actual generation early-return.

Evidence checked, not re-run:
- Working tree clean at `c0d490d46`; all eight manifest artifact SHA-256 values recomputed and match.
- `dx11-units-remediation.log`: 9/9 focused tests pass, including the new `dx11_incremental_push_reconciles_multiple_documents_without_workspace_progress` (2 open files, progress capability advertised, asserts 2 diagnostics + 0 progress on edit, and 2 progress notifications on `publish_all`).
- `push-progress-final/report.json` through the actual installed optimized binary (sha matches `product-final/receipt.json`, source commit `c0d490d…`): zero `$/progress` on `didChange`, begin+end `sifr/full-diagnostics-0` retained on the configuration-driven workspace publication.
- `acceptance-final/report.json` E01–E06 all present and passing on the same binary; `protocol-final.log` PASS; file-size (limit 900, largest touched `host/implementation.rs` 884) and HIR maintainability guards PASS; fmt log empty.
- Reuse proof independently confirmed: `capture_q09.py:63` calls `initialize(client, source.parent)` with the helper's `work_done_progress=False` default (`lsp_protocol_smoke.py:66`), so `ProgressState::begin` already returned `None` for the Q09 lanes — the remediation cannot have changed those measurements. `large-candidate.json` runs with diagnostics off (`diagnostics_seen: 0`), which returns from `publish_workspace` before progress selection. Q09/generations/large evidence keeps its original `dd282f9d8` provenance and claims no re-measurement on the final binary; the DX.10 F6 `+6720`-byte failure remains preserved and unrelabelled.

### Follow-up findings

- **suggestion** — `crates/sifr_lsp/src/server.rs:137`: the notification-error path still calls `publish_all`, so a *rejected* `didChange` (e.g. `reject_stale`) can open a `FullDiagnostics` progress scope with 2+ open documents. Not the typing hot path and not what the blocker covered (rejected mutations are rare and the full reconcile is defensible), but the same scope selection would be more consistent as `reconcile_changes` there.
- **infrastructure issue** — carried forward: per-edit O(open documents) push republication remains unbounded by latency evidence; `large-candidate.json` runs diagnostics off and the acceptance lanes use pull diagnostics with ≤2 documents. `push-progress-final` proves the progress behavior but is not a latency sample. Worth a scoped multi-document push-mode latency sample in DX.12–DX.14.
- **suggestion** — carried forward: `requests/folding_range.rs:11` / `selection_range.rs:19` return `-32603` for transiently invalid buffers via `parse_source_module`, where the prior host path degraded to an empty `Ok`. Terminal-response contract holds; degrading would avoid client error toasts while typing.
- **suggestion** — carried forward: `crates/sifr_driver/src/compiler_context.rs:49` `refreshed_toolchain()` rebuilds through `CompilerContext::new`, silently dropping `metadata_override` and an explicit `ResolvedSysroot`. Unreachable today; a latent hazard worth a debug assertion or explicit carry-over.
- **pre-existing issue** — carried forward: the "reduce demonstrated layout/clone/interner costs" bullet yields a post-close-only win (retained RSS −27.03 %, allocator chunks −98.83 %); steady +0.152 MiB / peak +0.127 MiB / switched-active +1.484 MiB are explicitly not claimed as improvements and are attributed. This is the DX.12–DX.14 baseline input, not an omission.
- **pre-existing issue** — the DX.10 F6 `+6720`-byte whole-file non-increase failure stays failed in its owning issue, untouched here.

PR #3862, opened 2026-09-18
Define DX.12 semantic dependencies and reusable result families

yaseralnajjar:
Project semantic persistence needs source and resolver facts that match the exact analyzed bytes, plus independently complete typed and diagnostic families. This change adds pinned source capture and ordered observations, package/configuration capture, conservative interface identities, and bounded typed HIR/interface transport through the existing metadata codec. Canonical diagnostics remap byte ranges into current source maps.

All 14 DX.12 tests pass on the exact candidate, covering I03, P03, P04, P07 and each planned result family. Coverage includes actual resolver precedence and cycles, source-set/package-root/lock edits, a real concurrent writer, source-table consistency, and UTF-8/16/32 diagnostic presentation. Format, file-size, maintainability and diff guards pass.

Cross-process reuse remains DX.13. Full create-PR/merge gates are deferred to phase end under the explicit Phase DX policy. Supplemental strict lint failures in unchanged code are recorded in their owning issues; the changed-line lint scan reports zero new diagnostics.


yaseralnajjar:
Scoped Opus review using [talk-to-claude-opus](https://github.com/sifr-lang/sifr/blob/cc2874a319feef6f96f45954b91dc3bed79cc2e8/.cursor/skills/talk-to-claude-opus/SKILL.md).

Base: f1d0e1d72ad107e816c16690cefddfb19df6200f  
Candidate: cc2874a319feef6f96f45954b91dc3bed79cc2e8  
Named acceptance/family tests: **14 passed, 0 failed**. Format, file-size, maintainability and diff guards passed. Full gates remain deferred to phase end under the approved phase policy.

Evidence: /home/yaser5/projects/sifr/dx12-evidence/cc2874a319feef6f96f45954b91dc3bed79cc2e8/qualification-manifest.json  
Validation manifest SHA-256: 14aa937944a74ab95f0cfc868aec1b99bb8db18da2cfba1bb642baf52510649a  
Review SHA-256: 91e9f228268edef441c8769b5f132f367d4d4cf026f99f0d95c429158c921cdf

The four nonblocking suggestions will be recorded as separate follow-up work in the phase record update.

---

Reviewed the full diff at the two exact SHAs, plus architecture §4/§9.1/§9.3–9.4 and the recorded evidence.

**What I checked**

- **Source capture**: `CapturingSourceProvider::read_file` (`crates/sifr_frontend/src/persistence/observations.rs:105-127`) memoizes per path, so the identity recorded and the bytes handed to parsing are the same read; `dx12_p07_*` in both the frontend and the driver (real `python3` writer against `DiskSourceProvider`) prove a concurrent writer cannot relabel the analyzed input.
- **Ordered resolution / absence**: the journal is a `Vec<Observation>` in actual call order, never sorted; only directory membership is sorted (`members`, line 282). `FileProbe{exists:false}` records absence, and `dx12_p03_*` proves a new higher-priority candidate and a newly-appearing previously-absent file both fail `unchanged()`. The driver test threads the capture through the real `ModuleResolver`/`parse_import_closure_source_modules` and asserts `compute_module_compile_order` still rejects the cycle — resolver precedence and cycle policy unchanged.
- **Package/external**: `capture_package_resolution` reuses `parse_metadata_json`, `derive_package_graph`, `PackageSourceMap::build` with the same provider; lock absence goes through a probe; `required_external` gates `SemanticInputs::complete()`.
- **Result families**: all five in §9.1 exist with serde round-trips, and `Family` models `Pending/Cancelled/Crashed/EnvironmentBlocked` as non-reusable. `check_complete()` holds under `SourceOutcome::DeterministicErrors` and never implies `checked`/`codegen` — the separation the DoD asks for.
- **Canonical diagnostics / remapping**: `CanonicalDiagnostic` strips every span including suggestion edits (`RenderedDiagnosticChild` carries no span, so coverage is complete), stores byte ranges validated against captured text char boundaries, and re-registers into a fresh `SourceMap` with pre-existing allocations. The LSP test renders one byte-identical fact under UTF-8/16/32 and asserts the persisted bytes are unchanged — presentation does not force recomputation.
- **Conservative interface**: `SemanticInterface.input_identity` folds the whole `ModuleInputs`, so a body-only edit invalidates; no unchanged-interface optimization is claimed (correctly left to DX.14).
- **Reuse not enabled**: `encode_project_results`/`decode_project_results`/`CapturingSourceProvider` have no non-test call sites; nothing writes a generation.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `CodegenHandoff.checked_module_identity` is validated against `CheckedModule.input_identity`, not `module_identity` (`crates/sifr_frontend/src/persistence/results.rs:549-554`, `crates/sifr_driver/src/metadata_reader/mod.rs:106-110`). Behavior is self-consistent and the conservative direction, but the name says "module" while the value is an input stamp; DX.13 will wire this and could easily pass the wrong one.
- suggestion: `CanonicalDiagnostic::render` registers every entry of the supplied `sources` map into the `SourceMap` even when no span references it (`crates/sifr_frontend/src/persistence/diagnostics.rs:99-108`), growing the map with unused sources when a caller passes a whole-project table.
- suggestion: `Observation`, `Family` and `SourceOutcome` lack the `#[serde(deny_unknown_fields)]` that every sibling struct in the module carries; tightening them before DX.13 reads foreign bytes keeps the decode contract uniform.
- suggestion: `SemanticInputs` gates external inputs with an explicit `required_external` set but has no equivalent for `components`, so an empty `components` map satisfies `complete()`. Fine while the outer owner populates it, but it is the one input class with no declared-completeness check.
- pre-existing issue: the supplemental strict lint findings (sysroot `integrity.rs`, `layout.rs`, `native_context.rs`, the frontend metadata overlay wildcard import, and the DX.10 profile items) are unchanged-owner warnings already recorded in `ad-hoc-dx-metadata-review-followups.md` and `ad-hoc-dx10-profile-review-followups.md`; not DX.12 scope, and the JSON scan confirms zero new/touched-line diagnostics.

PR #3864, opened 2026-09-18
Restore completed project checks across CLI processes (DX.13)

yaseralnajjar:
DX.13 restores completed source checking in a new CLI process. The frontend records canonical diagnostics and captured source/resolution observations; the driver validates those inputs against an immutable, versioned project generation before returning a restored result. Pure package checks retain their ordinary context resolution, and operations requiring live external work remain under their existing owners.

Generations inherit immutable records, publish transactionally, retain active-reader leases, and support pressure-driven project-only pruning. Deterministic source errors can be restored; interrupted or transient computations cannot publish completed success. Optional cache failures preserve language results, `.sifrbuildinfo` stays optional, and `--no-incremental` uses normal computation. Typed/codegen/editor families remain independently incomplete and compute normally when requested.

Validation: named C04/C08/C09/P01/P02/P08–P10/P12 acceptance and focused frontend, CLI, and storage regressions; format, file-size, HIR maintainability and diff guards. Independent CLI processes exercise dependency edits and fresh/restored diagnostics with counters, validation/serialization timing and peak RSS. Exact evidence is recorded in the review comment. The approved prospective policy defers the full gate to phase end. No DX.14 work or release qualification is included.


yaseralnajjar:
Exact candidate: `be67cc8dfa7d6db91f295557f08123d705cc79ff`
Base: `72e9490fc43a809648947d8cf6cc9594a30abd53`

All 12 qualification steps passed, including 12 DX.13, 3 DX.12 driver, 8 ownership, 9 frontend persistence tests; CLI schema/global regressions; 31 independent CLI process invocations; format/file-size/HIR/diff guards. Full gate deferred to phase end by the approved prospective policy.

Evidence: `/home/yaser5/projects/sifr/dx13-evidence/be67cc8dfa7d6db91f295557f08123d705cc79ff`
Qualification manifest SHA-256: `489d91b9810ad6cc0856fd68dacc6b3a3bdc6faf7afe1ec4a1409c8bc005a601`
Validation SHA-256: `f9a1f1d05762ef790ead5ab4c0cfe41af415a6ee8c26008fa09b12336aa1493c`
Review SHA-256: `3b769f729399e919a157be781789b375e0b6d4ff79da50dc5518ff16b1a0a78a`

I reviewed the diff `72e9490f..be67cc8d` against the DX.13 scope, definition of done, named acceptance matrix (C04, C08, C09, P01, P02, P08–P10, P12), architecture sections 7 and 9, and the evidence receipts. Working tree is clean and HEAD equals the candidate.

**What I verified**

- `project_cache::check` (mod.rs:113) — restore requires full record revalidation (`generation()` at storage.rs:152 re-derives every record digest, manifest identity, and `semantic_inputs.identity() == context`) plus `CompletedCheck::validate` replaying every source/dir/probe/canonical observation against live disk. Cancel, `!reusable`, `!capture.unchanged()`, and capture failure each map to a distinct non-publishing status.
- Transactional publication (storage.rs:190) — writer lock, inherit-by-hard-link with copy fallback that never truncates a finalized payload, staged dir + fsync + atomic rename, post-rename full revalidation before the pointer moves. Crash at both `before-rename`/`after-rename` yields a miss (tests.rs `dx13_c09_p01_p09_...`, real killed subprocesses).
- Leases/GC (storage.rs:304) — writer-then-generation order, readers hold shared leases, GC skips latest and any leased generation; `dx13_c09_concurrent_process_reader_and_gc` proves a live cross-process reader blocks removal and A/B/A reuse survives.
- Optional-cache behavior — unavailable root, read-only workspace, ENOSPC injection, and read-only store dir all preserve the exact fresh diagnostics (`c04-*` rows, `dx13_c04_*` tests). Metadata is resolved on the hit path too, so a restore can't skip stdlib pinning.
- Shallowness — `checked`/`interface`/`codegen` stay `Pending`; `p08-emit-after-check` output is byte-identical to `--no-incremental` emit, and the test asserts `hir_module_view(...).cache_status == Miss`.
- Package eligibility — `package_context::identity` rejects Python runtime, backend crates, rust backend, components, SQL/Python manifest config, and binds full graph/source-map projections; the empty generated `InteropBuildPlan` attestation (api.rs:170) is required in addition. `live-owner-context` confirms the fall-back to ordinary live checking.
- `--no-incremental` performs no reads or writes (manifest hashes unchanged across the disabled run) and `.sifrbuildinfo` removal still restores via the user-cache pointer. `members()` excludes only compiler-owned hint names from observation identity, leaving the real resolver's entries untouched.
- Guardrails: fmt, file-size, HIR maintainability, `git diff --check` all green in `validation.json`; no `.unwrap()`/`.expect()` introduced on data-dependent paths in shipped code.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- infrastructure issue: `validation.json` does not include the `sifr`-crate unit tests that directly exercise the rewritten entry points (`crates/sifr/src/diagnostics_and_packages_tests.rs`, which calls `check_entrypoint` ~20 times), nor clippy. The CLI acceptance runner covers the same paths through real processes, and the phase exception defers the merge gate, so this is not blocking — but those are the closest focused regressions to the changed behavior and are worth adding to the phase-end gate run.
- suggestion: `crates/sifr_driver/src/project_cache/storage.rs:240` — each publish creates a uniquely named stage lock inode under `generations/.locks/`, and lock inodes are never unlinked by design (`cache_storage.rs:1`). `prune` reclaims generation directories but not these, nor leaked `latest.stage-*` scratch files (storage.rs:288). Unbounded small-inode growth proportional to publish count.
- suggestion: `crates/sifr_driver/src/project_cache/mod.rs:220` — `prune_project_cache` is scoped to one named workspace; there is no reclamation for whole `projects/<workspace-id>/` namespaces whose workspace no longer exists. Every transient/deleted project (including each unit-test temp dir) leaves a permanent namespace in the user cache.
- suggestion: `crates/sifr_driver/src/project_cache/storage.rs:221` — manifests inherit every historical record, so a long-lived workspace accrues one record per distinct edit until `publish` permanently returns "project record capacity reached" at `RECORD_COUNT` 4096. Degrades safely to `write-unavailable` (source results stay correct), but caching then stops for that context with no eviction path.
- suggestion: `crates/sifr/src/main.rs:78` — `PROJECT_CACHE_OPTIONS` is set only in `run_cli`, so non-CLI in-process consumers silently default to `(enabled, quiet)` and publish into the real user cache. An explicit default for non-CLI entry would make that intentional rather than incidental.
- suggestion: `crates/sifr/src/cache_cli.rs:34` — with `--reserve-bytes` defaulting to 0, `available < reserve_bytes` is never true, so a bare `sifr cache prune-project <ws>` is always a no-op; the response key `eligible_generations` is also used for counts that were actually deleted.
- pre-existing issue: `crates/sifr/src/check_and_package_commands.rs` is now 898 lines against the 900-line guardrail. It passes, but the next edit to this file will need a responsibility split.

PR #3866, opened 2026-09-18
DX.14: preserve proven importer checks and restore saved editor diagnostics

yaseralnajjar:
DX.14 now recomputes edited modules and retains unchanged importer diagnostics when a conservative semantic-interface proof permits reuse. Saved CLI diagnostics can enter an existing editor generation only after the pinned compiler, metadata, context, sources and resolver observations agree; unsaved overlays never publish CLI authority. Deeper query and native work still run through their existing owners. Interface summaries retain full SHA-256 fingerprints so implementation-body size does not inflate resident signatures or graph cache keys.

The positive body-edit proof is intentionally limited to undecorated, nongeneric, zero-argument primitive-return functions with a single pure return expression in flat legacy source workspaces. Defaults, constants, generic/const bodies, effects, ownership, unknown scopes and package/external contexts remain conservative. Exact whole-project reuse keeps its existing eligibility and trust rules.

Validation covers P05/P06/P11, bounded randomized edits against fresh independent checking and explicit expected results, saved/stale/overlay editor paths, E01–E06 enabled/disabled, and Q09 paired memory/retention traces. Native P06 emits 1 then 2 after a helper edit while retaining the importer check. Focused DX.13/frontend/DX.11 regressions and format/size/maintainability guards pass. Evidence and scoped Opus review are published separately, keyed to the candidate SHA.

The approved phase-end policy defers the single full gate to the final DX.15 implementation. No full create-PR/merge gate or release qualification is claimed for this intermediate milestone.


yaseralnajjar:
## DX.14 final candidate qualification

Candidate: 2c3f23127966850e7ed961155bcee817514d56f1
Base: b54f8985c50e4f0da4ab9bcc9886da8dab8dec53
Verdict: **SATISFIED after one bounded remediation**.

P05/P06/P11, 35 targeted Rust tests, 44 independent CLI invocations, native P06 output refresh, E01–E06 with persistence enabled/disabled, and all 84 Q09 normal/heap samples passed. All 42 enabled memory samples actually restored one saved module; all 42 disabled samples restored none. The fixed-width interface-key regression covers a 2000-statement body. All final-candidate tests and the optimized installed product were freshly rebuilt/run after the correction.

No full create-PR/merge gate ran: the canonical prospective phase-end policy reserves that gate for the final phase implementation. No release was requested. Q09 retains the historical equal LSP PACKAGE0103 status; no performance or active-memory win is claimed.

The initial review blocked raw AST Debug strings retained in signatures and repeated graph keys. That regression was fixed with the full SHA-256 structural fingerprint; the earlier review and failed development/setup records remain preserved, not relabeled. The initial and final reviews' nonblocking findings will be recorded separately.

External evidence root: /home/yaser5/projects/sifr/dx14-evidence/2c3f23127966850e7ed961155bcee817514d56f1
Qualification manifest SHA-256: 4a3732ea151d5decd62ade500c1a6bb20ad7a7f88bc170231c6976ad35fe2c4b

<details><summary>Qualification manifest</summary>

```json
{
  "item": "DX.14",
  "base": "b54f8985c50e4f0da4ab9bcc9886da8dab8dec53",
  "candidate": "2c3f23127966850e7ed961155bcee817514d56f1",
  "implementation_pr": "https://github.com/sifr-lang/sifr/pull/3866",
  "workspace": "/home/yaser5/projects/sifr/compiler-dx-orchestration",
  "host": "yaser.tailaa73b4.ts.net",
  "rustc": "1.98.1",
  "evidence_root": "/home/yaser5/projects/sifr/dx14-evidence/2c3f23127966850e7ed961155bcee817514d56f1",
  "artifacts_sha256": {
    "validation.json": "e94965a3ec3c804117533b59fa4fa6a45ac53e75a3ae3336b1a4e16ce92ea78a",
    "product/receipt.json": "dbb82ef3fb736a8d33567110ef4427818eab2cdfb67c70b63c29ccb485cb53c3",
    "cli/report.json": "50a2243844f562002f9fe174e542f4f98d5d89bfd1c448c389e24ea8b378b484",
    "editor-enabled/report.json": "33f6ff6216d9f869fb0653d5aae2a0222cca1f5b6e9c297b250fbc90926f62ab",
    "editor-disabled/report.json": "24b2d2f1070779ee49dc3adb7c53c5673eeb684139daa8ef6fc34315eb0edd5e",
    "native-p06/report.json": "41835b7724bf1fb37e30e262d084c94a437b0b61d441508a833b768bd398ac0c",
    "q09-comparison.json": "c2eb67738ea1e5d53b0ddd57caf8be5dbf120d44e374094217050d638cae7067",
    "opus-review.md": "7195c7a65553aa175e588d6643dad94820d6de1c59ecbf3b086b449db211c5c8"
  },
  "validation": "Named P05/P06/P11; 24 deterministic unit edits and 16 independent CLI paired edits with semantic expectations; E01\u2013E06 enabled/disabled; Q09 84 normal/heap samples (four cohorts, first sample predetermined warmup). Native outputs 1 then 2 despite restored importer checks. Focused DX.13, frontend-cache and DX.11 regressions; format/size/maintainability/diff guards.",
  "evidence_reuse": "Final remediation candidate rebuilt the optimized installed artifact and reran named acceptance/focused checks. Earlier compiler-unchanged harness candidate evidence remains preserved, including original receipts.",
  "gate_policy": "Intermediate phase item: no create-pr/full merge gate. One full gate is due on final DX.15 implementation; no release requested.",
  "review": "SATISFIED after one bounded remediation of the initial AST-string residency/cost regression. Both completed reviews remain outside the reviewed Git tree, keyed to their candidate.",
  "historical_evidence": "Earlier interrupted/development/setup failures retained outside tree; not relabeled as passes.",
  "initial_review": {
    "candidate": "86cef8b58945d7f26366e25328b4beb8e49b5754",
    "verdict": "NOT SATISFIED",
    "path": "/home/yaser5/projects/sifr/dx14-evidence/86cef8b58945d7f26366e25328b4beb8e49b5754/opus-review-initial.md",
    "sha256": "522744a96f5a73f439543f65e3af4bb4f413b74feee63302475b25d829bfbc2e",
    "resolution": "Sole blocking body-AST string residency and repeated cache-key cost fixed in 2c3f23127; final scoped remediation review SATISFIED."
  }
}
```

</details>

## Final scoped Opus 5 remediation review

## Verdict: SATISFIED

I reviewed the remediation delta (`2c3f23127`) against the prior blocking finding, plus the owner code it touches, the DX.14 invalidation/restore paths, and the acceptance evidence.

**The sole prior blocking finding is genuinely fixed, not papered over.**

- `crates/sifr_frontend/src/module_signatures.rs:259` — `interface_projection` no longer materializes `format!("{:?}", ComparableStmt)` per statement. It streams the derived structural `Hash` into the existing `IdentityEncoder` (SHA-256) and returns only the 64-hex digest. `ModuleSignature.semantic_body` and therefore `cache_key_input()` (consumed at `graph_cache_and_queries/reuse.rs:412` in `semantic_graph_fingerprint`) are now fixed-size per module regardless of body size. No body-sized string reaches resident state or repeated graph hashing.

**No discrimination was lost by switching Debug → Hash** (the failure mode that would have made this a permissiveness regression):
- Every type in `third_party/ruff/.../comparable.rs` derives `Debug, PartialEq, Eq, Hash` together (108 derives, no manual `impl Hash`), so `Hash` and `Debug` are driven by identical fields.
- Source ranges are already discarded at `Comparable*` construction (`range: _` throughout), so the old projection was not range-sensitive either — the new one is not more permissive.
- `IdentityEncoder::field` length-prefixes both name and payload, so the byte stream is at least as unambiguous as the underlying Rust hash stream; the two statement classes are now explicitly tagged `0_u8`/`1_u8`, removing a namespace collision the old string form had.
- `ModuleSignature::default()` (parse failure) yields `semantic_body: ""`, which can no longer alias an empty-but-parsed module (which now yields a digest) — strictly more conservative than base.

**Conservative rules unchanged.** The erasure predicate is byte-identical: no type params, no decorators, not async, zero params, explicit primitive return, single pure return. Critically, `decorator_list.is_empty()` keeps `@const_eval` out of the erasure class, and `const_evaluator.rs` only evaluates explicitly-const functions (`FunctionNotConst`), so the erased return expression cannot feed cross-module constant evaluation. `dx14_tests.rs:144` pins that negative case directly.

**Stability boundary holds.** The digest is never persisted: `persistent_checks.rs:61` recomputes both sides in-process, and `cache_key_input()` only feeds `semantic_graph_fingerprint`, whose `CacheKeyContext` is namespaced by `CompilerFingerprint::for_identity` (`reuse.rs:373`). Native-endian `Hasher` defaults and rustc-version `Hash` drift therefore cannot cause a cross-build false restore.

**Acceptance still met.** Restore requires empty saved diagnostics, identity/context match, revalidated observations including absence, successful checking of every changed module, and `allow_body_edits: false` on the editor path (`project_cache/mod.rs` `restore_editor_checks`) so no unsaved overlay becomes CLI authority. Reuse effectiveness is separated from correctness by the `reused >= 10` floor in `dx14_bounded_edits_match_fresh_with_independent_error_expectations`. The new focused regression is real (`interface-key-size.log`: 1 passed) and the DX.14/DX.13/DX.11/frontend-cache suites and both editor cohorts pass on this candidate.

### Blocking findings

None.

### Follow-up findings

- **suggestion** — `module_signatures.rs:306` `InterfaceHasher::finish` clones and finalizes the full SHA-256 state on every call and returns a 64-bit truncation. It is unreachable today (all `Comparable*` `Hash` impls are derived and none calls `state.finish()`), but a future manual `Hash` impl that nests a sub-hash would silently make it O(n) per call and substitute a truncated value into the stream. Worth a `debug_assert!(false, ...)` or an explicit invariant test rather than a comment alone.
- **suggestion** — `module_signatures.rs:302` `write` tags every chunk with the 21-byte literal `"structural-hash-chunk"`, adding ~37 bytes of SHA-256 input per primitive write. Constant-factor only, and still far cheaper than the removed Debug formatting; an empty field name would halve it.
- **suggestion** — the doc comment at `module_signatures.rs:299` ("The compiler identity pins Rust/AST hash encoding for disk records") overstates the coupling: the digest never reaches a disk record; only compiler-identity-namespaced in-process cache keys use it.
- **pre-existing issue** — `ExportSignatureEntry::shape` (`module_signatures.rs:196`, `:206`) still renders `format!("{:?}", comparable_expr(&assign.value))` for module-level constants, so a large literal initializer still grows the resident signature and cache-key string. Same residency class as the fixed finding, but present at base `b54f898` and declaration-level rather than body-level, so outside this remediation batch.
- **suggestion** — chained interface restores compound: the record published on the `interface-restored` path asserts project-wide empty diagnostics, and a later hop proves interface stability only against that synthesized record. Each hop is individually sound under the current erasure class, but the transitivity assumption is not stated in `compiler_dx_architecture.md` §9.3 and would be worth recording before the erasure class is widened.

PR #3868, opened 2026-09-20
Complete DX.15 compiler performance and package qualification

yaseralnajjar:
Installed checks paid for unrelated virtual-workspace Cargo validation and eager metadata work. This completes DX.15 with proven standalone-source ownership, bounded lazy metadata format 4, byte-identical SHA256 acceleration, and explicit cache/metadata timing while preserving package, lock-mode, integrity and semantic validation.

The final integration repairs retain required assertions and deadlines, regenerate the affected pinned SQL components, and complete nonpublishing native package qualification on all four supported targets. DX.16 remains open.

Validation:
- Required merge-profile coverage is accounted across 32 compiler configurations, 20 areas and guardrails, using the original gate plus affected/never-run fail-fast continuations. Original failures remain preserved; no duplicate broad gate or create-PR gate.
- Fixed 100+1 cohorts: fresh CLI p95 46.461 ms, unchanged 45.838 ms, editor 2.415 ms; separate tracing, baseline, noise and preparation evidence.
- All four local sysroot suites pass, including all 727 source/provider and public CLI emission cases, native fixtures, live generations and invalid-artifact recovery.
- [Four-target native qualification](https://github.com/sifr-lang/sifr/actions/runs/35477213727) covers exact package bytes, install/update/downgrade/reinstall/rollback/relocation and native profiles.
- Actual Mac desktop and 12 GB Linux headless resource results retain their distinct hardware scopes. The literal combined 12 GB graphical-desktop scenario is not claimed.

Final head: 9ee360474655fde979dc5ddfdf33eb4dc39c9968. Qualified e2ab artifacts and the versioned 5127 local package keep their original receipts; subsequent changes are two acceptance-parser lines and documentation only, with explicit unchanged-input proofs.

External qualification index: /home/yaser5/projects/sifr/dx15-evidence/9ee360474655fde979dc5ddfdf33eb4dc39c9968/qualification-index.json
SHA256: 007655f307c57df622a7cca39aa298c64f3d826e17f59673ae0674fd048847cd

Scoped read-only Claude Opus 5 review: **SATISFIED**, no blocking findings. Response SHA256: 8bd196cef0661f2d0e51f59878a360eef70466f75e8e0cd37b4894a979f103a6. Nonblocking verification suggestions are recorded separately; no required implementation cleanup is deferred to DX.16.


yaseralnajjar:
Reviewed the base→candidate diff (`f14755ab` → `9ee36047`, 137 files), the DX.15 milestone scope, architecture §6.3/13/14/16, and the external evidence index/context. Spot-verified independently: SQL component digests, artifact-input-equivalence claim, test-selector renames, LSP split-brain checks, native-package contract tests, native-case scheduling tests, and the file-size guardrail.

**Verdict: SATISFIED**

Key verifications:
- **Format 4 decoder** (`crates/sifr_sysroot/src/metadata/physical.rs`, `physical_input.rs`, `store.rs:53-165`): all frame content lengths, physical/expanded lengths, directory bounds and per-record lengths are validated before allocation; the outer clear header is cross-checked against the decompressed inner header; record IDs are still required strictly increasing (`store.rs:101`), which is what makes the new `Directory` binary search sound; per-record SHA256 stays authoritative on read. No unchecked trusted path, no compatibility fallback. Slice/index paths are all length-guarded — no new panic surface.
- **Standalone-file boundary** (`crates/sifr/src/check_and_package_commands.rs:38-47`, `crates/sifr_package/src/ops/workspace_session.rs:13-93`): gated to `CargoLockMode::Normal`, requires manifest keys limited to `package`/`source`, a successful `SifrManifest::load` (so edition and `sifr-version` validation still rejects), a virtual `Cargo.toml` with no `[package]`, source-root containment, and no intervening Cargo/foreign `sifr.toml`. `plan_check` already ignores `selection` for explicit files and `validate_explicit_file` only checks source-root containment, so the bypass drops no existing check. `--locked/--offline/--frozen` and real packages are covered by `crates/sifr/tests/standalone_workspace_check.rs`.
- **`prepare_native=false`**: only reaches `SingleFile`/`Project`, and `from_entrypoint_with_stages` (`entrypoint.rs:450-555`) never consumes `cargo_resolution` on those arms. Every native/backend entrypoint still passes `true`.
- **Ignored-case scheduling** (`native_test_execution.py`): `--list --ignored` yields the complete ignored inventory, count/duplicate/truncation/encoding are validated before execution, each child re-asserts `1 passed`, per-child deadline stays at the executor default 2400s, fail-fast is default. `ProfileRunnerError` subclasses `VerificationError`, so widening the `timed_step` catch loses nothing.
- **Evidence honesty**: `e2ab8e1f9 → 5127a24ee` is exactly the two acceptance-parser lines, and `5127a24ee → 9ee36047` is docs-only — the equivalence claim checks out against the actual objects. The plan record and `compiler_dx_architecture.md:728` explicitly keep the literal 12 GB graphical-desktop scenario as not executed and do not raise the Linux floor. `remaining` contains only this review and the merge.

**Blocking findings:** none.

**Follow-up findings:**
- *suggestion* — `verification/areas/developer_tooling/dx13_project_acceptance.py:100,117,132,137` and `dx14_interface_acceptance.py:168,177`: canonical §16 case IDs (P01/P02/P05/P06/P08/P10) were replaced with descriptive keys in `report["cases"]`. Nothing consumes them today, so this is not a regression, but DX.16's reconciliation will need the ID→label mapping recorded somewhere.
- *infrastructure issue* — `verification/runner/sifr_verify/profile_reporting.py:51`: the Linux KiB→bytes correction is right, but `reports.py:155` compares `max_rss_bytes` against `DEFAULT_LANE_RSS_ADVISORY_BYTES`, so create-pr/merge lanes will start emitting the RSS advisory note that was previously suppressed by the 1024× undercount. Expect it; don't read it as a new regression.
- *infrastructure issue* — `verification/runner/sifr_verify/audit_fixtures.py:67-68`: tolerating a regular `.sifrbuildinfo` under audit fixture roots works around the CLI writing the project-cache hint into fixture directories. Symlink rejection was correctly added, but redirecting the hint out of fixture roots would be the cleaner boundary.
- *suggestion* — `crates/sifr_driver/src/tests/package_rust_interop_test_support.rs:65-99`: `reusable_scenario` deletes and recopies the tree on every acquisition, so the preparation split only pays off via the deterministic generated-project/Cargo cache key. If that key ever picks up a nondeterministic input, `prepare_advanced_data_native_graph` becomes pure added cost and fails silently. An explicit cache-hit assertion in the assertion-phase test would make that regression visible.

PR #3877, opened 2026-09-20
Bound project cache history and lookup work (DXF.2)

yaseralnajjar:
Project-cache publication previously stopped after 4096 historical records. Retain the newest 128 distinct publications within 64 MiB instead, preserving immutable generations, active reader leases and recent A/B/A reuse. Evicted history safely misses.

Exact lookup prefers recent records; cheap entrypoint/context/success/resolver filtering precedes at most one interface-proof attempt. Capture retains each distinct observation in first-occurrence order, including absence and conflicting outcomes, and declines persistence after 16384 distinct facts.

Validation: all 36 selected Rust tests pass (four exact history tests; project-cache 12, interface 8, package 3, frontend persistence 9). Canonical storage stress publishes 4097 records; retained counts at 1/128/1024/4097 inputs are 1/128/128/128. The named CLI harness passes five real edits against independent fresh checks. Format, file-size, maintainability, Python syntax and documentation checks pass. The initial harness fixture failure is preserved; its explicit legacy source-root setup was corrected and rerun.

Candidate: 94d35655a065859acdecf1aa3ee33c31a3165305. All Rust validation inputs and compiler bytes are unchanged from the initial implementation candidate; the external evidence records that reuse. Evidence: /home/yaser5/projects/sifr/dxf-evidence/94d35655a065859acdecf1aa3ee33c31a3165305/.

DXF.2 only. Physical orphan/staging cleanup remains DXF.3; existing explicit pressure pruning still owns predecessor generations. These are functional contributor/test-profile observations, with no product speedup or constant-total-I/O claim. Scoped Opus review is SATISFIED with no blockers: https://github.com/sifr-lang/sifr/pull/3877#issuecomment-5749545936. Nonblocking follow-ups are recorded separately. No per-item monolithic gates under the approved prospective policy.


yaseralnajjar:
## Verdict: SATISFIED

I read the exact diff (`614e7ea2..94d35655`), the DXF.2 scope in `plans/issues/active/ad-hoc-compiler-dx-followup-execution.md:194-206`, and the evidence under both candidate directories.

**What the change actually does, checked against scope**

- `storage.rs:18-19,27,167-188,224-250` replaces the `4096` capacity `Err` with schema-2 ordered `Vec` retention: newest-first suffix bounded by `RECORD_COUNT = 128` and `HISTORY_BYTES = 64 MiB`, republication refreshing position. `load` enforces the same two bounds plus a duplicate-reference check, so publish and load agree; retention is deterministic (metadata-size walk oldest-ward, `break` not `continue`). Schema 1 generations fail `generation()`, which `latest()` swallows into `None` — a safe miss, consistent with the no-backward-compat rule.
- Immutability/leases are intact: the inherit loop still hard-links (never truncates) and skips only the new record; stage/rename/point and the prune writer-lock ordering are untouched. `eviction_preserves_active_reader_and_latest` confirms a leased predecessor survives prune, reads its evicted record, and is reclaimed only after drop.
- `interface_reuse::candidate` is strictly a pre-filter — every condition it tests (`source.path`, `semantic_inputs`, `Success`, `sources.len() >= 2`) is re-tested inside `restore`, and `.sifr` file identities are exempted, so the proof class and `restore_completed_checks` authority are unchanged. No false-hit path is introduced.
- Observation dedup (`observations.rs:73-92`) keys on the whole `Observation`, so differing outcomes for the same path (absence then presence) are both retained in first-occurrence order; overflow at 16384 sets `overflow`, which forces `unchanged()` false and makes `CompletedCheck::capture` refuse publication (`check.rs:26-29`) while normal checking proceeds. Resolver/absence authority is preserved.
- Documentation (`compiler_dx_architecture.md:630-647`) states the cap honestly, including "not a cap on all physical generations" and "bounded proof work does not imply constant total storage I/O", and explicitly leaves predecessor reclamation to prune (DXF.3).

**Evidence**

Reuse justification is sound: only `dxf_history_acceptance.py` changed from `e2a296c9`, and the fix was a fixture `sifr.toml` source root, not a compiler mechanism. All four named tests selected exactly once and passed; `project_cache::tests` 12, `dx14_tests` 8, `package_reuse_tests` 3, `persistence::tests` 9 pass. `history-stress.json` shows retained 1/128/128/128 and lookup 0.42/21.7/21.0/22.1 ms — flat past the cap, with candidate 1, proofs 0, observations 1. The CLI harness shows the real A/B/A: `published → interface-restored → restored → published (exit 1, equal to fresh) → interface-restored`, each paired against an independent `--no-incremental` check.

## Blocking findings

None.

## Follow-up findings

- **suggestion** — `crates/sifr_driver/src/project_cache/storage.rs:326`: the `prune` doc comment still says "Current generation inherits all retained records, so inactive predecessors can go without losing unused results." With eviction, predecessors now hold records the latest generation does not; the loss is intentional and documented in `compiler_dx_architecture.md`, but this comment now asserts the opposite.
- **suggestion** — `crates/sifr_driver/src/project_cache/interface_reuse.rs:24-26`: `candidate` replays non-`.sifr` observations through the shared `CapturingSourceProvider`, so probes/directory reads from a *rejected* candidate are appended to the capture and end up in the record subsequently published. Only adds true facts (conservative, never unsound), but it inflates `observation_count` and the validation work of future lookups.
- **suggestion** — `crates/sifr_driver/src/project_cache/mod.rs:172`: records are now eagerly collected into a `Vec` before the cancel check, deserializing the whole retained generation (up to 64 MiB of JSON) per lookup; the previous lazy iterator observed cancellation mid-scan. Bounded by the new cap, so not a class regression, but it is the dominant cost of the measured ~21 ms lookup.
- **suggestion** — `crates/sifr_driver/src/project_cache/mod.rs:213`: `interface_proofs == 1` caps the attempt at the newest qualifying candidate; if that one fails the proof, older records that would have proven are never tried. Intentional and required by the "must not grow linearly with irrelevant history" criterion, but it is a reuse-hit-rate narrowing worth recording.
- **infrastructure issue** — `crates/sifr_driver/src/project_cache/history_tests.rs:19`: `history_over_4096_keeps_publishing_and_recent_aba` takes ~95 s and is not `#[ignore]`d, so it now dominates the ordinary `cargo test -p sifr_driver --lib` wall time (the other three named tests total under 3 s).

PR #3879, opened 2026-09-20
Reclaim abandoned and orphan project storage safely

yaseralnajjar:
Killed project writers left staging locks and pointer scratch, and deleted workspaces could not be selected for explicit pruning. This adds immutable workspace ownership, namespace leases shared by stores and detached readers, writer-serialized abandoned storage reclamation, and explicit orphan cleanup by the original absolute canonical workspace path. Active/ambiguous/replaced/symlink owners stay protected; namespace tombstones and live generation lock inodes remain stable.

The prune API/CLI distinguish examined, eligible and deleted entries. Zero reserve remains a documented no-op, and cleanup runs only under explicit pressure. No startup cleanup or DXF.4 changes.

Validation: cheap formatting/file-size/HIR/diff guards; three new exact housekeeping tests; the three required DX.13 process/reader regressions; bounded-history reader/latest regression; CLI schema compilation and exact contract test. Exact-list guards require one selected test. No per-item full gate under the approved follow-up policy. Scoped Opus review will be attached before merge.

Scope: DXF.3 in `plans/issues/active/ad-hoc-compiler-dx-followup-execution.md`. Candidate `0bc95340dfe6a5e4e32e0824a7b62455dae24df7`; external evidence `/home/yaser5/projects/sifr/dxf-evidence/0bc95340dfe6a5e4e32e0824a7b62455dae24df7`. Functional local-filesystem evidence only; no release or performance claim.


yaseralnajjar:
Scoped remote Claude Opus 5 review of candidate `0bc95340dfe6a5e4e32e0824a7b62455dae24df7` against `9af5e3b8d5b0d17c37c855f144d3747a937f9466`. Eight exact named tests and cheap guards passed on the same implementation. External review SHA-256: `c0e5c3613cb8de2667ec886640c37d406d684ee72de83e00df93c242716ab2db`. Nonblocking follow-ups will be recorded separately after merge; no new scope is absorbed.

Verdict: SATISFIED

Blocking findings:
- none

Basis (read-only inspection of the diff and surrounding source, not the summary):

- Ownership proof: `housekeeping.rs:52-113` writes an immutable `owner.json` (schema, canonical path, `dev`, `ino`, btime) with `write_new` (`O_CREAT|O_EXCL`, `AlreadyExists` tolerated), then re-reads and compares every field; mismatch is a hard error. `open_namespace` refuses a nonempty namespace lacking `owner.json` (`housekeeping.rs:74-80`), so no retroactive claim.
- Exclusion: the namespace lock lives outside the namespace (`projects/.locks/<id>`); stores and detached readers hold it shared (`Store.namespace_lease`, `Generation._namespace_lease`, `storage.rs:43,49,213`), orphan GC takes it exclusively (`housekeeping.rs:352-357`) and additionally exclusively locks every existing `.locks` entry in the context tree via `lock_tree` before deletion, re-checking `definitely_absent` immediately before each `reclaim`.
- Live-lock preservation: `Store::generation` switched from `process_entry_lock` to `existing_lease` (`storage.rs:167`) and publish creates the permanent generation lock before the stage rename (`storage.rs:266`), so read paths never create lock inodes; `prune_store` deletes only `stage()`-named lock files under writer serialization (`housekeeping.rs:239-269`), leaving generation and writer inodes stable — asserted by inode equality in `abandoned_stages_reclaimed_live_locks_preserved`.
- Ambiguity preservation: unvalidatable `latest` pointer skips generations (`housekeeping.rs:223-227`); `safe_path`/`safe_tree`/`check_owned` reject symlinks, foreign or group/other-writable entries and non-directory ancestors; `definitely_absent` treats dangling symlinks and replacement objects as present; workspace hint scratch requires a full schema/workspace/context match. Each of these has a matching assertion in `orphan_prune_requires_owner_and_inactivity` (replacement dir, dangling symlink, nested alias to an outside tree, corrupted `owner.json`, live writer lock, unknown ownerless namespace).
- Orphan policy accessibility without recreation: `prune_workspace` accepts the original absolute path, rejects a relative path for an absent workspace with an explicit error, and never calls `Store::open` (comment and manual `Store` construction at `housekeeping.rs:409-424`), so no directory is re-established during a dry run.
- Counters: `examined_entries` counts recognized candidates including protected ones; `eligible_*` only increments inside `reclaim` after `safe_tree`; `deleted_*` only after the unlink succeeds; dry-run evaluates the stage-lock branch equivalently (`housekeeping.rs:249-252`). I traced the exact expected tuples in `pressure_prune_counts_are_truthful` (4/2/0 dry, 4/2/2 actual, 2/0/0 repeat) against the code paths and they match, including `eligible_generations == 0` for non-generation candidates.
- Zero reserve: `available < reserve_bytes` is false at the default, and `prune_workspace`/`prune_store` return the zeroed report before any inspection or directory creation (asserted by `!cache.exists()` and `!empty.root.join(".locks").exists()`).
- Scope discipline: no startup cleanup, no compatibility path, no global scan, no DXF.4 work; `cache_cli.rs` only renames the payload and clarifies help, with the CLI contract left to DXF.6. Evidence index matches the changed-path digests, records one selected test per name, and makes no performance or gate claim.

Follow-up findings:
- suggestion: `housekeeping.rs:127` — `entries.into_iter()` on a `ReadDir` (already an `Iterator`) will likely trip `clippy::useless_conversion` under the workspace `-D warnings` merge gate. Clippy is not among this item's mandated cheap guards, so it was not run; worth resolving before the phase-end gate.
- suggestion: `storage.rs:266` — publish creates the permanent generation lock before staging, so a publication that fails after that point (e.g. the `ENOSPC` path) leaves a zero-byte lock file with no generation. `prune_store` reclaims only `stage()`-named locks, so these accumulate permanently. Tiny, but it is new leakage that did not exist when locks were created lazily.
- pre-existing issue / documented limitation: `housekeeping.rs:96-105` — a workspace directory deleted and recreated at the same path (re-clone, container rebuild) fails `open_namespace` with "project namespace owner changed" forever, and `prune_workspace` also refuses it (non-orphan identity mismatch, `definitely_absent` false), so the old namespace is neither reusable nor reclaimable. This is deliberate, matches the "replacement paths must preserve" criterion, and is documented in `compiler_dx_architecture.md`, so it does not block — but the unreclaimable residue is the kind of storage DXF.3 exists to reclaim and deserves an owner.
- pre-existing issue / infrastructure: `housekeeping.rs:44-50` — `created_ns` hard-fails where `statx` btime is unavailable (some overlayfs/NFS/older-kernel setups), silently disabling project persistence entirely. Honestly documented in `compiler_dx_architecture.md` §7.3; no host evidence exists for such a filesystem, so it remains an untested limitation rather than a verified defect.
- suggestion: under real contention the CLI reports all-zero counters with `"pressure": true` (WouldBlock swallowed per context at `housekeeping.rs:421-423`), which is truthful by the stated counter definition but indistinguishable from "nothing to do". A contention indicator would fit naturally in DXF.6's CLI contract work.

PR #3883, opened 2026-09-20
fix(trace): preserve bounded finalization and private directories

yaseralnajjar:
A trace that fits the 32 KiB cap before final timing serialization could fail solely because those numbers gained digits. Reapply bounded optional-report dropping after freezing final timing fields, preserving the versioned outcome and exact dropped count; real sink I/O errors remain errors. New Unix trace directories are created with mode 0700 while files remain 0600 and existing destinations are untouched.

Validation: all four exact DXF.5 Rust selections passed with one-test list guards, including below/at/above-cap timing/counter boundaries, all events dropped, genuine write failure and a child with permissive umask. All five installed optimized trace acceptance contracts passed against the candidate receipt. Formatting, file-size, HIR, Python syntax and diff guards passed. No per-item full gate under the approved follow-up policy.

Candidate: `e2a12e316beeaa950249b18e456b614fc77cd4df`. Evidence: `/home/yaser5/projects/sifr/dxf-evidence/e2a12e316beeaa950249b18e456b614fc77cd4df/evidence-index.json`. Scoped [Opus review](https://github.com/sifr-lang/sifr/pull/3883#issuecomment-5749892882): SATISFIED, no blockers.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Review basis: `crates/sifr/src/trace_artifacts.rs` (485 lines, under the guardrail) and the acceptance script at the exact candidate; both files' sha256 match the evidence index source entries, which binds the supplied logs/receipt/artifact to the bytes reviewed.

Criterion checks:
- Final timing/counter serialization before the cap: `finish_sink` (`trace_artifacts.rs:206`) runs `bounded_bytes` once, then `final_bytes` writes the widened `trace_overhead_us_before_final_write` / `invocation_us_before_final_write` and re-bounds, so the cap is enforced on the exact bytes written. The old `Err("trace output exceeds size bound")` path after timing rewrite is gone — this was the defect.
- Truthful drop accounting: `bounded_bytes` (`trace_artifacts.rs:245`) carries the counter in the JSON value itself and increments per pop, so drops from both passes accumulate; per-iteration reserialization absorbs counter digit growth. Required header fields (schema_version, identity, outcome, exit_code, timings) survive an empty `reports` array; an oversized header still errors.
- Real I/O errors: `rewind`/`write_all`/`set_len`/`flush` errors propagate unchanged and `exit` keeps original non-zero codes, mapping only success→2.
- Unix privacy: `create_trace_file` (`trace_artifacts.rs:41`) uses `DirBuilder::mode(0o700)` non-recursively, so the mode applies only to the directory it creates; no `set_permissions` on any existing path. Existing dir and symlink-to-dir destinations still fail with the collision error and are left unmodified; file mode stays 0600. `#[cfg(unix)]`-gated, and the test is `#[cfg(unix)]`.
- Acceptance addition is `os.name == "posix"`-gated and only runs when the CLI itself created the sink, so invalid-destination and preexisting-directory contracts are unaffected.

Follow-up findings:
- suggestion: `trace_directory_is_private_under_permissive_umask` asserts on libtest stdout containing `"1 passed"` (`trace_artifacts.rs:335`). It is correct under `cargo test`, but couples the assertion to harness output formatting; the child's exit status alone already carries the result, with the string check as a redundant guard against a filter that matched nothing.
- suggestion: `trace_overhead_us_before_final_write` excludes the second (final) bounding serialization. This is deliberate and documented in the comment, and consistent with `final_write_included: false`, so the field remains an honest lower bound — noting it only so the semantics stay explicit if the field is ever renamed.
- pre-existing issue: `dropped_reports` overflow yields `"invalid trace dropped report count"`. Unreachable in production given `MAX_REPORTS = 64`; it is only observable with the pathological `usize::MAX - 1` seed the new boundary test uses to force widest counter width.

PR #3885, opened 2026-09-20
test(dx): lock cache CLI and moved-workspace contracts

yaseralnajjar:
Cache commands lacked explicit parser/process contracts and workspace relocation lacked a focused regression. Add eager/deferred help, malformed argument, global placement and defaults checks; isolate process tests for JSON reserve/no-op/dry-run/deletion counts and orphan cleanup; prove moved source roots compute with fresh diagnostics. Repair the five audited documentation links using archived owners and an honest historical parity record.

DXF.6 candidate: `a7e45101600f89179fa1c5b2a80cb5beaa3e7d53`.

Validation passed: three exact nonzero Rust selections (all eager schemas, cache contract, moved-workspace diagnostics), 34 receipt-bound isolated CLI calls, formatting/source/HIR guards, documentation structure and 153 local link/anchor checks, diff and Python syntax. Rust/Markdown checks reuse unchanged inputs from `0297d3f14`; affected Python checks and process acceptance ran on the final candidate. Historical harness expectation failures remain recorded externally. Functional contributor lane; no performance claim. The approved follow-up policy requires named checks rather than per-item broad gates.

Scoped Claude Opus 5 review: SATISFIED, no blockers. External evidence index: `/home/yaser5/projects/sifr/dxf-evidence/a7e45101600f89179fa1c5b2a80cb5beaa3e7d53/evidence-index.json`, SHA-256 `83b4ecde6bbc94b211c98163982c93379769b9b845fc18372041503a9acea7d6`.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `plans/reviews/active/*` still cite `plans/issues/active/python-interop-verification-production.md`, `.../ad-hoc-stdlib-compiler-boundary-rearchitecture.md` and the archive file's own line 167 self-reference. These are historical review prose outside the five audited links and outside DXF.6's owner, so they were correctly left alone; they will resurface if a future audit widens the link sweep to `plans/reviews/**`.
- infrastructure issue: the local-link/anchor check that backs the documentation criterion is `dxf-evidence/check_dxf6_docs.py`, outside the Git tree. `verification/areas/documentation/check_structure.py` alone does not enforce link targets or anchors, so the repaired links have no in-tree gate protecting them from regressing. Worth folding into DXF.7's reconciliation or the documentation area rather than DXF.6.
- suggestion: `dxf_cache_cli_acceptance.py` hardcodes `examined_entries=4 / eligible_entries=3 / eligible_generations=1` from two publications plus two staging locks. That is the exact housekeeping shape corrected in `a7e45101`, but it couples the harness to internal staging-lock accounting; a future lock-layout change will fail here with a bare counter mismatch. A short comment naming the lock-count source (already partially present) or deriving the expected count from the observed `cache/projects/**` layout would keep the failure diagnosable.

Basis for the verdict:
- **CLI contract**: `cache` is added to the eager schema sweep (`crates/sifr/src/eager_cli_contract_tests.rs:400`) and `cache_cli_contract` covers all three actual subcommands (`inspect`, `prune`, `prune-project` — matching `crates/sifr/src/cache_cli.rs:8-30`, none uncovered), help text, `--reserve-bytes [default: 0]`, `--dry-run`, global placement at every position, and 14 malformed-argument cases all asserting exit code 2.
- **Isolated process contracts**: the harness binds `SIFR_CACHE_DIR` to a private output-scoped cache, never user storage; 34 calls recorded, JSON shapes, reserve default/no-op, dry-run byte-identical cache assertion, real deletion counts, idempotent re-run, and the `resource blocker:` stderr path.
- **Moved workspace**: `moved_workspace_misses_without_changing_diagnostics` (`crates/sifr_driver/src/project_cache/tests.rs:515`) proves miss + recompute at the new canonical root and equality against fresh `compute()` for both a clean and a diagnostic-producing source, with a warm re-hit at the new root. No relocation reuse was introduced.
- **Links**: all four moved targets resolve under `plans/issues/archive/`; `milestone_psp_7_parity_governance_inventory.md` exists nowhere in the tree or at its historical `verification/stdlib/` path (removed in `accf50125`), so the replacement historical reference plus "the referenced inventory report is no longer present" is honest, and its `#milestone_psp_7` anchor resolves to the real record. Roadmap status cells are unchanged; `item: T` was correctly not touched.
- **Evidence**: `reuse.json` binds the unchanged Rust/Markdown inputs to `0297d3f14`, whose `focused.json` preserves the failed harness row at `returncode: 1` rather than relabeling it; the final-candidate `focused.json` reruns diff, size guard, Python syntax and the receipt-bound harness. Receipt artifact sha256 matches the harness-recorded binary, `source_commit` is the candidate, lane is functional-only with no performance claim. Working tree is clean at `a7e45101` and all five file hashes match `evidence-index.json`.

PR #3887, opened 2026-09-20
Qualify final compiler DX follow-up contracts

yaseralnajjar:
Native exact-package qualification now covers the DX follow-up cache CLI, private trace modes/product identity, ownership, root movement and supplied embedding context contracts. The shared cache harness uses its caller’s sanitized environment; source tests use an owned canonical temporary root so macOS temporary-path aliases do not prevent cache fixture admission.

Final validation candidate: `db38e61773b351399a0217e502d8011b1212def7`. Exact package source: `f6865cdfb8aa8d0f7d62759ec4fe313e5edc3708`; compiler/package inputs are unchanged by the validation-only corrections. Temporary continuation configuration is retained externally and removed from the final tree.

Validation: 49 selected Rust tests (22 fresh, 27 input-equivalent retained), plus five affected source-fixture reruns; package/history/cache/interface/project process coverage; 34 installed hostile-environment calls; five fresh optimized installed trace contracts. Exact commands, nonzero counts, receipts, hashes, lock/configuration/toolchain/fixtures/environments and prior item reviews are in the external candidate-keyed evidence index. Source-size, syntax, format/HIR, documentation/link and diff checks passed as applicable.

Four native targets are covered. The [original run](https://github.com/sifr-lang/sifr/actions/runs/35513780133) remains failed: Linux targets passed, while Mac jobs retained passing install/update/rollback/relocation/cache/trace stages and failed at source fixture admission. [ARM](https://github.com/sifr-lang/sifr/actions/runs/35519810319) and [Intel](https://github.com/sifr-lang/sifr/actions/runs/35520966256) continuations passed only the failed/unreached source/profile/loader/metadata stages using the same archives. No package rebuild or repeat of passed transitions/cache/trace stages. The initial continuation artifact-name admission failure is preserved separately. Aggregate coverage includes 20 exact native source tests, 136 cache calls, four private trace checks, and 740 installed-metadata commands per target.

Scoped Opus: [SATISFIED, no blockers](https://github.com/sifr-lang/sifr/pull/3887#issuecomment-5751456217). Evidence index: `/home/yaser5/projects/sifr/dxf-evidence/db38e61773b351399a0217e502d8011b1212def7/evidence-index.json`, SHA-256 `897eabc6216e6992140ae14979a2ac7d566aba6011612789e12b58af493fa5a3`.

No performance/memory claim, old whole-phase gate/review restart, release/publication or DXF.8 local-user checkout/binary change. The pre-existing automatic workflow admission failure is recorded separately; it is not a compiler test failure.


yaseralnajjar:
Scoped read-only Opus review of base `4f9fc5c0d8ac82d5a9e01c4dc4b20450ebf23e09`, candidate `db38e61773b351399a0217e502d8011b1212def7`. Review input index SHA-256 `45c87fe3a1c0e55bc05cd731448f9a7f8feaf3a2d0dbc12f10417fdc5464c0db`. The response below is preserved verbatim outside the reviewed tree. Nonblocking index/layout clarifications are applied as record-only dispositions; historical mutable-target artifact limitations remain explicit. No code changed after review.

Read-only review complete. I verified the diff (4f9fc5c0 → db38e6177), the DXF.7 criteria in `plans/issues/active/ad-hoc-compiler-dx-followup-execution.md:391`, the immutable index at `/home/yaser5/projects/sifr/dxf-evidence/db38e617.../review-input-index.json`, its reconciliation/equivalence/reuse files, and the referenced run records. No tests or builds were run.

Key verifications performed:
- **Index integrity**: every `{path, sha256}` pair in the index resolves and re-hashes correctly (one exception, below). `review-input-index.json` is byte-identical to `evidence-index.json`; final input hashes for the three changed Python protocols match the final worktree bytes exactly.
- **Library reconciliation**: `rust-evidence.json` has 41 rows = 22 fresh selections (21 at `464d03c6`, 1 at `f6865cdf`) + 19 retained rows covering 27 tests = 49; all logs exist, hashes match, no failure text, each row carries expanded argv, nonzero `selected_count` and result summaries. `git diff 464d03c6..db38e617 -- crates Cargo.lock Cargo.toml rust-toolchain.toml .cargo stdlib` is empty, so the fresh selections are genuinely input-equivalent at the final candidate. The five post-correction source selectors are counted separately (`additional_affected_fixture_test_count: 5`), not folded into the 49.
- **Changed-harness reruns**: `native_dxf_contracts.py` (`a4acac96`) and `dxf_cache_cli_acceptance.py` (`ace06aab`) at `99250fccc` are byte-identical to final, so the Mac continuations executed final helper bytes; the contributor cache34 and the installed hostile-environment 34 were rerun at `a4dbfa387` with the final cache-harness bytes; `hostile-before.log` records the genuine `SIFR-STDLIB-0003` installed-compiler failure, and only the `after` probe passes the sanitized environment.
- **Native composition**: original run 35513780133 retains `conclusion: failure` with the two Mac `dxf-0-run` fixture failures preserved verbatim; the continuation script asserts prior `status == "fail"`, revalidates every prior evidence digest, requires the extracted archive binary to equal `binary_sha256` of the original package, retains transitions/cache34/trace, and runs only `qualify_source`, `native_profiles` (which contains the loader validation at `qualify_native_package.py:311`) and `metadata_corpus`. The failed admission run 35519633154 is preserved as a failure with no compiler tests claimed. The temporary workflow/launcher are absent from the final tree and `release-qualification.yml` at `db38e617` hashes identically to base.
- **No new claims**: limits explicitly disclaim speed/memory, old-binary relabeling, Windows, publication and DXF.8 installation; the pre-existing `local-first-validation` failure is recorded as pre-existing, unrepaired and not a compiler test failure.
- **Code sanity**: `qualify()`'s only `compiler_lanes` use in the shared path is the pure `digest()`, so the native caller needs no `configure()`; `owner.run` merges `self.env | env`, so the `TMPDIR` override keeps PATH/HOME; `output.resolve()` makes the owned TMPDIR canonical, which is exactly the `/var` alias cause; the transport PATH shim only shadows `curl`, so forwarding `owner.env` into the cache harness is behaviorally equivalent to the original CI environment.

Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: `native_qualification.protocols` lists `qualify_native_package.py`, `native_dxf_contracts.py` and `release-qualification.yml` but omits `verification/areas/developer_tooling/dxf_cache_cli_acceptance.py`, even though it is a changed native protocol and does appear under `executed_protocols`. Its final hash is still bound via `inputs`/`input-sha256.json`, so nothing is unproven — the asymmetry is only an indexing gap worth closing in the record-only update.
- suggestion: the `hostile-before` failure log (the demonstration that motivated the environment fix) is reachable only through the superseded `a4dbfa387/package-input-equivalence.json`, not from the final index, whose `environment_fix`/`hostile_environment_regression` entries point at the `after` evidence only. Linking the before log from the final index would make the preserved failure self-contained.
- pre-existing issue: `items[4]` (DXF.5) `historical_evidence.test_artifact` points at `target/debug/deps/sifr-152a076181dc3f50`, whose current bytes no longer match the recorded hash. This is inherited from the DXF.5 index and is correctly labeled historical, but a hash of a path inside the live rebuildable target directory is not re-verifiable later; future items should copy such artifacts into the candidate-keyed evidence directory.
- infrastructure issue: the Mac continuation executes `native_profiles`/`metadata_corpus` against the archive extracted directly to `owner.output/exact-package` rather than the installer-managed generation used on Linux. The binary bytes are asserted equal to the original package and the installer transitions were separately proven on the same platform in the original run, so coverage is intact; the distinction is just not spelled out in `native-reconciliation.json`'s prose, which says "executes actual installed archives".

PR #3889, opened 2026-09-20
docs(dx): record qualified local compiler handoff

yaseralnajjar:
DXF.8 completes the actual Mac ARM handoff: safely fast-forwards the user checkout while preserving IntoIterator, retains the complete legacy beta installation, and installs the exact DXF.7-qualified package through the unchanged atomic installer. Fresh login shells select the durable installed binary; existing shell profiles are unchanged.

Validation: full installed integrity and 89 metadata records; package body-edit/recheck matches a fresh reference with ten restored modules; all 34 isolated cache CLI checks; bounded redacted trace under umask 000; archive/installer/binary/receipt hashes and recovery preserved. Documentation links, diff and source-size guard pass. No rebuild, broad gate, performance claim or publication. Scoped Opus review pending.

Canonical record: plans/issues/active/ad-hoc-compiler-dx-followup-execution.md. Exact local/remote evidence and legacy recovery paths are recorded there.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- suggestion: the record states `.zshrc`/`.profile`/`.zprofile`/`.bashrc` hashes "remain unchanged" and that the installer regenerated sh/fish env files, but the retained external evidence contains only pre-install hashes (`before.json`) plus `installed-env` and `prior-sifr.env.fish`; no post-install hash record or new fish file is bound by `evidence-index.json`, so that specific claim is not re-verifiable from the external copy alone. Reason: reconstructable later by re-hashing on the Mac; the reviewed statement itself is consistent with all retained artifacts, so this is evidence completeness, not accuracy.
- suggestion: the legacy installation's schema-1 `install.json` body is referenced ("contained only bin/sifr, env and install.json") but only its SHA-256 is retained externally; the content survives solely inside the on-Mac backup root. Reason: recovery remains exact, yet the legacy-receipt characterization is not independently checkable off-host.
- pre-existing issue: the qualified installer names each permanent generation after its `mktemp` staging basename (`sifr-installer-0.1.0:632-636`), so the durable user-visible path retains a `.stage.dV0mkX` segment. The record reproduces the real path faithfully, and the package source is immutable and separate from main, so this is correctly out of DXF.8 scope; worth an owner-visible note for a future installer item.

Verification performed (read-only): the diff touches only `plans/issues/active/ad-hoc-compiler-dx-followup-execution.md` (100 insertions, 5 deletions), with status line, dependency-table row and "Current handoff" updated consistently; all 126 `evidence-index.json` entries exist and hash-match, with no extra or missing files; checkout SHAs, arm64, `?? IntoIterator` zero-byte hash, legacy beta.14 binary hash and complete three-file legacy root match `before.json`; installer/archive hashes (`210c67ad`, `38d7a7fe`) and binary/sysroot/lock digests match both `dxf7-handoff.json` and `qualification-aarch64-apple-darwin.json`; installed receipt, resolved generation path and `shell-after.stdout` agree with the recorded path and `0.1.0`; `smoke/report.json` shows status PASS with 89 doctor module records, integrity verified, `diagnostics_equal` true with exactly 10 restored / 1 computed modules, 34 cache calls (13 argument-error exits are the asserted expectations), and a 1,207-byte schema-1 trace with 0700/0600 modes and sentinel exclusion asserted in `smoke.py`; the installer's mutable-layout rejection at lines 619–624 corroborates the stated reason for owner-authorized backup; the `self version --json` typo record is preserved and no failed smoke is presented as passing.

PR #3893, opened 2026-09-20
test(driver): reject private imports after package cache reuse

yaseralnajjar:
A cached successful package check must not conceal an edit that imports a symbol through a private dependency module. Add an independent regression using actual Cargo metadata from an app package cwd and a local dependency: warm and restore legitimate importer checks, reject the private import with SIFR-PACKAGE-0203 identically in warm, fresh and empty-cache checks, then restore the public import and prove eligible reuse recovers.

No production defect was demonstrated; this closes only DXF1-F2's negative-evidence gap. The initial unsupported bare-import probe is retained as failed test-construction evidence.

Validation on Rust 1.98.1, default features, the existing private target and two jobs: formatting, diff, file-size and HIR guards passed; the exact new selector listed one test and passed; all four package_reuse_tests passed. Scoped remote Opus review: SATISFIED, no blockers, exact candidate 2f3dfadda5ab0c0498ea7f32b0d616eb6dddec31; https://github.com/sifr-lang/sifr/pull/3893#issuecomment-5753285448. No monolithic/performance gate is claimed; existing CI/Windows/cold-host/determinism owners remain unchanged.


yaseralnajjar:
**Verdict: SATISFIED**

Blocking findings:
- None.

Evidence checked against the fixed acceptance (`crates/sifr_driver/src/project_cache/package_reuse_tests.rs:344-410`):
- Real authority: `package()` (line 66) shells `cargo metadata --offline` from the app cwd and builds the actual `derive_package_graph` + `PackageSourceMap`; the private entrypoint is rebuilt after the edit (line 371), so no stale graph/token conceals resolver state. `dep/src/hidden.sifr` is materially created (line 351) before the first publish, so visibility — not a missing module — is what is exercised. Default-private semantics match the emitter contract in `crates/sifr_package/src/diag/mod.rs:303-325` and the existing driver test `crates/sifr_driver/src/tests/package_visibility_tests.rs:5`.
- Warm legitimate baseline then eligible body edit proving importer restoration: lines 356-364 with `assert_reused` (line 145) pinning `middle.sifr`/`main.sifr` as `restored` and the edited helper as `computed`.
- Supported `from dep.hidden import value as other` edit asserts exactly one diagnostic, `SIFR-PACKAGE-0203`, and the private module naming `private module 'dep.hidden'` (lines 374-378).
- Identical fresh/warm/independent-empty-cache diagnostics: `expected` (fresh), `actual` (warm, line 380), `cold` (line 393) all compared equal; cold cache asserted non-existent beforehand (line 385), reached via the new `run_with_cache` seam (line 128) which leaves `run` behaviorally unchanged.
- Zero restored checks / no restored success: `restored_checks == 0` (lines 381, 394) is the decisive assertion — every success-returning restore path sets it to 1 before returning (`crates/sifr_driver/src/project_cache/mod.rs:203` and `:235`), so the additional `status != "restored" / "interface-restored"` checks do not leave the `interface-restored-write-unavailable` / `-uncacheable` / `-changed-inputs` variants uncovered.
- Non-vacuity of the warm lookup: `package_context::identity` (`crates/sifr_driver/src/project_cache/package_context.rs:5-35`) digests the graph and source map, not module bytes, so the private edit keeps the same store identity and the prior published/interface-restored generation is genuinely consulted rather than bypassed.
- Restoration recovery: public source restored, success equals fresh (lines 396-400), then a further eligible helper edit re-proves importer restoration via `assert_reused` (lines 401-405).
- Scope respected: tests and issue record only; no production, resolver, reuse, or import-grammar change in the diff. The issue update (`plans/issues/active/ad-hoc-dxf1-package-reuse-review-followups.md:25-62`) records the acceptance, the gate exception, and the superseded bare-import probe as retained failed evidence without reclassifying external owners.

Follow-up findings:
- suggestion: `package_reuse_tests.rs:350` — `assert_private_import_rejected` is a single-caller helper parameterized by one import string; if no second private-import form is planned, inlining it into the test would remove the indirection. Non-blocking, and keeping it is reasonable if a second form is expected.
- suggestion: `package_reuse_tests.rs:380-394` — the warm negative relies on `restored_checks == 0` for non-vacuity but never positively asserts that history was present at that point (e.g. `retained_records > 0` or `candidate_checks > 0`). Today non-vacuity holds by construction via the stable package identity; an explicit assertion would keep the negative from silently degrading if identity derivation later becomes source-sensitive.
- infrastructure issue: CI admission, deferred Windows, and interrupted cold-host performance/determinism remain with their existing owners; nothing in this candidate touches or reclassifies them, and no performance claim is made or implied.

PR #3895, opened 2026-09-20
Reconcile strict Clippy for affected DXF targets

yaseralnajjar:
DXF.2/3 introduced two strict-Clippy failures in project-cache report aggregation and borrowed-record reuse. Borrow the report and pass the existing record reference directly. Reconcile the affected Rust packages and inline test configurations with mechanical test lint fixes, preserving assertions and evidence output. The NUL-count assertion uses the already pinned workspace memchr crate as a dev dependency; Cargo.lock adds only that dependency edge.

Validation: Rust 1.98.1, default features, existing private target and two jobs; strict Clippy for sifr_syntax/sifr_frontend/sifr_driver/sifr_lsp all targets, sifr production binary, and its explicit test profile (verified rustc --test). Formatting, source-size/HIR, diff and documentation guards. Two exact targeted assertions cover byte counting/cancellation and trace boundary lengths.

The broader attempted all-target selection exposed 46 pre-existing unwrap warnings in the unrelated formatter integration target. The new owning record preserves those failures and its next action; this PR does not claim workspace-wide lint qualification. The original suspected ReadDir.into_iter warning was not observed and is not patched. Cold-host CI performance, interrupted determinism and Windows portability remain separately recorded.


yaseralnajjar:
Verdict: SATISFIED

Blocking findings:
- none

Follow-up findings:
- pre-existing issue: `crates/sifr/tests/formatter_discovery.rs` (46 `unwrap_used` errors, lines 7–221) remains unqualified; it is a distinct unchanged standalone integration target, correctly excluded from the final selection and recorded with an exact next action in `plans/issues/active/ad-hoc-formatter-integration-test-lints.md`. No workspace/all-target clean claim is made anywhere in the record.
- infrastructure issue: the final `clippy-core.log` and `clippy-core-final.log` are a fully-cached replay (single `Finished` line, no unit output), so the four core packages' all-target cleanliness rests on the earlier compiling run `preflight/clippy-cleanup-corrected.log`, which compiled `sifr_syntax`/`sifr_frontend`/`sifr_driver`/`sifr_lsp` and failed only on the excluded formatter integration target. The chain holds and the input identity is bound by `candidate-inputs.json`, but the strongest positive evidence is the preflight log, not the named final one.
- suggestion: the `eprintln!` → `writeln!(std::io::stderr(), …).unwrap()` conversions in `crates/sifr_driver/src/metadata_producer/tests.rs:214,352,431` and `tests_process.rs:188` satisfy `clippy::print_stderr` by using a construct the lint does not see rather than by an `#[allow]`. That is not a suppression, matches existing repo prior art (`crates/sifr_sql_*_tools/src/main.rs`), preserves the required stderr evidence, and has identical panic-on-write-failure semantics — but the lint's intent is bypassed rather than met, which is worth stating in the record if stderr evidence in tests is meant to stay permanent.
- suggestion: `crates/sifr_driver/src/metadata_reader/generation_tests.rs:38-43` changes the `hex` helper implementation (`map`/`format!`/`collect` → `fold` + `write!`) but no focused test was run for that file. Output is byte-identical by construction and the change is within the authorized mechanical cleanup, so this is inspection-only equivalence rather than executed evidence; the record's "identical hex formatting" phrasing is accurate.
- suggestion: the new section header in `plans/issues/active/ad-hoc-dxf3-housekeeping-review-followups.md:21` is dated 2026-09-21 while the bound CI log (run 35533124294) and the candidate commit are 2026-09-20. Cosmetic, but it makes the record one day ahead of its own evidence.

Verification notes for the semantics-preserving claims I checked directly: `memchr_iter(0, …).count()` is exactly the NUL-byte count (non-overlapping single-byte matches); the trace boundary loop preserves `MAX_BYTES-1 / MAX_BYTES / MAX_BYTES+1` with `delta > 0 ⟺ target_len > MAX_BYTES` and `delta <= 0 ⟺ target_len <= MAX_BYTES`; `*b"a/*?!.-"` equals the prior 7-byte alphabet in order; `std::ptr::from_ref` yields the same address as the prior `as *const _`; `HirNode::Type(_)` covers exactly the prior `_` arm (enum has only `Type`/`Expr`/`Function`, so exhaustiveness is now enforced, not weakened); `records` is `Vec<String>`, so `.first()` equals `.iter().next()`; `ProjectPruneReport::add(&Self)` and the `&record` removal change only ownership at the call sites, with counters, lease and cache-proof behavior untouched. The two production fixes match the two advisory failures in `preflight/ci-lint.log` (`needless_pass_by_value` on `ProjectPruneReport`, `needless_borrow`). `housekeeping.rs:128 .into_iter()` is unchanged and now measured clean under Rust 1.98.1, and DXF3-F1 is re-recorded as a suspicion rather than a failure. `memchr` was already a pinned workspace dependency, so the lockfile edge is one line with no version change. `cargo fmt --check` passes on the candidate tree; the file-size guard (4110 files) and HIR guard pass; all touched files are well under 900 lines; the working tree is clean at `f6ffa577a`.

